PR #96, opened 2025-10-08
add test for 2 cardano addresses registered with the same DUST address

PavelMal:
## 📝 Description

Add e2e for checking that 2 cardano addresses could be linked with the same DUST address

## 🧪 Testing Evidence

Before execution: 
1. Your local env should be run, to run go to local-environment folder and run `npm run run:local-env `
You should see that all containers are up (current MN node version: version 0.17.0-b1c138f4)

Execution:
1. Run command from e2e directory `cargo test -- register_2_cardano_same_dust_address_production --nocapture`
2. You should see that test is passed

### DIFF
--- a/tests/e2e/tests/dust.rs
+++ b/tests/e2e/tests/dust.rs
@@ -70,6 +70,135 @@ async fn register_for_dust_production() {
 }
 
 #[tokio::test]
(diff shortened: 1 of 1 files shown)


rsporny:
cherry picked to https://github.com/midnightntwrk/midnight-node/pull/164

PR #155, opened 2025-10-22
feat: automate macOS LLVM setup for WASM compilation

cosmir17:
## Problem
macOS developers hit WASM build failures because XCode's default LLVM doesn't support WASM targets:
error: unable to create target: 'No available targets are compatible with triple "wasm32-unknown-unknown"'

## Solution

Automates LLVM setup via `.envrc`:
  - Detects Homebrew LLVM on both Apple Silicon and Intel Macs
  - Exports required env vars (`PATH`, `LDFLAGS`, `CPPFLAGS`) automatically
  - Warns with install instructions if missing

Since direnv is a prerequisite, docs simplified to reference `.envrc` automation.

## Credit
- Solution identified by @simongellis. Initial commit co-authored with Simon.
- Implements @gilescope's feedback to use `.envrc` automation.

PR : https://github.com/midnightntwrk/midnight-node/pull/155


cosmir17:
> I'd prefer it if these env vars were set up in the `.envrc` file. We could have a check to see if it's installed if we're on mac and suggest it should be installed if it's not on the path (or the version is not sufficient?)

@gilescope Done! Updated to use `.envrc` with:
  - macOS platform detection via `uname`
  - Checks both Apple Silicon (`/opt/homebrew`) and Intel (`/usr/local`) paths
  - Warns with `brew install llvm` if not found
  - Auto-exports PATH/LDFLAGS/CPPFLAGS when present

Skipped version checking - just need LLVM present for WASM. Let me know if you want that added!

cosmir17:
superseded by https://github.com/midnightntwrk/midnight-node/pull/285

PR #240, opened 2025-11-10
ci: allow jobs to be kicked off from merge queue.

gilescope:
This is prep for turning on merge queues.

https://shielded.atlassian.net/browse/TPM-773


chrisferry (approved):
👍

PR #256, opened 2025-11-13
fix: split out hardfork compiles

gilescope:
# Overview

Chef stopped working (around PR #42 but not clear why, so let's do something more robust and split out the hardfork compile so that the rest of the tests can get started earlier).


gilescope (commented):
LGTM

PR #288, opened 2025-11-20
[Governance] Improve inherents

NachoPal:
# Overview

This PR aims to improve the `pallet_federated_authority_observation` inherent mechanism by:

- Rejecting duplicated Midnight (sr25519) members. This was a security issue, as duplicated members cannot vote multiple times. Having too many duplicated members could potentially make it impossible to reach the required authority body motion thresholds. For example, with a Council of 6 members and a threshold of 2/3, if one member occupies 4 seats, it would be impossible to reach the required 4 votes.
- Fixing the `check_inherent` function to verify that the queried members match those set in the inherent `call`.

Closes: https://shielded.atlassian.net/browse/PM-20680

**UPDATE:** The inherent call now never returns an Error to prevent the chain from stalling. The call now will log the error and ignore the changes.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [x] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A


pallets/federated-authority-observation/src/lib.rs
```diff
@@ -179,38 +176,49 @@ pub mod pallet {
...
+			let council_members_len = council_members.len() as u32;
+			let technical_committee_members_len = technical_committee_members.len() as u32;
 
+			// ---- Council ----
 			// Make sure an empty set of members is not allowed
 			ensure!(!council_members.is_empty(), Error::<T>::EmptyMembers);
 			council_members.sort();
 
+			// Make sure there are not duplicated members
+			ensure!(
```
ozgb: Haven't seen this macro before - neat!

ozgb: Checking this - is it intentional to stop block production if applying new members fails due to input validation? This applies to all the `ensure!` statements in this call

I think better behaviour would be to emit an error event rather than returning an error

PR #367, opened 2025-12-12
fix: prevent feeless blockspace DDoS via pre-dispatch validation

m2ux:
## Summary

Add pre-dispatch validation to prevent transactions whose guaranteed part will fail from being included in blocks, closing a DDoS attack vector where attackers can fill blocks without paying fees.

🎫 [Ticket](https://shielded.atlassian.net/browse/PM-20944)  📐 [ADR](https://github.com/midnightntwrk/midnight-node/blob/fix/prevent-feeless-blockspace-ddos/docs/decisions/adr-prevent-feeless-blockspace-ddos.md)  🧪 [Test Plan](https://github.com/midnightntwrk/midnight-node/blob/fix/prevent-feeless-blockspace-ddos/docs/tests/test-plan-prevent-feeless-blockspace-ddos.md)

---

## Motivation

Midnight transactions have a two-phase execution model where the guaranteed part always executes (and extracts fees) while the fallible part may succeed or fail. A DDoS vulnerability exists where transactions can pass structural validation (`well_formed()`) but fail during guaranteed execution—consuming blockspace without paying fees.

An attacker can exploit this by flooding the network with structurally valid transactions designed to fail the guaranteed part (e.g., calling non-existent contracts, replaying transactions), filling blocks at zero cost and denying service to legitimate users.

---

## Changes

- **Ledger** - Add `validate_guaranteed_execution()` to simulate guaranteed part execution without modifying state
- **Bridge/Host API** - Expose `validate_guaranteed_execution` through the Bridge API and host functions
- **Midnight Pallet** - Enhance `pre_dispatch` hook to call validation before block inclusion
- **Tests** - Unit tests for all rejection vectors; E2E tests for attack simulation

---

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] I have included a change file, or skipped for this reason: Change file added
- [x] No new todos introduced

---

## 🔱 Fork Strategy

- [x] Node Runtime Update
- [x] Node Client Update
- [ ] Other
- [ ] N/A

---

## 🗹 TODO before merging

- [x] Ready for review

[PM-20944]: https://shielded.atlassian.net/browse/PM-20944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ledger/src/versions/common/api/ledger.rs
```diff
@@ -192,6 +192,56 @@ impl<D: DB> Ledger<D> {
...
+	/// - `Ok(())` if the guaranteed part would succeed (fees would be paid)
+	/// - `Err(LedgerApiError)` if the guaranteed part would fail
+	pub(crate) fn validate_guaranteed_execution<S: SignatureKind<D>>(
+		sp: Sp<Self, D>,
+		tx: &Transaction<S, D>,
+		block_context: &BlockContext,
+	) -> Result<(), LedgerApiError> {
+		// First, perform structural validation (well_formed check)
+		let ctx = sp.get_transaction_context(block_context.clone());
+		let valid_tx =
```
justinfrevert: Is this an extra check, since we already do this in the unsigned validation?

justinfrevert (approved):

ledger/src/versions/common/api/ledger.rs
m2ux: Let me evaluate the implications on weight ASAP

m2ux: No weight adjustment is needed for block execution because:

1. **Pre-dispatch is a validation phase, not execution** — The Substrate weight system charges `#[pallet::weight(...)]` during dispatch execution. The `pre_dispatch` hook runs *before* weight is charged and is part of block authoring, not block execution. The declared `ConfigurableTransactionSizeWeight` remains accurate for execution metering.

2. **Authoring vs execution cost distinction:**
   - Block *execution* weight: unchanged (validators replaying blocks don't run `pre_dispatch`)
   - Block *authoring* cost: increased (authors run both `pre_dispatch` and `apply`)

3. **Authoring overhead is the intended trade-off** — The dry-run in `pre_dispatch` approximately doubles CPU work per transaction during authoring. This is the cost of preventing the DDoS vector. Since authoring happens on a single node (the block producer), this is preferable to allowing feeless blockspace consumption that affects all validators.

PR #377, opened 2025-12-17
Register twice with the same cardano address

sergey-evseenko:
# Overview

Add e2e test for register twice with the same cardano address

## 📌 Submission Checklist

[✅] Pull request description explains why the change is needed
[✅] Self-reviewed the diff

## 🧪 Testing Evidence

You can run single test by executing: 
cargo test --test e2e_tests --no-default-features --features local -- register_twice_with_same_cardano_address --no-capture

Jira link:
https://shielded.atlassian.net/browse/PM-18635


rsporny (changes requested):
Test has no assertions, see my comment above

rsporny (approved):

PR #453, opened 2026-01-09
ci: don't rebuild everything if a change file changes.

gilescope:
When you make a PR you need a change file with a PR reference in it.
When you stick the PR in the change file, we don't want to restart the majority of the CI checks as they're unaffected.

https://github.com/midnightntwrk/midnight-node/pull/453


ozgb (approved):
LGTM

PR #459, opened 2026-01-10
feat: No need to use sp_std

gilescope:
Non-functional change: Removes dependency on `sp_std` crate.

https://github.com/midnightntwrk/midnight-node/pull/459


ozgb (approved):
LGTM - removes a layer of magic

PR #463, opened 2026-01-12
Feat: Add Utility Pallet

### LINKED ISSUE #1143: [Governance] Add `pallet-utility` to the Runtime

Add `pallet-utility` to the Runtime to allow Governance to batch calls. Without this, two governance actions can happen at the same block but one can get voted in while the other does not, or one can work while the other fails, leaving the chain in a potentially broken state. This allows multiple extrinsics to happen atomically.

## User Need

Governance participants need the ability to batch multiple extrinsics atomically to prevent the chain from ending up in an inconsistent state when multiple governance actions are submitted in the same block.

## Expected Benefit

Prevents broken chain state from partially-applied governance actions by ensuring multiple extrinsics execute atomically.

## Details

Special attention must be paid to `struct GovernanceAuthorityCallFilter;` to make sure unwanted nested calls in the batch are not allowed.

---

_Migrated from [PM-21016](https://shielded.atlassian.net/browse/PM-21016)_

[PM-21016]: https://shielded.atlassian.net/browse/PM-21016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

NachoPal:
# Overview

Added `pallet-utility` from polkadot-sdk to the runtime at pallet index 20.

With the addition of `pallet-utility` to the runtime, midnight transactions can now be dispatched through governance via batch calls (`Utility::batch`, `Scheduler::schedule`, etc.). The initial approach tried to extract these transactions by inspecting the inner calls of `FederatedAuthority::motion_close`, but this breaks when a Scheduler is involved — the transaction is detected at motion-close time but actually executed in a later block, causing `StateRootMismatch` in the toolkit's ledger replay.

The new approach mirrors what `pallet-midnight-system` does for system transactions: a new root-only extrinsic (`send_mn_root_transaction`, call_index 2) that emits `RootTxApplied` or `RootTxPartialSuccess` events containing the full serialized transaction payload. This makes extraction purely event-driven — no need to scan governance calls or worry about deferred execution via the Scheduler.

Additionally, `send_mn_transaction` is now restricted to unsigned origin only (`ensure_none`), so governance actions must use the new root extrinsic instead. This prevents midnight transactions from being dispatched through a path the indexer can't track.

### Changes

- **`pallet-midnight`**: Added `send_mn_root_transaction` extrinsic (Root origin only), `RootTxApplied` and `RootTxPartialSuccess` events with full tx payload, restricted `send_mn_transaction` to unsigned only, extracted shared logic into `apply_and_emit_tx_events` helper
- **Toolkit fetcher**: Added `root_tx_applied` to `MidnightMetadata` trait and event scanning in `compute_task.rs` (placeholder until metadata is rebuilt)
- **`pallet-utility`**: Added to runtime at pallet index 20

Closes: https://shielded.atlassian.net/browse/PM-21016

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [x] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A


ozgb (changes requested):
We should update toolkit to support this before merging - the `fetch` mechanism should be updated to handle batch txs correctly

Klapeyron (dismissed):
This PR is in conflict with https://github.com/midnightntwrk/midnight-node/pull/894, adding utility pallet can allow to bypass the filter via batch/batch_all extrinsics

ozgb (approved):
LGTM

ozgb (changes requested):
Let's add some integration tests

scripts/tests/toolkit-batch-governance-e2e.sh
```diff
@@ -0,0 +1,168 @@
...
+
+set -euxo pipefail
+
+NODE_IMAGE="$1"
+TOOLKIT_IMAGE="$2"
+
+NETWORK="batch-gov-e2e-net"
+NODE_NAME="midnight-node-batch-gov"
+WASM_DIR=$(mktemp -d)
+WASM_FILE="$WASM_DIR/runtime.wasm"
```
Klapeyron, line 31: which one it is?

- midnight_node_runtime.compact.compressed.wasm
- midnight_node_runtime.compact.wasm
- midnight_node_runtime.wasm

ozgb: Updated - we're now extracting the wasm (`midnight_node_runtime.compact.compressed.wasm`) from the node image rather than via RPC

runtime/src/lib.rs
```diff
@@ -556,6 +556,13 @@ impl pallet_scheduler::Config for Runtime {
 	type BlockNumberProvider = frame_system::Pallet<Runtime>;
 }
 
+impl pallet_utility::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type RuntimeCall = RuntimeCall;
+	type PalletsOrigin = OriginCaller;
```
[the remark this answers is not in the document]
Klapeyron: @ozgb please do not forget about this one

runtime/src/lib.rs
```diff
@@ -517,6 +517,13 @@ impl pallet_scheduler::Config for Runtime {
 	type BlockNumberProvider = frame_system::Pallet<Runtime>;
 }
 
+impl pallet_utility::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type RuntimeCall = RuntimeCall;
```
Klapeyron, line 522: we should also update the CheckCallFilter for batch calls & check BaseCallFilter in frame_system::Config if it works with TxPause -> maybe it is also worth to document somewhere that batch can bypass TxPause filter when authorized as root.
Also it seems reasonable to temporary filter-out everything apart batch like `pallet_utility::dispatch_as` and other utility calls.

pallets/midnight/src/lib.rs
```diff
@@ -368,44 +380,111 @@ pub mod pallet {
 	impl<T: Config> Pallet<T> {
 		#[pallet::call_index(0)]
 		#[pallet::weight(Pallet::<T>::get_tx_weight(midnight_tx))]
-		pub fn send_mn_transaction(_origin: OriginFor<T>, midnight_tx: Vec<u8>) -> DispatchResult {
+		pub fn send_mn_transaction(origin: OriginFor<T>, midnight_tx: Vec<u8>) -> DispatchResult {
+			// Only unsigned transactions are allowed — governance must use
+			// send_mn_root_transaction instead.
```
Klapeyron, line 385: I am thinking how to express that in code, but it may be worth to document somewhere that if in the future this call will be meant to be changed from unsigned to singed or authorized with transaction extensions for v5 extrinsics, then it has consequences for filtering gateway.
For gateway it will be possible to wrap send_mn_transaction with batch so it will bypass the filter.

LGLO:
We'll proceed with #1541

PR #474, opened 2026-01-14
feat: bump node version to 0.21

gilescope:
Bump version of node following node 0.20 rc1 being cut.

https://github.com/midnightntwrk/midnight-node/pull/474


gilescope:
outdated

PR #485, opened 2026-01-15
add rotate cNight script and update README

PavelMal:
- Document that deregister.sh alone does not immediately stop DUST production (DUST may still be produced until UTxOs are rotated).

 - Add a follow-up step to run rotate_tokens.sh for the deregistered users to rotate UTxOs.

 = Clarify that after rotation, the previous cNIGHT-containing UTxOs stop participating in DUST production (assuming deregistration already happened).

jira link: https://shielded.atlassian.net/browse/PM-21274


scripts/cnight-generates-dust/README.md
```diff
@@ -28,6 +28,8 @@ A brief summary of what the scripts do:
   * `receive_cnight.sh` - mint 10 cNIGHT tokens.  These are dummy tokens made
     only for testing.
 
+  * `rotate_tokens.sh` - rotation of 10 (this is a configurable parameter in the scrip ) cNIGHT tokens is carried out by sending them back to your own address. If deregistration happens, the   UTxO that contains cNIGHT no longer participates in DUST generation.
```
rsporny:
```suggestion
  * `rotate_tokens.sh` - rotation of 10 (this is a configurable parameter in the script) cNIGHT tokens is carried out by sending them back to your own address. If deregistration happens, the   UTxO that contains cNIGHT no longer participates in DUST generation. 
```

PR #505, opened 2026-01-20
chore: update genesis + chainspecs

gilescope:
Update genesis due to recent removal of sudo. Fix for RC.2

https://github.com/midnightntwrk/midnight-node/pull/505


hseeberger:
> Fix for RC.2.

But this will result in a rc.3, right?

PR #575, opened 2026-01-30
fix: prevent duplicate inherent execution within same block

mpskowron:
# Overview

Mandatory extrinsics cannot be rejected, thus, if we won't protect from duplicated execution, malicious block producer could flood the block with them.

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [x] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A

## Links

Ticket: https://shielded.atlassian.net/browse/PM-21649


pallets/cnight-observation/src/lib.rs
```diff
@@ -582,6 +587,13 @@ pub mod pallet {
 		}
 	}
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_finalize(_n: BlockNumberFor<T>) {
+			InherentExecutedThisBlock::<T>::kill();
+		}
+	}
```
Klapeyron: there is a requirement, that for each on_finalize call, we need to add its weight in advance to on_initialize call:
https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/trait.Hooks.html#method.on_finalize
it will be single DB write in particular, so extending it with `T::DbWeight::get().reads_writes(0, 1)` is enough. If there is no on_initialize, we need to introduce so and just return the weight

Klapeyron (approved):

NachoPal (approved):

PR #723, opened 2026-02-19
fix: upgrade deps to fix audit issue.

gilescope:
and move some dependencies to be dev dependencies.


gilescope:
someone else fixed it

PR #744, opened 2026-02-22
feat: replace soft tx cache with revalidation-based validation cache

### LINKED ISSUE #1159: Implement Ledger's validation improvements [mempool]

Implement ledger re-validation for transactions in the mempool, based on the re-validation feature from [midnight-ledger-prototype PR #551](https://github.com/midnightntwrk/midnight-ledger-prototype/pull/551).

## User Need

The mempool re-validates each transaction on every new block import. Using the lighter re-validation instead of full validation for already-validated transactions would improve performance.

## Expected Benefit

Reduced mempool overhead on block imports. The soft-cache can be removed in favor of re-validation.

## Details

- The mempool currently re-validates each transaction on every new block import — we can run re-validation instead if the transaction has already been validated
- We can then remove the soft-cache, and re-validate instead
- A benchmark should be run before making this change — re-validation is cheap, but reading from a cache is cheaper

---

_Migrated from [PM-21736](https://shielded.atlassian.net/browse/PM-21736)_

[PM-21736]: https://shielded.atlassian.net/browse/PM-21736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### LINKED ISSUE #1178: Implement Ledger's validation improvements [Block Proposal]

Implement ledger re-validation for transactions in the block proposal path, based on the re-validation feature from [midnight-ledger-prototype PR #551](https://github.com/midnightntwrk/midnight-ledger-prototype/pull/551).

## User Need

Block proposers need an optimized validation path. With the re-validation feature, they only need to re-validate transactions when adding them to a block instead of performing full validation.

## Expected Benefit

Reduced block proposal overhead by using lighter re-validation instead of full validation for already-validated transactions.

## Details

This gives us an optimization for block proposers — they only have to re-validate the transactions when adding them to the block, rather than performing full validation from scratch.

PR: https://github.com/midnightntwrk/midnight-node/pull/744

---

_Migrated from [PM-18691](https://shielded.atlassian.net/browse/PM-18691)_

[PM-18691]: https://shielded.atlassian.net/browse/PM-18691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mpskowron:
# Overview
Remove the soft transaction cache and introduce tx revalidation: cached VerifiedTransaction entries are reused when state changes by revalidating against a RevalidationReference instead of re-running full ZK proof verification. Adds cache metrics (miss, strict hit, revalidation hit) and tests covering the full validation lifecycle. Metrics are not backwards compatible - require grafana dashboards update.

Revalidation skips most costly part of transaction verification which is independent of ledger state - ZK proofs. This should be a significant performance improvement. We also rerun apply on every mempool revalidation now - this is slower than the previous solution (cache it once and never rerun in mempool), but from the other hand it lets us invalidate transaction from the mempool earlier.

## 🗹 TODO before merging

- [x] Benchmark performance
- [x] Ready

## 📌 Submission Checklist

- [x] Changes are backward-compatible (or flagged if breaking): grafana metrics need updating
- [x] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

TBD

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [x] Node Client Update
- [ ] Other: 
- [ ] N/A

## Links

Jira1: https://shielded.atlassian.net/browse/PM-21736
Jira2: https://shielded.atlassian.net/browse/PM-18691


primitives/ledger/src/lib.rs
mpskowron: Changes here require the update of grafana dashboards, but I think they will give us much better insight on the performance impact of this change

ledger/src/versions/common/mod.rs
```diff
@@ -842,91 +839,138 @@ where
...
+	fn revalidate_transaction(
+		ledger: &Ledger<D>,
+		tx: &Transaction<S, D>,
+		block_context: &BlockContext,
+		prev_state: &LedgerState<D>,
+		tx_validation_key: TxValidationKey,
+	) -> Result<(VerifiedTransaction<D>, TxValidationCacheOutcome), LedgerApiError> {
+		let ctx = ledger.get_transaction_context(block_context.clone())?;
+		let revalidation_ref = mn_ledger_local::verify::RevalidationReference {
+			previously_validated_state: prev_state.clone(),
```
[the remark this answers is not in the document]
mpskowron: I've misread your question, I thought that you are proposing state_key instead of LedgerState. We cannot store only state_hash, cause revalidation function from ledger expects full old ledger state. However, as I wrote in one of the previous comments, LedgerState is a bag of pointers, so it is relatively small.
I'd keep it as is.

ledger/src/versions/common/mod.rs
```diff
@@ -842,91 +839,138 @@ where
...
+						"🚫 Rejected transaction {} from mempool: {e}",
+						tx_hash_hex
+					);
+					return Err(e);
+				},
+			};
 
-		// Check soft cache first (quick tx_hash-only lookup for mempool revalidation)
-		if let Some(cached) = SOFT_TX_VALIDATION_CACHE.get(&soft_key) {
-			return cached.map(|_| true);
```
mpskowron, line 913: Now we revalidate tx for every state change. Revalidation is several times faster than full validation, but this change will add some work for mempool. However, revalidating more often for the mempool makes revalidation during pre-dispatch cheaper (smaller state change). I think that pre-dispatch run-time is more important though. Thoughts?

ledger/src/versions/common/mod.rs
```diff
@@ -87,25 +86,53 @@ use {lazy_static::lazy_static, moka::sync::Cache};
 pub const LOG_TARGET: &str = "midnight::ledger_v2";
 pub const MINT_COINS_DOMAIN_SEPARATOR: &[u8; 10] = b"mint_coins";
 
-#[derive(PartialEq, Eq, Hash)]
-pub struct StrictTxValidationKey {
-	state_hash: Hash,
```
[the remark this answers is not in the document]
ozgb: There was no specific reason - using the transaction's own hash makes more sense

ledger/src/versions/common/mod.rs
```diff
@@ -844,86 +841,127 @@ where
...
-		// Cache in strict cache (soft cache is managed by do_validate_transaction)
-		STRICT_TX_VALIDATION_CACHE.insert(strict_key, Arc::new(verified_tx.clone()));
-
-		Ok(verified_tx)
+		let verified_tx = Self::is_well_formed(tx, &ctx.ref_state, ctx.block_context.tblock)?;
+		TX_VALIDATION_CACHE.insert(
+			tx_validation_key,
+			Arc::new(TxValidationValue {
+				verified_tx: verified_tx.clone(),
+				state: ledger.state.clone(),
```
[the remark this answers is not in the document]
ozgb: To make this even less expensive and for clarity of intention, you could store an `Sp` of the ledger state here instead:
```rust
default_storage::<D>()
			.arena
			.alloc(ledger.state)
```

Multiple clones of the ledger state has no impact on Ledger performance

ledger/src/versions/common/mod.rs
ozgb: We should performance test this before we merge - re-validating every transaction in the mempool on every new block is more work than before

~~The `apply()` call on every new block is also new - likely this will have a big performance hit.~~


~~Depending on results from perf testing, we could cache the `apply()` result for now with some small TTL (60s == 10 blocks?) as a compromise - when Ledger adds an API to run apply only for the guaranteed section, we can likely run that on every mempool validation without issue.~~

**EDIT:** Having re-read, this function will only re-run apply when there's a cache miss

ledger/src/versions/common/mod.rs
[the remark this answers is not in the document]
mpskowron: Changed

ledger/src/versions/common/mod.rs
```diff
@@ -90,25 +89,53 @@ use {lazy_static::lazy_static, moka::sync::Cache};
 pub const LOG_TARGET: &str = "midnight::ledger_v2";
 pub const MINT_COINS_DOMAIN_SEPARATOR: &[u8; 10] = b"mint_coins";
 
-#[derive(PartialEq, Eq, Hash)]
-pub struct StrictTxValidationKey {
-	state_hash: Hash,
+#[cfg(feature = "std")]
+#[derive(PartialEq, Eq, Hash, Clone)]
+struct TxValidationKey {
+	runtime_version: u32,
```
Klapeyron, line 95: I am thinking if we should also reflect somehow version of the ledger, in case it will be updated and for instance the cost model will be adjusted or just the underlying logic will be different

mpskowron: I don't think that cost model matter. I don't know if revalidation will work between old state and new state though. 
I'm assuming, that if we have a serious changes like that in between ledger versions, we will bump runtime as well. 
@ozgb what do you think about it?

ozgb: If there are breaking changes in the ledger version, we'll bump the runtime version too - so having just `runtime_version` is enough here

chrispalaskas:
## Regression found in perfnet A/B: watched txs report terminal `Invalid` after inclusion; `InFinalizedBlock` never delivered

We ran this branch (`214eebf`, = main `8fa07df` + this PR) against plain main in a 4-vs-4 validator A/B on perfnet (8 validators, 4 regions, 6s blocks), driving register-dust + funding load through `midnight-node-toolkit` `submit_and_watch`. The branch shows the expected perf win, but it also introduces a `submit_and_watch` correctness regression.

### Symptom

Under ≥ ~24 concurrent watched txs against one submit node running this branch: a tx is included in a block authored by **another** validator, the watcher receives `InBestBlock` — and then, when the submit node reaches its own authoring slot before the including block finalizes (~13–18s window), the stale pool copy fails pre-dispatch with `ReplayProtectionViolation(IntentAlreadyExists)` (it "already exists" because the tx itself is on-chain). The pool reports terminal `Invalid` and the subscription ends. `InFinalizedBlock` never arrives for a tx that finalizes normally seconds later.

Single-tx timeline (`0xf8e97065…9066`, submitted to euw1-v1 running this branch, 2026-07-28 UTC; GRANDPA healthy throughout — finality ≤3 blocks behind, no reorgs):

| Time | Event |
|---|---|
| 02:25:27 | validated into mempool on all 8 validators |
| 02:25:30–32 | included in block **#3113** (authored by a main-branch node); all nodes apply it |
| 02:25:31.6 | submit node imports #3113; client gets `InBestBlock` |
| 02:25:42.1 | submit node, authoring: `🚫 Rejecting transaction … at pre-dispatch: … ReplayProtectionViolation(IntentAlreadyExists)` → watcher gets terminal `pool reported Invalid: Extrinsic marked as invalid`, stream ends |
| 02:25:44.7 | **#3113 finalized** — 13s after inclusion; watcher already dead |

This held for **45/45** watched txs that reported finality failure in that run: every one included, never reverted, every one pre-dispatch-rejected on the submit node between inclusion and finality.

### A/B result (instrumented sender: on watcher death, falls back to checking whether the `InBestBlock` block finalized)

Two identical 24-worker runs, 1000 watched txs each, one submit node per leg:

| Submit node | Arm | Watcher killed after inclusion (block finalized anyway) | True finality failures |
|---|---|---|---|
| euw1-v1 | **this PR** (`214eebf`) | **85 / 1000 (8.5%)** | 0 |
| use2-v1 | **main** (`8fa07df`) | **0 / 1000** | 0 |

Disposal of the 45 included-but-"failed" txs across the fleet splits exactly on the arm boundary: all four main nodes evicted their stale copies via prompt post-import mempool revalidation (~0.5–1.5s, watchers unaffected); all four branch nodes never revalidation-evicted and only rediscovered the txs at their own pre-dispatch, killing watchers. At ~15 concurrent txs the branch does not reproduce (maintenance keeps up); at 24 it reproduces consistently, in bursts aligned with the submit node's authoring slots (~30–40% of in-flight txs per burst).

### Impact

Any `submit_and_watch` consumer (toolkit, SDK, indexer/partner tooling) sees landed, finalized txs reported as `Invalid`. Clients that retry then hit genuine `DustDoubleSpend` (196) / `NullifierAlreadyPresent` (239) — this looked like a ledger double-spend bug until node logs proved the "failed" txs were on-chain.

### Ask

Included txs need to be pruned as *included* (or otherwise excluded from pre-dispatch re-attempts / watcher invalidation) even with the revalidation-based cache — the regression seems to be that this branch drops the prompt post-import revalidation that previously evicted them before the node could re-author them. A regression test suggestion: watched tx included via node B while node A holds it pooled; node A authors before finalization; assert the watcher on A still gets `InFinalizedBlock` and never terminal `Invalid`.

Happy to share full node logs / client logs / analysis for the runs (perfnet 2026-07-27→29), or rerun the A/B against a fixed revision.

ledger/src/versions/common/mod.rs
```diff
@@ -926,91 +923,138 @@ where
...
+					);
+					return Err(e);
+				},
+			};
 
-		// Check soft cache first (quick tx_hash-only lookup for mempool revalidation)
-		if let Some(cached) = SOFT_TX_VALIDATION_CACHE.get(&soft_key) {
-			return cached.map(|_| true);
+		if !matches!(cache_outcome, TxValidationCacheOutcome::CacheMiss) {
+			return Ok(cache_outcome);
```
ozgb, line 1180: We should remove this `if` block, and let the function proceed to validate guaranteed execution even on a cache hit - this will ensure that doulble-spend-like errors are caught in mempool revalidation (`well_formed` alone won't check this)

`apply` is low-cost, so shouldn't affect mempool performance too much

ozgb (changes requested):
The revalidation cache is not fork-aware

If the cache contains a state reference belonging to a different fork, the transaction may revalidate with an incorrect result

This is due to the conditional gates in `RevalidationReference` - the gates may report no changes between the previous and current state, where on the current fork there are changes to the state that require re-running the checks:
https://github.com/midnightntwrk/midnight-ledger/blob/409fe3cb4b790c3f17de200c337e18aebf74c801/ledger/src/verify.rs#L216

Suggested fixes:
- Create a new implementation of `StateReference` that re-runs all the stateful checks unconditionally i.e. doesn't contain a `previously_validated_state` field. In other words, only the `stateless_check` is skipped.
- Keep track of state history for all in-flight ledger states in the node. This means we would be able to tell which cache-entry to re-validate against depending on which fork we're on.

ledger/src/versions/common/mod.rs
mpskowron: Looking at this, on main the apply is also not called, is it? 

```
if let Some(cached) = SOFT_TX_VALIDATION_CACHE.get(&soft_key) {
			return cached.map(|_| true);
```

that if above is short circuiting

ozgb:
> The revalidation cache is not fork-aware

After talking to @tkerber - the transaction revalidation is fork-agnostic, so our use of it here is safe. No need to track the origin fork of the state.

ledger/src/versions/common/mod.rs
```diff
@@ -118,39 +145,32 @@ const STRICT_TX_VALIDATION_CACHE_CAPACITY: u64 = 600;
...
+/// Time-to-live for cache entries.
 /// Unlike TTI, TTL evicts entries unconditionally after this duration regardless of access.
-/// This is critical for relay nodes (non-block-producers) where soft cache entries are never
+/// This is critical for relay nodes (non-block-producers) where cache entries are never
 /// invalidated by block authoring — without a TTL, revalidation keeps accessing entries and
 /// resetting the TTI timer, so invalid transactions persist in the mempool indefinitely.
 /// Set to 60s (~10 blocks at 6s/block) to balance eviction latency against revalidation cost.
 #[cfg(feature = "std")]
-const SOFT_TX_VALIDATION_CACHE_TTL: Duration = Duration::from_secs(60);
+const TX_VALIDATION_CACHE_TTL: Duration = Duration::from_secs(60);
```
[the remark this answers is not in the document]
ozgb: Agreed - we can remove this now: https://github.com/midnightntwrk/midnight-node/pull/744/commits/d1d6cc9f4630a3794be03faa4e283fff4267a9ea

ledger/src/versions/common/mod.rs
ozgb: Fixed in this commit - this PR now calls `apply` even on a re-validation cache hit: https://github.com/midnightntwrk/midnight-node/pull/744/commits/7165959b35a1d7c4314bed4a7a4733175bb764b6

ledger/src/versions/common/mod.rs
ozgb: Changed the behaviour - `apply` now runs on cache hits too: https://github.com/midnightntwrk/midnight-node/pull/744/commits/7165959b35a1d7c4314bed4a7a4733175bb764b6

`apply` is very cheap compared to proof verification, so performance impact should be limited

ozgb (dismissed):
This PR is looking in good shape now:
- Confirmed that re-validation is fork agnostic
- Txs are now correctly evicted from the mempool at block import if they turn stale

ozgb (approved):

ledger/src/versions/common/mod.rs
mpskowron: Do we need both do_validate_transaction and do_validate_guaranteed_execution anymore after this change? If we added `tblock_correction: Option<&TBlockCorrection>` as a param to do_validate_transaction wouldn't we be able to call it in place of do_validate_guaranteed_execution? Only difference is short circuiting on StrictCacheHit, but it should be correct for both cases, shouldn't it? Or am I missing something?

PR #764, opened 2026-02-24
fix: [PM-20013] validate slot duration and epoch configuration invariants

m2ux:
## Summary

Add three-layer defense-in-depth validation for consensus-critical timing parameters, preventing silent consensus divergence when nodes are misconfigured. Introduces config-file validation, on-chain canonical storage with startup consistency check, and per-block config hash verification in block headers.

**Solution needs a further deep dive to validate approach - MC**

🎫 [Ticket](https://shielded.atlassian.net/browse/PM-20013)  📐 [Engineering](https://github.com/shieldedtech/midnight-agent-eng/blob/main/.engineering/artifacts/planning/2026-02-24-slot-duration-configurability/README.md)  🧪 [Test Plan](https://github.com/shieldedtech/midnight-agent-eng/blob/main/.engineering/artifacts/planning/2026-02-24-slot-duration-configurability/05-test-plan.md)

---

## Motivation

Midnight nodes derive consensus-critical timing parameters — Cardano mainchain epoch duration, slot duration, first epoch timestamp, first epoch number, and first slot number — from operator-supplied configuration files, with no cross-validation, on-chain canonical reference, or mechanism for peers to detect disagreement. If any node's configuration diverges, it silently computes different slot indices and epoch boundaries, leading to consensus forks that are difficult to diagnose. This change introduces a three-layer defense-in-depth solution: first, fallible constructors and config-file validation that reject invalid or internally inconsistent parameters at startup; second, a new FRAME pallet that commits the five mainchain epoch parameters to on-chain storage and allows each node to compare its local values against the chain's canonical state before participating; and third, a per-block configuration hash deposited as a `DigestItem::Consensus` entry in every block header, enabling any peer to independently verify that the block author's consensus parameters match its own.

Because `MainchainEpochConfig` from the partner-chains framework is not SCALE-encodable — its field types (`sp_core::offchain::Timestamp`, `Duration`) lack `Encode`/`Decode` derives — the pallet stores each parameter as an individual `StorageValue` of primitive type (`u64` or `u32`), following the same pattern used by `pallet-sidechain` for `EpochDurationMillis`. On a new chain, genesis configuration populates these five storage items directly from network-specific values provided by the `MidnightNetwork` trait. On an existing chain receiving the pallet via runtime upgrade, storage starts uninitialized — the `on_initialize` hook skips the config hash deposit and the startup consistency check logs a warning and allows the node to proceed. This avoids the complexity of compile-time migration constants that cannot be correct across all networks in a single WASM binary. Storage is write-once with no update mechanism; Cardano mainnet epoch parameters are effectively immutable (unchanged since the Shelley hard fork in July 2020).

---

## Changes

- **Config Validation (`midnight_cfg`)** — Added `validate_mc_epoch_params` custom validator rejecting zero-value epoch/slot/timestamp parameters and enforcing epoch-slot divisibility at config load time. 7 unit tests.
- **Constructor Validation (`inherent_data`)** — Replaced `#[derive(new)]` with fallible `CreateInherentDataConfig::new()` returning `InherentDataConfigError` on cross-source divisibility violation (`mc_epoch_duration % mc_slot_duration == 0`). Resolves TODO ETCM-4079. 2 unit tests.
- **On-chain Storage Pallet (`pallet-consensus-config`)** — New FRAME pallet storing five Cardano mainchain epoch parameters as decomposed `StorageValue<u64/u32>` items with `GenesisConfig`. Write-once storage populated at genesis; no runtime migration. Inline `DbWeight` estimates used for constant-cost hooks (no input-dependent variation). 13 unit tests.
- **Network Configuration (`MidnightNetwork` trait)** — Extended `MidnightNetwork` trait with 5 mc_epoch methods providing correct per-network Cardano parameters (mainnet, preprod, dev). Values flow through genesis config naturally alongside other network-specific data.
- **Startup Consistency Check (`service`)** — `check_consensus_config_consistency()` compares local `MainchainEpochConfig` against on-chain canonical values via `ConsensusConfigApi` runtime API. Warns and skips when on-chain values are zero (pre-genesis or uninitialized); aborts on non-zero mismatch.
- **Per-block Config Hash (`pallet-consensus-config`)** — `on_initialize` deposits `DigestItem::Consensus(*b"MNCC", blake2_256(config))` each block. Skips when storage is uninitialized. Includes `decode_config_hash` helper.
- **Chain Spec** — Genesis config sourced from `MidnightNetwork` trait methods at chain spec build time. Each network provides its own correct Cardano epoch parameters.
- **Runtime Integration** — Pallet registered at index 12, `ConsensusConfigApi` runtime API implemented.
- **Tests** — 41 tests total (13 pallet + 18 node + 10 runtime), 0 clippy warnings, clean formatting.

---

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [ ] I have included a change file, or skipped for this reason: change file deferred pending team review of implementation approach
- [x] No new todos introduced

---

## 🔱 Fork Strategy

- [x] Node Runtime Update
- [x] Node Client Update
- [ ] Other
- [ ] N/A

---

## 🗹 TODO before merging

- [x] Ready for review


NachoPal (changes requested):
Is it really necessary right before launching mainnet? it seems like a very sensitive change for a very small benefit.

PR #780, opened 2026-02-25
feat: reduce bin sizes, image sizes, reduce linking mem usage

gilescope:
Non functional change. Don't ship with full debug symbols.

https://github.com/midnightntwrk/midnight-node/pull/780


oxidyn0 (approved):
Interesting. I did not know Node uses `debug` in release mode at all.

PR #790, opened 2026-02-26
fix: add escape hatch for when signing service is down.

gilescope:
We need to be able to publish RC builds when the service is down.

https://github.com/midnightntwrk/midnight-node/pull/790


adamreynolds-io (approved):
LGTM

PR #883, opened 2026-03-07
feat: reduce deps and reduce toolkit compile times

gilescope:
non-functional change to toolkit except pre 0.20 runtimes are put behind a flag and not compiled by default. 30 seconds saved.


metadata/src/lib.rs
```diff
@@ -1,25 +1,27 @@
 // Generate an interface that we can use from the node's metadata.
+#[cfg(feature = "legacy-metadata")]
 #[subxt::subxt(runtime_metadata_path = "static/midnight_metadata_0.17.0.scale")]
 pub mod midnight_metadata_0_17_0 {}
 
+#[cfg(feature = "legacy-metadata")]
```
ozgb: Would rather these old runtimes were deleted - afaik they are unused

PR #889, opened 2026-03-09
feat: tree-cache-guard — skip CI jobs when nothing changed

gilescope:
# Tree Cache Guard — Never re-run a green job for the same code

Git's tree hash is a content-addressable fingerprint of the entire repo. Same files = same hash, regardless of commit history. This PR uses it to skip CI jobs that have already passed for a given tree.

**A composite action wraps each job:** check `HEAD^{tree}` against the Actions cache:

- hit = exit in seconds, 
- miss = run normally then save a sentinel on success.

**Cache survives across events:** PR goes green → enters merge queue (same tree) → merges to main (same tree). One run pays for all three. Rebases, "Update branch", close/reopen — all cache hits if the content hasn't changed.

## Scope

20 leaf jobs guarded across 6 workflows (actionlint, scan, cargo test, cargo deny, fmt/lint, feature-unification, test-pallet-fixtures, check-unused-deps, rebuild-genesis, test-toolkit, metadata-check, and 9 E2E suites).

**Not guarded:** image-producing jobs (use GHCR checks instead), fast metadata jobs, bot workflows, reusable workflow calls.

**Also:** merged hardfork build + test into single "Hardfork E2E" job; replaced ad-hoc `docker manifest inspect` skip logic with the guard.

## Checklist

- [x] Changes are backward-compatible
- [x] Pull request description explains why
- [x] Self-reviewed the diff
- [x] Change file in #783
- [x] No new todos introduced
- [x] CI-only change, no impact on node binary or runtime

## Links

- Builds on #783 | ADR: `docs/decisions/0004-tree-content-hash-image-tags.md`


adamreynolds-io (approved):

.github/workflows/actionlint.yml
```diff
@@ -28,8 +27,17 @@ jobs:
...
+
       - name: Run actionlint
+        if: steps.guard.outputs.hit != 'true'
         uses: raven-actions/actionlint@205b530c5d9fa8f44ae9ed59f341a0db994aa6f8 # v2
         with:
           matcher: true # Enable GitHub annotations for errors
           fail-on-error: true # Fail CI when errors are found
+
+      - uses: ./.github/actions/tree-cache-guard/save
+        if: steps.guard.outputs.hit != 'true'
```
ozgb, line 41:
```suggestion
        if: !fromJson(steps.guard.outputs.hit)
```

We can use `fromJson` to convert to bool here:
https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#fromjson

ozgb (approved):
LGTM

.github/workflows/actionlint.yml
jsidorenko: IMO: `if: steps.guard.outputs.hit != 'true'` looks more readable than `if: !fromJson(steps.guard.outputs.hit)`

jsidorenko (approved):

cosmir17 (approved):
LGTM from external-reviewer perspective. Solid design, correctly scoped (guards leaf jobs, leaves image-producing jobs alone). Indexer might adopt the same pattern.

night-warden (approved):

PR #1056, opened 2026-03-21
fix: make sure qanet isn't unusual

gilescope:
Remove hardcodings for qanet and chain reset it.

https://github.com/midnightntwrk/midnight-node/pull/1056


cosmir17 (approved):
LGTM

One small thing to note for traceability: `qanet` is still in the match arm at `ledger/src/versions/common/utxo_ordering_override.rs`, so qanet will warn-log `"No UTXO ordering overrides for qanet (tried res/utxo-ordering-override-qanet.json): No such file or directory"` once per process (the `OnceLock::get_or_init` caches the empty map afterwards). It's a single startup-time log line, not per-transaction noise, so probably fine to leave for the follow-up PR that introduces `--non-determinism-compatability-mode` from #1057 along with the whole-module cleanup.

rsporny:
@gilescope reset qanet? I don't think we ever wanted to reset qanet, fyi @sineadplunkett

sineadplunkett:
Correct we if at all possible do not want to chain reset QANet unless it's absolutely necessary

Issue #1083, opened 2026-03-26
[C-to-M Bridge] Implement MidnightTokenTransferHandler


LGLO:
Current implementation is doing nothing. Proper implementation should be provided.

AI output below:
```
The handler must:

1. **Map the recipient address** (depends on Gap 2)
2. **Route by transfer type** — UserTransfer → credit recipient;
  InvalidTransfer → credit treasury (Gap 3); ReserveTransfer → deferred
3. **Construct a system transaction** —
  `SystemTransaction::DistributeNight(CardanoBridge, outputs)`
4. **Execute via `MidnightSystemTransactionExecutor`**

This also requires a new host API function
`construct_distribute_night_system_tx` (only `construct_cnight_generates_dust_*`
exists today). The `TransferHandler` trait is bare — no access to runtime
storage or ledger APIs. The implementation must be a pallet with `Config`
bounds providing `LedgerApi` and `MidnightSystemTransactionExecutor`,
following the `pallet-cnight-observation` architecture.
```

PR #938, opened 2026-03-13
[PM-22085] fix: ensure motion removal on failed dispatch in federated-authority pallet

m2ux:
## Summary

Ensure that approved motions in the federated-authority pallet are always removed from on-chain storage after `motion_close`, regardless of whether the dispatched call succeeds or fails. Adds test coverage for the failure scenario and removes dead error variants.

🎫 [Ticket](https://shielded.atlassian.net/browse/PM-22085)  📐 [Engineering](https://github.com/shieldedtech/midnight-agent-eng/blob/mike/.engineering/artifacts/planning/2026-03-13-fix-motion-removal-failed-dispatch/README.md)

---

## Motivation

The `motion_close` extrinsic in the federated-authority pallet used `motion_result?` to propagate dispatch errors after calling `motion_remove`. While this appeared to work, the `#[pallet::call]` transactional storage layer rolls back all mutations when the extrinsic returns `Err` — including the motion removal. This left motions with failing dispatches permanently stuck in storage with no recovery path, as identified by the Least Authority security audit (PM-19974).

---

## Changes

- **Dispatch error handling** — Remove `motion_result?` from the approved branch of `motion_close`. The dispatch outcome is captured in the `MotionDispatched` event; the extrinsic returns `Ok` regardless of dispatch success since the close operation itself (removal + events) always succeeds.
- **Test coverage** — Add `motion_close_removes_motion_on_failed_dispatch` test that creates an approved motion with a Root-failing call (`pallet_collective::Call::vote`), closes it, and verifies the motion is removed and correct events are emitted.
- **Dead code cleanup** — Remove 3 unused error variants (`MotionTooEarlyToClose`, `MotionAlreadyExists`, `MotionExpired`) that were never constructed.

---

## 📌 Submission Checklist

- [x] Changes are backward-compatible (or flagged if breaking)
- [x] Pull request description explains why the change is needed
- [x] Self-reviewed the diff
- [x] I have included a change file, or skipped for this reason: audit fix, no new features
- [ ] If the changes introduce a new feature, I have bumped the node minor version
- [x] Update documentation (if relevant)
- [x] No new todos introduced

---

## 🔱 Fork Strategy

- [x] Node Runtime Update
- [ ] Node Client Update
- [ ] Other
- [ ] N/A

---

## 🗹 TODO before merging

- [x] Ready for review

PR #1188, opened 2026-04-01
Cardano to Midnight bridge: handler implementation and pallet events added

LGLO:
# Overview

Implements Cardano-to-Midnight bridge handler.
According to the newest agreement, the pallet emits events that contain both Cardano Tx Hash and Midnight Ledger Tx hash.

This is backward compatible, because the bridge is not active in any environment.

This is not final, list of things to add is long and there are GH issues prefixed with `[C-to-M Bridge]`.

From functional perspective #1248 is the biggest chunk of work left for future.

Also integration and testing are blocked by the problem with ledger that crashes when bridge transacitons are made.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A

## Links

Closes: https://github.com/midnightntwrk/midnight-node/issues/1083

### DIFF
--- a/.grype.yaml
+++ b/.grype.yaml
@@ -24,3 +24,9 @@ ignore:
   # - No signals_based_traps config — stays at default (enabled)
   # TODO: Remove this ignore after upgrading to a polkadot-sdk version using wasmtime >= 36.0.7
   - vulnerability: GHSA-jhxm-h53p-jm7w   # Critical: wasmtime 35.0.0, fixed in 36.0.7
+
+  # wasmtime 35.0.0 critical vulnerability
+  # Only applies when using the non-default Winch compiler - we use cranelift
+  # More info: https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-xx5w-cvp6-jv83
+  # TODO: Remove this ignore after upgrading to a polkadot-sdk version using wasmtime >= 36.0.7
+  - vulnerability: GHSA-xx5w-cvp6-jv83   # Critical: wasmtime 35.0.0, fixed in 36.0.7


cosmir17 (commented):
Reviewed from the indexer side. The pallet Transfer event shape (mc_tx_hash, amount, result, recipient) matches what we agreed and gives the indexer everything it needs for bridge deposit tracking (#938-#944).

when the ledger repo adds a proper system tx type (GH #1277), node is going to update construct_distribute_treasury_system_tx to use it instead of PayBlockRewardsToTreasury?

LGLO:
> when the ledger repo adds a proper system tx type (GH #1277), node is going to update construct_distribute_treasury_system_tx to use it instead of PayBlockRewardsToTreasury?

Yes. Issue is there to not forget

runtime/src/lib.rs
```diff
@@ -834,11 +837,93 @@ parameter_types! {
 	pub const BridgeMaxTransfersPerBlock: u32 = 256;
 }
 
-impl pallet_partner_chains_bridge::TransferHandler<BridgeRecipient>
+/// Storage key for the bridge transfer nonce counter (transient, reset each block).
+const BRIDGE_TRANSFER_NONCE_COUNTER_KEY: &[u8] = b":bridge_transfer_nonce_counter:";
+
+impl MidnightTokenTransferHandler {
```
justinfrevert: Perhaps we could put midnight node specific handling and values inside its own pallet?

LGLO: What do you think about making bridge pallet dependent on required midnight code and putting it there?

ledger/src/versions/common/mod.rs
```diff
@@ -1040,6 +1040,37 @@ where
...
+		let api = api::new();
+		let system_tx = SystemTransaction::DistributeReserve(amount);
+		api.tagged_serialize(&system_tx)
+	}
+
+	pub fn construct_distribute_treasury_system_tx(
+		amount: u128,
+	) -> Result<Vec<u8>, LedgerApiError> {
+		let api = api::new();
+		//TODO: this is wrong transaction, ledger is missing the correct one yet. https://github.com/midnightntwrk/midnight-node/issues/1277
```
Klapeyron, line 1070: shall we wait until https://github.com/midnightntwrk/midnight-node/issues/1277 is fixed?

tests/e2e/src/api/midnight.rs
```diff
@@ -160,6 +163,50 @@ impl MidnightClient {
...
+                        return Ok(events);
+                    }
+                }
+            }
+            Err("Did not find registration event".into())
+        };
+
+        timeout(Duration::from_secs(60), inner)
+            .await
+            .unwrap_or_else(|_| Err("Timeout waiting for registration event".into()))
```
Klapeyron: "Timeout waiting for registration event" looks duplicated, shall we try to find some other description?

runtime/src/c2n_bridge.rs
```diff
@@ -0,0 +1,96 @@
...
+impl pallet_partner_chains_bridge::TransferHandler<BridgeRecipient, MaybeMidnightTxHash>
+	for MidnightTokenTransferHandler
+{
+	fn handle_incoming_transfer(
+		transfer: BridgeTransferV1<BridgeRecipient>,
+	) -> MaybeMidnightTxHash {
+		let amount = transfer.amount;
+		let serialized_tx = match transfer.recipient {
+			TransferRecipient::Address { recipient } => {
+				let recipient_bytes = recipient.as_bytes().to_vec();
```
Klapeyron: this one can be removed:
```suggestion
```
later we can use it as
```rust
recipient.as_bytes(),
```

partner-chains/toolkit/bridge/pallet/src/lib.rs
```diff
@@ -243,8 +244,11 @@ pub mod pallet {
 		/// Transfer recipient
 		type Recipient: Member + Parameter + MaxEncodedLen;
 
+		/// User defined handler returns this type. Values are attached to pallet events.
+		type HandlerResult: Member + Parameter + MaxEncodedLen;
```
Klapeyron, line 261: there is doc comment in line 97 in this file, we should extend it as well

runtime/src/c2n_bridge.rs
```diff
@@ -0,0 +1,96 @@
...
+
+impl MidnightTokenTransferHandler {
+	/// Generate a deterministic unique nonce for a bridge transfer.
+	///
+	/// Uses the parent hash (unique per block) combined with a monotonically
+	/// increasing counter (unique within a block) to guarantee uniqueness.
+	fn generate_nonce() -> [u8; 32] {
+		let counter: u32 =
+			frame_support::storage::unhashed::get(BRIDGE_TRANSFER_NONCE_COUNTER_KEY).unwrap_or(0);
+		frame_support::storage::unhashed::put(BRIDGE_TRANSFER_NONCE_COUNTER_KEY, &(counter + 1));
```
Klapeyron: I am not sure about that, sounds like we should introduce StorageValue for it and clear it in on_finalize

Klapeyron: some draft that should work for u64, u32 also sounds like it will never overflow?
```rust
let foo = Foo::<T>::mutate(|foo| {
    use sp_runtime::Saturating;
    foo.saturating_inc();
    *foo
});
```

runtime/src/c2n_bridge.rs
```diff
@@ -0,0 +1,96 @@
+use crate::{LedgerApi, MidnightSystem, Runtime};
+use alloc::vec::Vec;
+use midnight_primitives::{BridgeRecipient, MidnightSystemTransactionExecutor};
+use sp_partner_chains_bridge::{BridgeTransferV1, TransferRecipient};
+
+pub struct MidnightTokenTransferHandler;
+
+/// Storage key for the bridge transfer nonce counter (transient, reset each block).
```
Klapeyron: comment says it is reset each block, but I do not see code doing that

partner-chains/toolkit/bridge/pallet/src/lib.rs
```diff
@@ -324,8 +344,14 @@ pub mod pallet {
...
-			for transfer in transfers {
-				T::TransferHandler::handle_incoming_transfer(transfer);
+			for transfer in transfers.into_iter() {
+				let result = T::TransferHandler::handle_incoming_transfer(transfer.clone());
+				Self::deposit_event(Event::Transfer {
+					mc_tx_hash: transfer.mc_tx_hash,
+					amount: transfer.amount,
+					result,
+					recipient: transfer.recipient,
+				})
```
Klapeyron: it is intentional that we are emitting event when result is None/Error?

partner-chains/toolkit/bridge/pallet/src/lib.rs
```diff
@@ -324,8 +344,14 @@ pub mod pallet {
 			ensure_none(origin)?;
 			ensure!(!InherentExecutedThisBlock::<T>::get(), Error::<T>::InherentAlreadyExecuted);
 			InherentExecutedThisBlock::<T>::put(true);
-			for transfer in transfers {
-				T::TransferHandler::handle_incoming_transfer(transfer);
+			for transfer in transfers.into_iter() {
```
Klapeyron:
```suggestion
			for transfer in transfers {
```

runtime/src/c2m_bridge.rs
```diff
@@ -0,0 +1,96 @@
...
+		sp_core::hashing::blake2_256(&data)
+	}
+}
+
+pub(crate) type MaybeMidnightTxHash = Option<[u8; 32]>;
+
+impl pallet_partner_chains_bridge::TransferHandler<BridgeRecipient, MaybeMidnightTxHash>
+	for MidnightTokenTransferHandler
+{
+	fn handle_incoming_transfer(
```
Klapeyron, line 28: this function has multiple error logs with return of None, shall we make dedicated error for it?

ledger/src/versions/common/mod.rs
LGLO: No way. This PR is first in a series. #1277 is expected to be fixed in a month. We'll fix it later and add tests for invalid transfers in e2e tests and if everything is signed off, then we can enable bridge in environments.

tests/e2e/src/api/midnight.rs
LGLO: Yes, this is C-P. Thanks!

partner-chains/toolkit/bridge/pallet/src/lib.rs
LGLO: It was intentional. This code doesn't know that `result` is `Option` or `Result`. No decision can be made.

Now, I think it will be better if the handler returns `Option` (we won't handle errors anyway). Events will be deposited only for `Some(x)`. This will make Indexer team life easier.

runtime/src/lib.rs
```diff
@@ -118,6 +118,7 @@ pub const SLOTS_PER_EPOCH: u32 = 300;
 
 pub mod authorship;
 pub mod beefy;
+mod c2n_bridge;
```
Klapeyron:
```suggestion
mod c2m_bridge;
```

runtime/src/c2n_bridge.rs
LGLO: This will go to in memory counter per block.

partner-chains/toolkit/bridge/pallet/src/lib.rs
LGLO: Done

partner-chains/toolkit/bridge/pallet/src/lib.rs
LGLO: I needed enumerate(), so this doesn't apply, but you were right.

runtime/src/lib.rs
LGLO: Done

runtime/src/c2m_bridge.rs
LGLO: Handlers should deal with errors internally. As documented in the pallet. In the end I used Option to signal if event should be emitted.

PR #1235, opened 2026-04-03
chore: removed custom pre_dispatch for pallet_midnight

Klapeyron:
# Overview

**Non-functional refactor.** `pallet_midnight` no longer overrides `ValidateUnsigned::pre_dispatch`; it uses the Substrate trait default, which forwards to `validate_unsigned(TransactionSource::InBlock, call)` and maps the result to `Ok(())` / `TransactionValidityError`.

To keep behavior identical to the old custom `pre_dispatch`, `validate_unsigned` now branches on `TransactionSource`: the in-block path uses the real block context and `LedgerApi::validate_guaranteed_execution` (what the removed `pre_dispatch` did); the local/external path keeps the existing pool validation (`LedgerApi::validate_transaction`, adjusted `tblock`, tags, longevity, etc.). No change to what the runtime accepts or rejects—only where that logic lives so the default `pre_dispatch` is correct.

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A


justinfrevert (dismissed):

pallets/midnight/src/lib.rs
```diff
@@ -441,48 +441,51 @@ pub mod pallet {
...
+					let runtime_version = <frame_system::Pallet<T>>::runtime_version().spec_version;
+
+					LedgerApi::validate_guaranteed_execution(
+						&state_key,
+						midnight_tx,
+						block_context,
+						runtime_version,
+					)
+					.map_err(|e| Self::invalid_transaction(e.into()))?;
+					Ok(ValidTransaction::default())
```
justinfrevert: A default transaction would not have the normal transaction tags, which would change the tx pool behavior. Is that okay?

Klapeyron: thanks, yes that will be wrong, changed in https://github.com/midnightntwrk/midnight-node/pull/1235/changes/0cfd08361271c3c6f475e8f601991cb295d0b29a to use txhash

Klapeyron:
Lets make it other way, in current shape removing custom pre_dispatch forces inclusion checks into validate_unsigned(InBlock), which also serves tx-pool prune and weakens tag discovery versus keeping the current split that is on main

PR #1261, opened 2026-04-08
ci(feat): split release workflow into independent node, toolkit, and runtime components

NachoPal:
# Overview

Replace the single node-only boolean in the release workflow with three independent skip flags (`skip-node`, `skip-toolkit`, `skip-runtime`) so any combination of components can be released independently.

## Key changes

- **Separate versioning**: Toolkit images are tagged with the toolkit version from `util/toolkit/Cargo.toml`; runtime releases use `spec_version` from `runtime/src/lib.rs` (e.g. `001_000_000` → `1.0.0`)
- **Dynamic git tags**: Concatenated from included components with suffix at the end (e.g. `node-1.0.0-toolkit-1.0.0-runtime-1.0.0-rc.1`)
- **Per-component release notes**: `changes/` restructured into `node/`, `toolkit/`, and `runtime/` subdirectories; skipped component dirs are removed before `season build` so only relevant sections render
- **Per-component archiving**: `archive-changes` and `reset-changes` run once per included component instead of a single call
- **Per-component version bump**: "Bump components version on main" now bumps node (`Cargo.toml`), toolkit (`Cargo.toml`), and runtime (`spec_version`) independently based on skip flags
- **Release notes header**: Root `index.md` renders Git tag + Components section with icons (📦 Node, 🧰 Toolkit, ⚙️ Runtime); component sections include Docker Images where applicable
- **Security fix**: `${{ inputs.* }}` expressions moved from `run:` blocks to `env:` to prevent shell injection

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [x] I have included a change file, or skipped for this reason: change only affects CI/release workflow
- [x] No new todos introduced

## 🧪 Testing Evidence

- YAML validated with `python3 -c "import yaml; yaml.safe_load(open(...))"` — passes
- Manually traced all 7 release scenarios through the workflow:
  - **Full release** (default): all jobs run, node uses node version, toolkit uses toolkit version
  - **Toolkit-only** (skip-node + skip-runtime): only toolkit image/SBOM/asset jobs run, git tag = `toolkit-X.Y.Z`, release notes from `changes/toolkit/`
  - **Node-only** (skip-toolkit + skip-runtime): only node image/SBOM/asset jobs run
  - **Runtime-only** (skip-node + skip-toolkit): only srtool WASM build runs, no archive PR
  - **All-three-skipped**: rejected by `validate-inputs` job
- Verified `!cancelled() && !failure()` pattern correctly allows downstream jobs when upstream SBOM jobs are skipped
- Verified `inputs.skip-X != true` (step-level) vs `github.event.inputs.skip-X != 'true'` (job-level for reusable workflows) context rules

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [x] Other: CI/release workflow only — no runtime or client changes
- [ ] N/A

## Links

- Depends on season v0.6.0 (`--changes-dir` and `--changes-template-dir` flags)
- Builds on top of #1131 (season-action removal)

### DIFF
--- a/.github/workflows/release-image.yml
+++ b/.github/workflows/release-image.yml
@@ -516,7 +516,7 @@ jobs:
           # Only create individual tags for multi-component releases
           # (single-component releases already have the individual tag as the combined tag)
           if [ "$INCLUDED" -gt 1 ]; then
-            RELEASE_SHA=$(gh api "repos/${GITHUB_REPOSITORY}/git/refs/tags/${GIT_TAG_RELEASE}" -q '.object.sha')
+            RELEASE_SHA="$GITHUB_SHA"
 
             if [ "$SKIP_NODE" != "true" ]; then
               echo "Creating tag node-${IMAGE_TAG_RELEASE}"


gilescope:
will need a changelog entry to point out to people that versioning is changing for toolkit.

.github/workflows/release-image.yml
```diff
@@ -105,154 +132,248 @@ jobs:
...
         env:
-          FORCE: ${{ github.event.inputs.force }}
+          SKIP_NODE: ${{ inputs.skip-node }}
+          SKIP_TOOLKIT: ${{ inputs.skip-toolkit }}
+          SKIP_RUNTIME: ${{ inputs.skip-runtime }}
+          SUFFIX: ${{ github.event.inputs.suffix }}
+        run: |
+          # Build git tag by concatenating included components with bare versions,
+          # then appending suffix once at the end (e.g. node-1.0.0-runtime-1.0.0-rc.1)
+          GIT_TAG=""
```
ozgb, line 178: I like the long tag - could we also push individual git tags too? It makes searching through the tags to find a version of a specific component a lot easier

So e.g. the example commit would be tagged with: `node-1.0.0-runtime-1.0.0-rc.1`, `node-1.0.0`, and `runtime-1.0.0-rc.1`

AGENTS.md
ozgb, line 145: This section of AGENTS.md needs updating

NachoPal:
> If you haven't already, it would be good to create a `-test` pre-release with this workflow to verify it's working and see what the output looks like

Yep, I did multiple times until I got the expected result :) I left a draft release as sample if you wanna check

ozgb:
> > If you haven't already, it would be good to create a `-test` pre-release with this workflow to verify it's working and see what the output looks like
> 
> Yep, I did multiple times until I got the expected result :) I left a draft release as sample if you wanna check

Looks great!

One thing I noticed, runtime asset is the only one not versioned: 
image

That's always been the case, but could be good to fix that in this PR too

PR #1330, opened 2026-04-14
ci: split long running job

gilescope:
# Overview

Splits the `Feature Unification Check` CI job into 2 parallel shards to reduce wall-clock time. Workspace crates are partitioned deterministically by `cksum(name) % SHARDS`, so each shard checks a stable subset.

Also restricts the job to `pull_request` events only — it's skipped in the merge queue and on push to `main` - as long as this passes at the PR level that's enough checks (so total cpu used will be reduced with this change). No cost impact as it's running on github standard runners.

## Changes

- `Earthfile`: `+check-feature-unification` now accepts `--SHARD` / `--SHARDS` args. Default (`SHARDS=1`) preserves existing behaviour (full workspace in one pass). When sharded, uses `cargo metadata` + `jq` to enumerate workspace crates and runs `cargo hack check --no-dev-deps -p <subset>`.
- `.github/workflows/continuous-integration-checks.yml`: matrix over `shard: [0, 1]` with `SHARDS=2`; job name includes shard index; `fail-fast: false` so both shards report independently.

## 📌 Submission Checklist

- [x] Changes are backward-compatible (default `SHARDS=1` is unchanged)
- [ ] Self-reviewed the diff
- [x] I have included a change file, or skipped for this reason: CI-only change
- [x] Update documentation (if relevant) — Earthfile comment documents the new args
- [x] No new todos introduced

## 🧪 Testing Evidence

Observe CI on this PR: two `Feature Unification Check (shard 0/2)` and `(shard 1/2)` jobs run in parallel.

## 🔱 Fork Strategy

- [x] N/A (CI-only)

### DIFF
--- a/.github/workflows/continuous-integration-checks.yml
+++ b/.github/workflows/continuous-integration-checks.yml
@@ -68,12 +68,19 @@ jobs:
           . ./.envrc && earthly --ci +check
 
   feature-unification:
-    name: Feature Unification Check
+    name: Feature Unification Check (shard ${{ matrix.shard }}/2)
+    # Only run at PR time — skipped in merge queue and on push to main.
+    if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
     permissions:
       contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [0, 1]
     env:
       FORCE_COLOR: 1
+      SHARDS: 2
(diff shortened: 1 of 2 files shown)


gilescope:
There's a better way - we can only do the crates that changed their Cargo.tomls

PR #1332, opened 2026-04-14
node 1.0.0 rc2 changes

gilescope:
Merge from main into 1.0.0 branch.

It might seem like there's a few commits but partner chains is in there.
(The unverified sigs is due to this)


runtime/src/lib.rs
cosmir17, line 281: `spec_version` stays at `001_000_000` (same as rc.1), but the extrinsic format has changed:

- `SignedExtra` → `TxExtension` with two new extensions added:
  `frame_system::AuthorizeCall<Runtime>` and `frame_system::WeightReclaim<Runtime>`
- `UncheckedExtrinsic` now parameterised on `TxExtension` instead of `SignedExtra`

If any environment runs rc.1 and upgrades to rc.2 via runtime upgrade (rather than a chain reset), blocks before and after the upgrade would have the same `spec_version` but different extrinsic encoding. The indexer uses `spec_version` to select the runtime module for decoding, so it wouldn't be able to
distinguish the two formats.

Is this intentional? If rc.2 is only ever deployed to fresh chains (or after a reset), this is fine. If there's any chance of a runtime upgrade from rc.1 to rc.2 on a running chain, the spec_version should be bumped.

runtime/src/c2m_bridge.rs
```diff
@@ -0,0 +1,90 @@
+use crate::{LedgerApi, MidnightSystem, Runtime};
```
cosmir17, line 1: The `MidnightTokenTransferHandler` previously just logged bridge transfers.
Now it constructs and executes actual system transactions via the ledger API:

- `construct_distribute_night_cardano_bridge_system_tx` (user transfers)
- `construct_distribute_reserve_system_tx` (reserve)
- `construct_distribute_treasury_system_tx` (invalid recipient fallback)

These system transactions will appear in blocks as `SystemTransactionApplied` events, which the indexer already picks up via `make_block_details`. This unblocks the indexer's bridge protocol work (GH #938-#944) since there's now real on-chain data to index.

cosmir17 (dismissed):
Approved from the indexer perspective. No pallet index changes, runtime API surface the indexer uses (get_contract_state, get_zswap_state_root, etc.) appears unchanged. The extrinsic format change (Comment 1) needs clarification on whether rc.1 → rc.2 runtime upgrades are planned.

runtime/src/lib.rs
gilescope: It's intentional. I will bump `transaction_version` to 3 to indicate this.
(We shouldn't really be upgrading chains from rc.1 to rc.2 as that's not an appropriate test, we should be upgrading from what's on mainnet.)

runtime/src/c2m_bridge.rs
gilescope: The bridge is behind a config flag and is not yet ready to turn on so I'm afraid you'll probably not see said transactions (talk with Lech about how to turn them on locally).

runtime/src/lib.rs
cosmir17: Thanks Giles, transaction_version bump makes sense. Agreed, the relevant upgrade path to test is from what's currently on mainnet to v1.0, not rc.1 to rc.2.

runtime/src/c2m_bridge.rs
cosmir17: Good to know it's behind a config flag. Will coordinate with Lech on how to enable it locally when we start the bridge indexing work (#938-#944).

cosmir17 (approved):

Fentonhaslam (approved):

PR #1357, opened 2026-04-17
chore: remove partner-chains Cargo.toml

LGLO:
# Overview

* Removes unused Cargo.toml and Cargo.lock from partner-chains
* Enable build and test of all crates in partner-chains tree
* partner-chains-demo-runtime isn't build by default in CI
* `cargo build -p partner-chains-demo-node` works OOTB

## 🗹 TODO before merging

- [ ] Decide what to build and what should be removed or intentionally left off build process
- [ ] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A


partner-chains/demo/runtime/src/test_helper_pallet.rs
```diff
@@ -129,7 +129,10 @@ pub mod pallet {
 	}
 
 	impl<T: Config> pallet_partner_chains_bridge::TransferHandler<AccountId, ()> for Pallet<T> {
-		fn handle_incoming_transfer(transfer: BridgeTransferV1<AccountId>) -> () {
+		fn handle_incoming_transfer(
+			_index: u32,
+			transfer: BridgeTransferV1<AccountId>,
+		) -> Option<()> {
```
LGLO, line 135: This is meaningful change: it before this PR partner-chains-demo-runtime was not tested

Cargo.toml
```diff
@@ -28,6 +28,9 @@ members = [
     "tests/e2e",
     "relay",
     "docs",
+    "partner-chains/demo/node",
```
LGLO, line 31: This is added to the workspace and all it local dependencies are now build and tested. This will cost time and resources during each CI cycle. It is time to make decision if we need partner-chains-demo-node and its runtime.

Alternatively we can put pallets here instead and it will prevent the two heavies dependencies to be build. Midnight should never depend on partner-chains-demo-node nor partner-chains-demo-runtime.

I would remove PC demo node and runtime.

partner-chains/demo/node/src/cli.rs
```diff
@@ -63,31 +63,3 @@ pub enum Subcommand {
 	/// Db meta columns information.
 	ChainInfo(sc_cli::ChainInfoCmd),
 }
-
-#[cfg(test)]
-mod registration_signatures_tests {
```
LGLO, line 68: It is removed because it needs binary to be build before running tests. It was causing trouble from time to time in different CI setups. Since partner-chains-demo-node is at most secondary now, I'm not going to pay CI time for this one. Also, there is no business requirements for registration process ATM.

gilescope (approved):

PR #934, opened 2026-03-13
Add cache for multi asset id to avoid excessive joins

justinfrevert:
# Overview

Cache multi_asset.id to avoid excessive joins:

Make the database do less work by pre-caching join data that is fixed. Perf testing has confirmed this improves the query times under load.

Adds an in-memory cache for `multi_asset.id` lookups, replacing repeated `JOIN multi_asset` in db-sync queries with a single cached lookup per (policy, name) pair. This eliminates the `multi_asset` join from registration, deregistration, asset create/spend, and candidate token queries, reducing query complexity and improving observation performance.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [ ] Changes are backward-compatible (or flagged if breaking)
- [ ] Pull request description explains why the change is needed
- [ ] Self-reviewed the diff
- [ ] I have included a change file, or skipped for this reason: 
- [ ] If the changes introduce a new feature, I have bumped the node minor version
- [ ] Update documentation (if relevant)
- [ ] Updated AGENTS.md if build commands, architecture, or workflows changed
- [ ] No new todos introduced

## 🧪 Testing Evidence

Please describe any additional testing aside from CI:

- [ ] Additional tests are provided (if possible)

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A

PR #1365, opened 2026-04-20
Conservative improvements to syncing

gilescope:
# Overview

Back-ports four conservative performance fixes to the cNIGHT db-sync observation path onto `release/node-0.22.5`. Bundles upstream work from #934 plus follow-ups into the release branch.

The four commits:

- **Add cache for multi asset id to avoid excessive joins (#934)** — resolves `(policy, asset_name)` → `multi_asset.ident` once via `MultiAssetCache`, eliminating the `JOIN multi_asset` from every subsequent observation query. Also includes a change file (`changes/changed/cache-multi-asset-id.md`) and tidies candidate-data-source decoding.
- **Increase performance of cNight db-sync queries** — pre-queries coarse `tx` / `tx_out` / `ma_tx_out` id bounds for the block-range window, then constrains the four observation queries (registration / deregistration / asset-create / asset-spend) by primary-key range. Postgres can then prune rows before the expensive joins. Old compile-time sqlx cache entries are replaced.
- **Add `tx_in` indexes as well** — adds the `tx.id` bounding to the `tx_in`-keyed queries (spend / deregistration) so they benefit from the same pruning.

Net effect: significantly lower postgres CPU / query latency during cNIGHT observation, no behavioural change.

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [ ] I have included a change file, or skipped for this reason: `changes/changed/cache-multi-asset-id.md` covers #934; the other three commits are perf fixes to the same path with no user-visible surface change.
- [ ] No new todos introduced

## 🧪 Testing Evidence

Observed reduced db-sync query times on a live cNIGHT observation workload.

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [x] Node Client Update
- [ ] Other:
- [ ] N/A

## Links

Upstream: #934


ozgb (dismissed):

primitives/mainchain-follower/src/data_source/cnight_observation.rs
```diff
@@ -127,13 +133,43 @@ impl MidnightCNightObservationDataSource for MidnightCNightObservationDataSource
...
+				crate::db::get_high_bounds(&self.pool, end.block_number.into())
+					.await
+					.map_err(Into::<Box<dyn std::error::Error + Send + Sync>>::into)
+			},
+			)?;
+		let low_bounds = low_bounds.expect("Start position contains block hash that exists in database");
+		let high_bounds = high_bounds.expect("End position contains block hash that exists in database");
 		// Increment the end position to tx_index + 1 of the current mainchain position
 		let end = end.increment();
+		log::warn!("Bounds:\n{:?}\n{:?}\nfor positions:\n{:?}\n{:?}", low_bounds, high_bounds, start_position, end);
```
cosmir17: Wondering if this should be `debug!` rather than `warn!`? It runs on every observation call and dumps the full bounds, so it might end up flooding prod logs at WARN level. Maybe it was just left in from local debugging?

primitives/mainchain-follower/src/data_source/cnight_observation.rs
```diff
@@ -127,13 +133,43 @@ impl MidnightCNightObservationDataSource for MidnightCNightObservationDataSource
...
+					.map_err(Into::<Box<dyn std::error::Error + Send + Sync>>::into)
+			},
+			async {
+				crate::db::get_high_bounds(&self.pool, end.block_number.into())
+					.await
+					.map_err(Into::<Box<dyn std::error::Error + Send + Sync>>::into)
+			},
+			)?;
+		let low_bounds = low_bounds.expect("Start position contains block hash that exists in database");
+		let high_bounds = high_bounds.expect("End position contains block hash that exists in database");
```
cosmir17, line 169: `get_low_bounds` COALESCEs the outer block to 0 so it always returns `Some`, but `get_high_bounds` doesn't, so it can return `None` if the end block isn't in db-sync. Probably safe today because `get_block_by_hash` on line 151 validates it, but the asymmetry feels fragile. Also on the low side, a missing start block silently returns `QueryBounds{0,...}` and loses all low-bound pruning without signal?

Minor: panic messages say "block hash" but the args are `block_no`?

cosmir17 (dismissed):
LGTM, thanks for pushing through this one. The perf gains look solid.

primitives/mainchain-follower/src/data_source/cnight_observation.rs
gilescope:
```suggestion
		log::debug!("Bounds:\n{:?}\n{:?}\nfor positions:\n{:?}\n{:?}", low_bounds, high_bounds, start_position, end);
```

cosmir17 (approved):

justinfrevert (approved):

primitives/mainchain-follower/src/data_source/cnight_observation.rs
LGLO: `get_high_bounds` doesn't need it because high block always exists. The problem with low is that on some environments the first block has number  `0` it has number `1`.
Alternatively we could redefine `Default` of CardanoPosition to return block_number 1.

> Also on the low side, a missing start block silently returns QueryBounds{0,...}

So in practice it won't happen AFAIK

PR #1366, opened 2026-04-20
baked sync

gilescope:
> **⛔ Do not merge until #1365 has landed.** This branch includes every commit from #1365 plus the file-backed sync work; merging out-of-order will cause conflicts and obscure review.

# Overview

Adds a pre-baked ("snapshot") cNIGHT observation source so a fresh node can skip re-querying db-sync for every Cardano block already observed at snapshot generation time. On top of the DB-query perf work from #1365, this cuts initial-sync time substantially on mainnet/preprod/preview.

## What's in this PR on top of #1365

- **File-backed `CNightObservationDataSource`** (`primitives/mainchain-follower/src/data_source/cnight_observation_file.rs`, ~500 LoC):
  - In-memory snapshot of the four observation event streams (registrations, deregistrations, creates, spends), each sorted by `tx_position`; queries resolved by `partition_point` slice + merge + tx-capacity truncation.
  - Wire format v2: `magic (8) || version (1) || inputs_hash (32) || content_sha256 (32) || zstd(scale(events))`. Content sha256 verified at load; `inputs_hash` binds the file to `(network_magic, policy, asset, validator, auth_token)` and is re-verified against the live pallet config on first call — rejecting cross-network snapshots.
  - Per-call DB traffic reduced to a single `get_block_by_hash(current_tip)` lookup, memoised per distinct tip.

- **Pre-built snapshots** at `res/cnight-observations/{mainnet,preprod,preview}.bin` (≈92 MB mainnet, ≈22 KB preprod, ≈272 KB preview — all zstd-compressed).

- **Generator binary** (`util/cnight-observation-generator/`) — runs the legacy DB-backed path across `[0 .. end_block_no]`, sorts events, and writes a snapshot file. Used offline to produce the shipped `.bin` files.

- **Entrypoint auto-selection** (`node/bin/entrypoint.sh`): maps every known `CFG_PRESET` onto its Cardano network (`mainnet` / `preprod` / `preview`) and sets `CNIGHT_OBSERVATION_FILE` + `CARDANO_NETWORK_MAGIC` accordingly when a matching `.bin` exists. Operator can still override both directly.

- **Runtime switch** in `node/src/main_chain_follower.rs`: if `CNIGHT_OBSERVATION_FILE` is set the file-backed source is used; otherwise the DB-backed source from #1365 is used unchanged.

- **Extra perf polish on the DB path** (also applies when no snapshot is configured):
  - In-process `LastObservation` cache on `MidnightCNightObservationDataSourceImpl` — serves unchanged-tip calls from memory, including a start-position-advanced slice filter for the common case where the pallet walks forward while Cardano tip stays still.
  - `utxo_capacity = tx_capacity * 64` → `* 4`; the pallet post-truncates per-tx anyway, so the old multiplier over-fetched ~16×.
  - Downgraded a noisy per-call bounds log from `warn` → `debug`.
  - Visibility bumps (`pub async fn`) on the four observation query helpers so the generator can drive them.

## Caveats / follow-ups

- The file-backed source has **no built-in fallback** to the DB once live Cardano tip exceeds the snapshot's `end_block_no` — it will silently report "window complete" while real events exist beyond. Safe while the snapshot is regenerated for each release; a hybrid (snapshot-until, DB-after) source is a natural follow-up.
- Snapshot binaries are committed to the repo. If that's undesirable, move them to release-artifact download + entrypoint fetch.

## 🗹 TODO before merging

- [ ] #1365 merged first
- [ ] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [ ] I have included a change file, or skipped for this reason: see `changes/changed/cache-multi-asset-id.md` (from #1365), `changes/changed/cnight-query-coarse-bounds.md` (from #1365), and the `feat: add changefile` commit on this branch for the baked-sync entry.
- [ ] No new todos introduced

## 🧪 Testing Evidence

- Generated mainnet/preprod/preview snapshots via `util/cnight-observation-generator` and loaded them on boot — `inputs_hash` verification passed, total-events logged.
- Ran node with and without `CNIGHT_OBSERVATION_FILE` set; confirmed the file-backed path serves observations without hitting db-sync beyond the per-tip `get_block_by_hash`.

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [x] Node Client Update
- [ ] Other:
- [ ] N/A

## Links

Depends on: #1365
Upstream DB-perf origin: #934


node/src/main_chain_follower.rs
```diff
@@ -220,11 +220,42 @@ pub async fn create_cached_data_sources(
 
 	let cnight_observation_pool =
 		get_connection(postgres_uri, CNIGHT_OBSERVATION_POOL_CFG, cfg.allow_non_ssl).await?;
-	let cnight_observation = MidnightCNightObservationDataSourceImpl::new(
-		cnight_observation_pool,
-		metrics_opt.clone(),
-		1000,
-	);
+	let cnight_observation: Arc<dyn MidnightCNightObservationDataSource + Send + Sync> =
+		if let Ok(path) = std::env::var("CNIGHT_OBSERVATION_FILE") {
```
ozgb, line 224: Suggest instead we use the existing config mechanism for this - named something like `CNIGHT_OBSERVATION_SNAPSHOT`

Then we can add the correct paths in the `cfg/*.toml` files

This would remove the need for special-case config in `entrypoint.sh`

util/cnight-observation-generator/src/main.rs
ozgb, line 1: Could be cleaner to add this as a node subcommand instead?

primitives/mainchain-follower/src/data_source/cnight_observation_file.rs
```diff
@@ -0,0 +1,500 @@
...
+	/// nodes refuse the file unless their own inputs produce the same hash.
+	/// `content_sha256` covers the compressed payload so integrity can be
+	/// verified before any decompression work.
+	pub fn write_to_path(&self, path: &Path, inputs_hash: [u8; 32]) -> std::io::Result<()> {
+		let payload = self.encode();
+		let compressed = zstd::bulk::compress(&payload, 19)
+			.map_err(|e| std::io::Error::other(format!("zstd compress: {e}")))?;
+		let content_hash: [u8; 32] = Sha256::digest(&compressed).into();
+		let mut out = Vec::with_capacity(HEADER_LEN + compressed.len());
+		out.extend_from_slice(SNAPSHOT_MAGIC);
```
ozgb, line 110: Custom serialization - consider using a struct with some serialization derive instead (`midnight-serialize`?)

primitives/mainchain-follower/src/data_source/cnight_observation_file.rs
```diff
@@ -0,0 +1,500 @@
...
+	}
+
+	/// Read from disk, verifying magic + version + content sha256. Does **not**
+	/// verify `inputs_hash` — the caller must compare `LoadedSnapshot::inputs_hash`
+	/// against its own `SnapshotInputs::hash` before using the snapshot.
+	pub fn read_from_path(
+		path: &Path,
+	) -> Result<LoadedSnapshot, Box<dyn std::error::Error + Send + Sync>> {
+		let bytes = std::fs::read(path)?;
+		if bytes.len() < HEADER_LEN {
```
ozgb, line 125: Same here - could use a struct with derive `Serialize` and a validate() method

gilescope:
This has been replaced by https://github.com/midnightntwrk/midnight-node/pull/1436

PR #1393, opened 2026-04-23
feat: handling of subminimal transfers

### LINKED ISSUE #1248: [C-to-M Bridge] Accumulate too small transactions
If we have a lot of too small transfers (invalid because they are below given threshold), store them in pallet unless they sum together above the threshold, so then we transfer sum to treasury.

Implement e2e-test using framework developed in #1291

LGLO:
# Overview

Implements logic of subminmal transfers handling in Cardano-to-Midnight bridge pallet

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A


pallets/c2m-bridge/src/tests.rs
```diff
@@ -82,3 +91,37 @@ fn nonce_influences_addressed_transfers() {
...
+		assert!(mock_pallet::Transfers::<Test>::get().is_empty());
+		assert!(frame_system::Pallet::<Test>::events().is_empty());
+		//180
+		C2MBridge::handle_incoming_transfer(subminimal_transfer());
+		assert_eq!(pallet::SubminimalTransfersSum::<Test>::get(), 180);
+		assert!(mock_pallet::Transfers::<Test>::get().is_empty());
+		//270 > 250. Should flush everything in one transfer.
+		C2MBridge::handle_incoming_transfer(subminimal_transfer());
+		assert_eq!(pallet::SubminimalTransfersSum::<Test>::get(), 0);
+		assert_eq!(mock_pallet::Transfers::<Test>::get().len(), 1);
```
Klapeyron, line 120: should we also check the value inside the transfer?

pallets/c2m-bridge/src/lib.rs
```diff
@@ -228,7 +258,34 @@ pub mod pallet {
...
+						Self::execute_transfer(
+							counter,
+							BridgeTransferV1::new_invalid(transfer.mc_tx_hash, subminimal_sum),
+						);
+						SubminimalTransfersSum::<T>::put(transfer.amount);
+					},
+				}
+			} else {
+				Self::execute_transfer(counter, transfer);
+			}
```
Klapeyron: with early return and conditional match it can be slightly flattened, POC:
```suggestion
			if transfer.amount >= Self::minimal_transfer_amount() {
				Self::execute_transfer(counter, transfer);
				return;
			}

			let subminimal_sum = SubminimalTransfersSum::<T>::get();
			let flush_threshold =
				SubminimalTransfersConfiguration::<T>::get().subminimal_transfers_flush_threshold;

			let (updated_sum, maybe_transfer) = match subminimal_sum.checked_add(transfer.amount) {
				// Keep accumulating until the buffered amount is large enough to flush.
				Some(accumulated_sum) if accumulated_sum <= flush_threshold => {
					(accumulated_sum, None)
				},
				// Flush the buffered sum as a single invalid transfer once the threshold is exceeded.
				Some(accumulated_sum) => {
					(0, Some(BridgeTransferV1::new_invalid(transfer.mc_tx_hash, accumulated_sum)))
				},
				// On overflow, flush the previous buffer and restart accumulation from this transfer.
				None => (
					transfer.amount,
					Some(BridgeTransferV1::new_invalid(transfer.mc_tx_hash, subminimal_sum)),
				),
			};

			SubminimalTransfersSum::<T>::put(updated_sum);

			if let Some(transfer) = maybe_transfer {
				Self::execute_transfer(counter, transfer);
			}
```

pallets/c2m-bridge/src/lib.rs
```diff
@@ -31,6 +31,19 @@ pub use pallet::*;
 /// Hash of a Midnight ledger transaction, returned by the system transaction executor.
 pub type MidnightTxHash = [u8; 32];
 
+/// Amount of STARS. One NIGHT is 10^6 STARS.
+pub struct Stars(pub u128);
```
Klapeyron: From may be more idiomatic here in place of pub internal

pallets/c2m-bridge/src/lib.rs
```diff
@@ -144,6 +163,17 @@ pub mod pallet {
...
 		}
 
+		/// Returns the minimum bridge transfer amount, read from ledger parameters.
+		/// Falls back to 0 if the ledger parameter cannot be read.
+		fn minimal_transfer_amount() -> u64 {
+			T::MinBridgeAmountProvider::get_c_to_m_bridge_min_amount()
+				.map(|v| v.to_cnight())
+				.unwrap_or_else(|e| {
+					log::error!("Failed to read c_to_m_bridge_min_amount from ledger: {e:?}");
+					0
```
Klapeyron: if reading of ledger parameters will fail, returned 0 here may bypass the subminimal accumulation?

pallets/c2m-bridge/src/lib.rs
```diff
@@ -31,6 +31,19 @@ pub use pallet::*;
 /// Hash of a Midnight ledger transaction, returned by the system transaction executor.
 pub type MidnightTxHash = [u8; 32];
 
+/// Amount of STARS. One NIGHT is 10^6 STARS.
+pub struct Stars(pub u128);
+
+impl Stars {
+	fn to_cnight(&self) -> u64 {
+		// 'as u64' is safe because cNight supply is 24*10^9.
+		(self.0 / STARS_PER_NIGHT) as u64
```
Klapeyron: here may happen the rounding if self.0 is not a multiplication of STARS_PER_NIGHT

LGLO: Replaced division with multiplication. No longer applies.

pallets/c2m-bridge/src/lib.rs
LGLO: Ok

pallets/c2m-bridge/src/lib.rs
LGLO: Then we make each subminimal transfer a ledger operation, but this isn't most expensive ledger operation.
Also, if we can't read params I expect problems with creating or executing any transaction.

pallets/c2m-bridge/src/tests.rs
LGLO: I didn't go that far, because it would require deserialization of the transfer. Checking amount in the event is enough for me.

pallets/c2m-bridge/src/lib.rs
LGLO: The logic is going to be more complicated, so I'll extract functions.

pallets/c2m-bridge/src/lib.rs
```diff
@@ -31,6 +31,25 @@ pub use pallet::*;
 /// Hash of a Midnight ledger transaction, returned by the system transaction executor.
 pub type MidnightTxHash = [u8; 32];
 
+/// Amount of STARS. One NIGHT is 10^6 STARS.
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Stars(u128);
+
+impl Stars {
+	fn from_cnight(c_night: u64) -> Stars {
```
Klapeyron: it is infallible now, can we add some compile-time check to keep that invariant not violated in the future?

```rust
// Build-time proof: `from_cnight` must handle the largest bridge `amount` without `u128` overflow.
const _: Stars = Stars::from_cnight(u64::MAX);
```

pallets/c2m-bridge/src/lib.rs
```diff
@@ -144,17 +203,22 @@ pub mod pallet {
...
 		}
 
 		/// Generate a deterministic unique nonce for a bridge transfer.
 		///
 		/// Uses the parent hash (unique per block) combined with an
 		/// increasing counter (unique within a block) to guarantee uniqueness.
-		fn generate_nonce(counter: u32) -> [u8; 32] {
+		fn generate_nonce() -> [u8; 32] {
+			let counter = TransferCounter::<T>::get();
+			TransferCounter::<T>::put(counter + 1);
```
Klapeyron, line 222: there is mutate available, but I am not convinced it is more readable than existing version:

```suggestion
			let counter = TransferCounter::<T>::mutate(|c| {
				let prev = *c;
				*c = prev + 1;
				prev
			});
```

pallets/c2m-bridge/src/mock.rs
```diff
@@ -36,6 +40,13 @@ pub mod mock_pallet {
 			Ok([count; 32])
 		}
 	}
+
+	impl<T> MinBridgeAmountProvider for Pallet<T> {
+		// Returns value in STARS, pallet denominates it to cNIGHT
+		fn get_c_to_m_bridge_min_amount() -> Result<Stars, LedgerApiError> {
+			Ok(Stars(STARS_PER_NIGHT * u128::from(TEST_MINIMAL_TRANSFER_CNIGHT)))
```
Klapeyron:
```suggestion
			Ok(Stars::from_cnight(TEST_MINIMAL_TRANSFER_CNIGHT))
```

pallets/c2m-bridge/src/lib.rs
```diff
@@ -163,72 +227,112 @@ pub mod pallet {
...
-				},
-				Err(e) => {
-					log::error!("Failed to construct tx for '{transfer:?}': {e}");
-					None
-				},
+		fn handle_subminimal_transfer(transfer: BridgeTransferV1<BridgeRecipient>) {
+			let (count, accumulated) = SubminimalTransfers::<T>::get();
+			let config = SubminimalTransfersConfiguration::<T>::get();
+
+			// + is safe, because all existing cNight fits in u64.
```
Klapeyron: maybe we can make similar trick with u128 as two u64s can't overflow u128

pallets/c2m-bridge/src/mock.rs
Klapeyron: also maybe after that `STARS_PER_NIGHT` does not need to be public anymore

pallets/c2m-bridge/src/lib.rs
Klapeyron: but in this scenario, if we return 0 -> may it stop the aggregation, so every tiny transfer can trigger one system transaction each?

pallets/c2m-bridge/src/lib.rs
```diff
@@ -144,17 +203,22 @@ pub mod pallet {
...
-		fn next_counter() -> u32 {
-			let counter = TransferCounter::<T>::get();
-			TransferCounter::<T>::put(counter + 1);
-			counter
+		/// Returns the minimum bridge transfer amount, read from ledger parameters.
+		/// Falls back to 0 if the ledger parameter cannot be read.
+		fn minimal_transfer_amount() -> Stars {
+			T::MinBridgeAmountProvider::get_c_to_m_bridge_min_amount().unwrap_or_else(|e| {
+				log::error!("Failed to read c_to_m_bridge_min_amount from ledger: {e:?}");
+				Stars(0)
```
Klapeyron: we should not suppress that error here, it shall be rather handled in the caller code by `unwrap_or` or similar function, but still I think we should handle that separately with some action

LGLO: This is private function. This is the place where I intentionally `unwrap`. I could inline this code but it is less readable.

pallets/c2m-bridge/src/lib.rs
LGLO: I think get and put are ok. Storage is killed anyway, so there is only one read and then puts are not hashing anyway, so it is as costly as mutate, but simpler.

pallets/c2m-bridge/src/lib.rs
LGLO: Then I have `new_sum` as `u128`. Either I can't safely put it into `u64` storage or take it out as `u64` (if I make storage `u64`).

pallets/c2m-bridge/src/mock.rs
LGLO: Done. I wanted to keep `STARS_PER_NIGHT` public, because it is well known domain fact, but domain doesn't specify the type of it, so there is a reason to hide it. I'll go with pub(crate).

pallets/c2m-bridge/src/lib.rs
LGLO: Done

pallets/c2m-bridge/src/lib.rs
LGLO: Yes, this is decision, what to do is such a case.

pallets/c2m-bridge/src/lib.rs
LGLO: inlined at call site

pallets/c2m-bridge/src/lib.rs
```diff
@@ -163,72 +227,117 @@ pub mod pallet {
...
+					recipient.as_bytes(),
+					nonce,
+				),
+				|midnight_tx_hash| Event::UserTransfer {
+					mc_tx_hash,
+					amount,
+					recipient,
+					midnight_tx_hash,
+				},
+				&alloc::format!("'Reserve' transfer of {} from Cardano Tx: {}", amount, mc_tx_hash),
```
cosmir17: Copy-paste from `handle_reserve_transfer`: this log says "Reserve" but the function handles user transfers.

Suggested change? :
```diff
-            &alloc::format!("'Reserve' transfer of {} from Cardano Tx: {}", amount, mc_tx_hash),
+            &alloc::format!("'User' transfer of {} from Cardano Tx: {}", amount, mc_tx_hash),

File: pallets/c2m-bridge/src/lib.rs [hunk 12]
```diff
@@ -163,72 +227,117 @@ pub mod pallet {
...
+			F: FnOnce([u8; 32]) -> Event<T>,
+		{
+			match result {
+				Ok(serialized_tx) => {
+					log::debug!("Serialized transaction for {}", description);
+					match T::MidnightSystemTransactionExecutor::execute_system_transaction(
+						serialized_tx.clone(),
+					) {
+						Ok(tx_hash) => {
+							log::debug!("Serialized transaction for {}", description);
```
cosmir17: Same message as line 239. This one fires after successful execution, suggest:
```diff
-                        log::debug!("Serialized transaction for {}", description);
+                        log::debug!("Executed system transaction for {}", description);

File: pallets/c2m-bridge/src/lib.rs [hunk 13]
```diff
@@ -163,72 +227,117 @@ pub mod pallet {
...
+					LedgerApi::construct_distribute_treasury_system_tx(sum.into()),
+					|midnight_tx_hash| Event::SubminimalFlushTransfer {
+						amount: sum,
+						count,
+						midnight_tx_hash,
+					},
+					&alloc::format!("subminimal transfers flush of total {}", sum),
+				);
+			} else {
+				SubminimalTransfers::<T>::put(SubminimalTransfersState { count, sum });
```
cosmir17, line 282: Kill (SubminimalTransfers::<T>::kill();) runs before execute. If execute fails, no event is emitted but the accumulator is gone, so the indexer never sees the flushed amount.

Either document the trade-off, or move the kill into the success path of the closure?:

```rust
Self::execute_serialized_tx(
    LedgerApi::construct_distribute_treasury_system_tx(sum.into()),
    |midnight_tx_hash| {
        // Side effect: only on the success path. Reset the accumulator only
        // when the flush has actually been recorded on the ledger.
        SubminimalTransfers::<T>::kill();
        Event::SubminimalFlushTransfer { amount: sum, count, midnight_tx_hash }
    },
    ...
);
```

pallets/c2m-bridge/src/lib.rs
```diff
@@ -163,72 +227,117 @@ pub mod pallet {
...
-					log::error!("Failed to construct tx for '{transfer:?}': {e}");
-					None
-				},
+		fn handle_subminimal_transfer(transfer: BridgeTransferV1<BridgeRecipient>) {
+			let SubminimalTransfersState { count, sum } = SubminimalTransfers::<T>::get();
+			let config = SubminimalTransfersConfiguration::<T>::get();
+
+			// Safe, because all existing cNight fits in u64.
+			let sum = sum.saturating_add(transfer.amount);
+			let count = count + 1;
```
cosmir17: `sum` uses `saturating_add`, `count` uses `+ 1`. Inconsistent. The u32 case is unreachable in
practice but for free consistency?:
```diff
-            let count = count + 1;
+            let count = count.saturating_add(1);
```

pallets/c2m-bridge/src/lib.rs
```diff
@@ -163,72 +227,117 @@ pub mod pallet {
...
+			make_event: F,
+			description: &str,
+		) where
+			F: FnOnce([u8; 32]) -> Event<T>,
+		{
+			match result {
+				Ok(serialized_tx) => {
+					log::debug!("Serialized transaction for {}", description);
+					match T::MidnightSystemTransactionExecutor::execute_system_transaction(
+						serialized_tx.clone(),
```
cosmir17: `serialized_tx` is not used after this call? the clone seems a redundant `Vec<u8>` allocation.

Suggested change:
```diff
-                match T::MidnightSystemTransactionExecutor::execute_system_transaction(
-                    serialized_tx.clone(),
-                ) {
+                match T::MidnightSystemTransactionExecutor::execute_system_transaction(serialized_tx) {
```

pallets/c2m-bridge/src/lib.rs
```diff
@@ -163,72 +227,117 @@ pub mod pallet {
...
 
 	impl<T: Config> pallet_partner_chains_bridge::TransferHandler<BridgeRecipient> for Pallet<T> {
 		fn handle_incoming_transfer(transfer: BridgeTransferV1<BridgeRecipient>) {
-			let counter = Self::next_counter();
-			Self::execute_transfer(counter, transfer);
+			let amount = transfer.amount;
+			let minimal_transfer_amount =
+				T::MinBridgeAmountProvider::get_c_to_m_bridge_min_amount().unwrap_or_else(|e| {
+					log::error!("Failed to read c_to_m_bridge_min_amount from ledger: {e:?}");
+					Stars(0)
```
cosmir17: This is the `unwrap_or Stars(0)` you and Klapeyron discussed. Worth a one-line comment for future
readers explaining the chosen behaviour?, since the consequence (subminimum aggregation bypassed
under ledger read failure) is not obvious:

```rust
        log::error!("Failed to read c_to_m_bridge_min_amount from ledger: {e:?}");
        // Treat the threshold as 0 on read failure so all transfers fall through to
        // per-transfer handling rather than aggregating with stale data. Wider ledger
        // failures will surface separately.
        Stars(0)
```

pallets/c2m-bridge/src/lib.rs
```diff
@@ -163,72 +227,117 @@ pub mod pallet {
...
+				});
```
cosmir17: `get_c_to_m_bridge_min_amount` is read inside `handle_incoming_transfer`, which fires once per
inbound transfer? can there be a better algorithm?

pallets/c2m-bridge/src/lib.rs
```diff
@@ -31,6 +31,34 @@ pub use pallet::*;
...
+}
+
+impl From<u128> for Stars {
+	fn from(value: u128) -> Self {
+		Self(value)
+	}
+}
+
+/// Domain constant
+pub(crate) const STARS_PER_NIGHT: u128 = 1_000_000;
```
cosmir17, line 54: "Domain constant" alone doesn't say what it represents. The midnight-ledger spec
(`midnight-ledger/spec/dust.md`) defines this relationship verbatim:

> The atomic unit of Night is the Star, with 1 Night = 10^6 Stars

Suggested (spec wording, with the codebase's caps convention from the existing `Stars` struct
doc on line 34):

```diff
-/// Domain constant
+/// The atomic unit of NIGHT is the STAR, with 1 NIGHT = 10^6 STARS.
 pub(crate) const STARS_PER_NIGHT: u128 = 1_000_000;
```

cosmir17 (commented):
Posting line comments and nits as a non-blocking review. @Klapeyron has the merge call as code owner.

cosmir17:
closing and re-opening PR as ci is not running.

pallets/c2m-bridge/src/lib.rs
LGLO: Thanks, fixed.

pallets/c2m-bridge/src/lib.rs
LGLO: :+1:

pallets/c2m-bridge/src/lib.rs
LGLO: I'll use saturating_add for both. I think `+` is enough, because in practice `sum` can't overflow as well.

pallets/c2m-bridge/src/lib.rs
LGLO: Only if the bridge-pallet API changes to handling all transfers at once.

pallets/c2m-bridge/src/lib.rs
LGLO: Edited this comment.

pallets/c2m-bridge/src/lib.rs
LGLO: Commented

pallets/c2m-bridge/src/lib.rs
LGLO: Fixed

pallets/c2m-bridge/src/lib.rs
LGLO: You're right with premature storage kill. I moved it after execute. `deposit_event` is infallible, so if transaction is executed, there is nothing left to fail.
I don't want to call `kill` from a closure meant for construction of the event.

pallets/c2m-bridge/src/lib.rs
```diff
@@ -163,72 +228,125 @@ pub mod pallet {
...
-			let counter = Self::next_counter();
-			Self::execute_transfer(counter, transfer);
+			let amount = transfer.amount;
+
+			let minimal_transfer_amount =
+				T::MinBridgeAmountProvider::get_c_to_m_bridge_min_amount().unwrap_or_else(|e| {
+					// If ledger read fails, then subminimal transfers functionality is bypassed.
+					// Most likely, if ledger reads fail, the code will never succeed making a transaction.
+					log::error!("Failed to read c_to_m_bridge_min_amount from ledger: {e:?}");
+					Stars(0)
```
Klapeyron: consider handling of this case explicit in a match
```rust
		fn dispatch_standard_incoming_transfer(transfer: BridgeTransferV1<BridgeRecipient>) {
			let mc_tx_hash = transfer.mc_tx_hash;
			let amount = transfer.amount;
			match transfer.recipient {
				TransferRecipient::Invalid => Self::handle_invalid_transfer(mc_tx_hash, amount),
				TransferRecipient::Reserve => Self::handle_reserve_transfer(mc_tx_hash, amount),
				TransferRecipient::Address { recipient } => {
					Self::handle_user_transfer(mc_tx_hash, amount, recipient)
				},
			}
		}
	}

	impl<T: Config> pallet_partner_chains_bridge::TransferHandler<BridgeRecipient> for Pallet<T> {
		fn handle_incoming_transfer(transfer: BridgeTransferV1<BridgeRecipient>) {
			let amount = transfer.amount;

			match T::MinBridgeAmountProvider::get_c_to_m_bridge_min_amount() {
				Ok(minimal_transfer_amount) if Stars::from_cnight(amount) < minimal_transfer_amount => {
					Self::handle_subminimal_transfer(transfer)
				},
				Ok(_minimal_transfer_amount) => {
					Self::dispatch_standard_incoming_transfer(transfer)
				},
				Err(e) => {
					log::error!("Failed to read c_to_m_bridge_min_amount from ledger: {e:?}");
					Self::dispatch_standard_incoming_transfer(transfer)
				},
			}
		}
	}
```

Klapeyron (approved):

cosmir17:
As Klapeyron approved, I too approve this PR 🙏  Thanks for your work.

PR #1403, opened 2026-04-23
fix: try deadlock fix

gilescope:
Trial of ledger fix.

UPDATE: Fixes https://github.com/midnightntwrk/midnight-node/issues/1352

- [ ] Need more official ledger release for storage crate


gilescope:
(worked, doing official pr)

PR #1432, opened 2026-04-27
fix(cnight-observation): bound IDP queries by CardanoBlockWindowSize

skylar-simoncelli:
# Overview

The cNight observation Inherent Data Provider was querying cardano-db-sync from `NextCardanoPosition` all the way to the current Cardano tip on every Midnight block. For sparse assets like cNight — which appear in only a tiny fraction of Cardano outputs — this caused db-sync queries that effectively scanned the entire Cardano history on every 6-second Midnight block, blocking block import to ~0.1 bps whenever the IDP's start position fell behind tip.

`CardanoBlockWindowSize` already exists as runtime storage on `pallet-cnight-observation` (default 1000) and is already exposed via `CNightObservationApi::get_cardano_block_window_size`, but the IDP never actually read it. The "window" was effectively `[start, ∞)`. This PR finishes that wiring.

## Change

1. `MidnightCNightObservationInherentDataProvider::new` reads `block_window_size` from the runtime API.
2. `MidnightCNightObservationDataSource::get_utxos_up_to_capacity` takes a new `block_window_size: u32` parameter.
3. The data source clamps the query upper bound to `start + block_window_size`. When tip is closer than `start + window`, behavior is unchanged.
4. A new `db::get_block_by_block_no` helper is added to fetch the bounded block for the upper bound.
5. The genesis-creation tool passes `u32::MAX` to preserve its existing full-history scan (it's a one-time, off-the-hot-path job).
6. The previously-unused `DEFAULT_CARDANO_BLOCK_WINDOW_SIZE` constant in the IDP module is removed (the value now comes from the runtime API).

The cap on `end` also tightens the `get_high_bounds` integer-id ranges that constrain `tx`/`tx_out`/`ma_tx_out`/`tx_in` joins, so each query benefits twice from the smaller window.

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

- `SQLX_OFFLINE=true cargo check -p midnight-primitives-mainchain-follower` — clean
- `SKIP_WASM_BUILD=1 SQLX_OFFLINE=true cargo check -p midnight-node` — clean
- `SKIP_WASM_BUILD=1 SQLX_OFFLINE=true cargo clippy -p midnight-primitives-mainchain-follower -p midnight-node` — clean
- `cargo fmt -p midnight-primitives-mainchain-follower -p midnight-node` — no diff

Functional verification on a node that's currently stuck at 0.0–0.1 bps with ~118s db-sync queries: expect query latency to drop to sub-second and `bps` to recover to normal AURA cadence once a node running this build replaces the existing one. The window-size pallet storage is already populated at default 1000 across networks; no chain-spec or genesis change required.

## 🔱 Fork Strategy

This is a node-client behavior change only — the runtime API surface is unchanged (the `get_cardano_block_window_size` runtime API already exists and was already being implemented; only its consumer changed). Old nodes will continue to function (with the existing slow behavior); new nodes will use the bounded window.

- [ ] Node Runtime Update
- [x] Node Client Update
- [ ] Other: 
- [ ] N/A

## Links

### DIFF
--- a/primitives/mainchain-follower/src/data_source/cnight_observation.rs
+++ b/primitives/mainchain-follower/src/data_source/cnight_observation.rs
@@ -59,6 +59,8 @@ pub struct TxPosition {
 pub enum MidnightCNightObservationDataSourceError {
 	#[error("missing reference for block hash `{0}` in db-sync")]
 	MissingBlockReference(McBlockHash),
+	#[error("missing reference for block number `{0}` in db-sync")]
+	MissingBlockNumber(u32),
 	#[error("Error querying database: {0}")]
 	DBQueryError(#[from] sqlx::error::Error),
 	#[error("Error extracting network id from Cardano address")]
@@ -110,6 +112,7 @@ impl MidnightCNightObservationDataSource for MidnightCNightObservationDataSource
 		start_position: &CardanoPosition,
 		current_tip: McBlockHash,
 		tx_capacity: usize,
+		block_window_size: u32,
 	) -> Result<ObservedUtxos, Box<dyn std::error::Error + Send + Sync>> {
 		let cnight_asset_name = config.cnight_asset_name.as_bytes();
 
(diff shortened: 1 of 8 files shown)


primitives/mainchain-follower/src/idp/cnight_observation.rs
```diff
@@ -105,6 +103,7 @@ impl MidnightCNightObservationInherentDataProvider {
 			.try_into()
 			.map_err(|_| IDPCreationError::AuthTokenAssetNameNotString)?;
 		let cardano_position_start = api.get_next_cardano_position(parent_hash)?;
+		let block_window_size = api.get_cardano_block_window_size(parent_hash)?;
```
LGLO, line 106: This is possibly breaking change.

If the runtime api of mainnet didn't have this method it will fail when syncing chain from genesis.
Otherwise, if the runtime api had this function, new node, when syncing from genesis, will apply this limit and return possibly different result what chain has observed without this addition.

`ObservedUtxo` contains bounds. So, if it happened on the `mainnet` that CNightObservation returned `observed_utxos` spanning more than 10000 blocks, it won't sync.

I think that solution here is (unfortunately) to use more complex code. `block_window_size` usage should be enabled conditionally. Condition can be presence of a runtime API, but a new one, not one that existed before and wasn't used.

primitives/mainchain-follower/src/data_source/cnight_observation.rs
```diff
@@ -153,12 +156,30 @@ impl MidnightCNightObservationDataSource for MidnightCNightObservationDataSource
...
-		let end: CardanoPosition = crate::db::get_block_by_hash(&self.pool, current_tip.clone())
+		let tip: CardanoPosition = crate::db::get_block_by_hash(&self.pool, current_tip.clone())
 			.await?
 			.ok_or(MidnightCNightObservationDataSourceError::MissingBlockReference(current_tip))?
 			.into();
 		drop(_block_timer);
 
+		// Clamp the upper query bound to `start + block_window_size`. Without this, every
+		// per-block invocation scans from `start_position` to the current Cardano tip, which
+		// for sparse assets (e.g. cNight) means scanning the whole chain on every Midnight
```
LGLO, line 167: `start` advances every Cardano block and being sparse doesn't matter here.
This code "fixes" only one case of the first cNIGHT observation queries, when `start` is somewhere in distant past (it is defined in configuration) and `end` (renamed to `tip`) is the highest stable block.

PR #1433, opened 2026-04-27
fix(cnight-observation): bound IDP queries by CardanoBlockWindowSize (0.22.5 backport)

skylar-simoncelli:
# Overview

Backport of #1432 to `release/node-0.22.5`.

The cNight observation Inherent Data Provider was querying cardano-db-sync from `NextCardanoPosition` all the way to the current Cardano tip on every Midnight block. For sparse assets like cNight — which appear in only a tiny fraction of Cardano outputs — this caused db-sync queries that effectively scanned the entire Cardano history on every 6-second Midnight block, blocking block import to ~0.1 bps whenever the IDP's start position fell behind tip.

`CardanoBlockWindowSize` already exists as runtime storage on `pallet-cnight-observation` (default 1000) and is already exposed via `CNightObservationApi::get_cardano_block_window_size`, but the IDP never actually read it. The "window" was effectively `[start, ∞)`. This PR finishes that wiring on the 0.22.5 release branch.

## Change

1. `MidnightCNightObservationInherentDataProvider::new` reads `block_window_size` from the runtime API.
2. `MidnightCNightObservationDataSource::get_utxos_up_to_capacity` takes a new `block_window_size: u32` parameter.
3. The data source clamps the query upper bound to `start + block_window_size`. When tip is closer than `start + window`, behavior is unchanged.
4. A new `db::get_block_by_block_no` helper is added (uses runtime-checked `sqlx::query_as` so it doesn't require regenerating the offline sqlx cache).
5. The genesis-creation tool passes `u32::MAX` to preserve its existing full-history scan.
6. The previously-unused `DEFAULT_CARDANO_BLOCK_WINDOW_SIZE` constant in the IDP module is removed.

The cap on `end` also tightens the `get_high_bounds` integer-id ranges that constrain `tx`/`tx_out`/`ma_tx_out`/`tx_in` joins, so each query benefits twice from the smaller window.

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

- `SQLX_OFFLINE=true cargo check -p midnight-primitives-mainchain-follower` — clean
- `SKIP_WASM_BUILD=1 SQLX_OFFLINE=true cargo check -p midnight-node` — clean
- `cargo fmt -p midnight-primitives-mainchain-follower -p midnight-node` — no diff

Functional verification: replace the running node binary with this build on a node that's currently stuck at 0.0–0.1 bps with ~118s db-sync queries. Expect query latency to drop to sub-second and `bps` to recover to normal AURA cadence.

## 🔱 Fork Strategy

Node-client behavior change only — the runtime API surface is unchanged (the `get_cardano_block_window_size` API already exists in 0.22.5 and is implemented; only its consumer changed). Old nodes continue to function with the existing slow behavior; new nodes use the bounded window.

- [ ] Node Runtime Update
- [x] Node Client Update
- [ ] Other: 
- [ ] N/A

## Links

- main PR: #1432

### DIFF
--- a/changes/changed/bound-cnight-observation-window.md
+++ b/changes/changed/bound-cnight-observation-window.md
@@ -1,0 +1,10 @@
+#node #cnight-observation #performance
+# Bound cNight observation queries by `CardanoBlockWindowSize`
+
+The cNight observation IDP previously queried cardano-db-sync from `NextCardanoPosition` all the way to the current Cardano tip on every Midnight block. For sparse assets like cNight, this resulted in multi-minute db-sync queries scanning effectively the full Cardano history, blocking block import to ~0.1 bps.
+
+The `CardanoBlockWindowSize` storage value already existed and was already exposed via `CNightObservationApi::get_cardano_block_window_size`, but was unused by the IDP. This change reads the window size in the IDP, threads it through `MidnightCNightObservationDataSource::get_utxos_up_to_capacity`, and clamps the query upper bound to `start + window` in the data source. Each query now scans a bounded window (default 1000 Cardano blocks) instead of the full chain. The genesis-creation tool passes `u32::MAX` to preserve its existing full-history scan.
+
+Backport of #1432 to release/node-0.22.5.
+
+PR:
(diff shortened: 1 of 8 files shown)


gilescope:
0.22.5 is already released.

PR #1443, opened 2026-04-29
fix: stop persisting intermediate ledger states

### LINKED ISSUE #1442: Ledger Storage: Stop persisting intermediate states
With the first implementation of ledger storage, we’re persisting all states. This includes intermediate states that we create in the process of importing or otherwise constructing a block.

We should ensure that these states are not kept in the DB, while also not accidentally dropping states that we need.

QA Approach:
- Verify that integration and unit tests are added that cover this behaviour:
  - Ledger states referenced from all blocks remain valid
  - Ledger states created as intermediate states while constructing a block are not persisted
- Executing multiple transactions per block should result in fewer ledger states committed to the backend DB than before this change.

ozgb:
# Overview

Resolves https://github.com/midnightntwrk/midnight-node/issues/1442

> With the first implementation of ledger storage, we’re persisting all states. This includes intermediate states that we create in the process of importing or otherwise constructing a block.

This PR wraps the ledger state key in a new enum:
```rust
enum LedgerStateKey {
    Anchored(Vec<u8>),
    Transient(Vec<u8>),
}
```

When the state key is `Anchored`, we never call `unpersist` on it. When it's `Transient`, the state key will be unpersisted when applying transactions.

## 🗹 TODO before merging

- [x] Ready
- [x] Self-review

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [x] Node Runtime Update
- [x] Node Client Update
- [ ] Other: 
- [ ] N/A


ledger/src/host_api/ledger_7.rs
```diff
@@ -139,15 +139,15 @@ pub trait LedgerBridge {
 	#[version(2)]
 	fn apply_transaction(
 		&mut self,
-		state_key: PassFatPointerAndRead<&[u8]>,
+		state_key: PassFatPointerAndDecode<LedgerStateKey>,
```
gilescope: This changes the host function so we need a new version of these.

gilescope (changes requested):

ledger/src/host_api/ledger_7.rs
ozgb: Fixed https://github.com/midnightntwrk/midnight-node/pull/1443/commits/149e8ca8f695dee9057a363b5f1ea99fe58b9097

ozgb:
Closing in favour of https://github.com/midnightntwrk/midnight-node/pull/2050

Performance tests showed this implementation is a regression - and since it's a runtime change, it doesn't help with intermediate states created during sync

`persist/unpersist` was architected to be called infrequently, i.e. once per block

PR #1451, opened 2026-05-01
fix(ledger): dispatch host API on storage mode to fix unified panic

ozgb:
# Overview

`Bridge<S, D>` registers `Storage<D>` keyed by `D`'s TypeId, but the host
API pinned `D` to the default `ParityDb` (column offset 0) regardless of
mode. In `Unified` mode, `Bridge::set_default_storage` registers a
`Storage<ParityDb<_, _, NUM_COLUMNS_POLKADOT>>`, so subsequent
`default_storage::<D>()` lookups in `get_ledger`, `flush_storage`, etc.
miss and panic on the first authored block.

Read `LedgerStorageExt` at the host API boundary and dispatch to the
matching `D` for every method, including stateless ones — the latter
gain `&mut self` so future storage dependencies stay correct by default.
The addition of `&mut self` is still a client-only change - the runtime_interface
macro filters out `self` so the WASM-ABI is unchanged.

Also extend `drop_default_storage_if_exists` to drop both type variants
so shutdown / version-rollback works regardless of mode.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] I have included a change file, or skipped for this reason:  Skipped because it's a fix for an unreleased change
- [x] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [x] Node Client Update
- [ ] Other: 
- [ ] N/A


gilescope (approved):
(can we have a ticket on the backlog to later not support separate DBs so we can tidy up the code?)

PR #1234, opened 2026-04-03
[PM-21799] (fix): replace panic with error handling in get_data_from_inherent_data

m2ux:
## Summary

Replace `.expect()` on inherent data decoding in `get_data_from_inherent_data` with typed `Result<Option<...>, InherentError>` error handling, preventing simultaneous validator panics on malformed inherent data.

🎫 [PM-21799](https://shielded.atlassian.net/browse/PM-21799)  📐 [Engineering](https://github.com/shieldedtech/midnight-agent-eng/blob/mike/.engineering/artifacts/planning/2026-04-03-inherent-data-panic-fix/README.md)

---

## Motivation

The `get_data_from_inherent_data` function in the `cnight-observation` pallet decodes inherent data on every block through three consensus-critical entry points: `create_inherent`, `check_inherent`, and `is_inherent_required`. The function uses `.expect()` on the decode result, which means any malformed inherent data — from serialization bugs, version mismatches, or data corruption — causes all validators to panic simultaneously, permanently halting the chain with no automatic recovery.

This fix adopts the error handling pattern already established in the sibling `federated-authority-observation` pallet: returning `Result<Option<...>, InherentError>` with a new `DecodeFailed` variant, allowing each caller to handle decode failures appropriately without crashing.

---

### Primitives (`primitives/cnight-observation/src/lib.rs`)
- Add `DecodeFailed` variant to `InherentError` enum with `#[cfg_attr(feature = "std", error("Inherent data decode failed"))]`
- Variant inherits `is_fatal_error() = true` from the blanket `IsFatalError` impl

### Pallet (`pallets/cnight-observation/src/lib.rs`)
- **`get_data_from_inherent_data`** — Return type changed from `Option<T>` to `Result<Option<T>, InherentError>`; `.expect()` replaced with `.map_err(|_| InherentError::DecodeFailed)`
- **`create_inherent`** — Match on `Ok`/`Err`; error path logs `log::error!` with target `cnight-observation` and returns `None`
- **`check_inherent`** — Propagate decode error via `?`, then `.ok_or(InherentError::Other)?` for the missing-data case
- **`is_inherent_required`** — Propagate decode error via `?` before checking `.is_some()`

### Tests (`pallets/cnight-observation/tests/tests.rs`)
- Add `create_malformed_inherent()` helper (puts invalid bytes for `INHERENT_IDENTIFIER`)
- `create_inherent_with_malformed_data_returns_none` — verifies no panic, returns `None`
[...]

PR #1466, opened 2026-05-06
[PM-19896]: reduce panic risk in cnight-observation genesis construction

m2ux:
## Summary

Reduce panic risk in `cnight-observation` genesis construction by replacing the four remaining `expect()` calls in `BuildGenesisConfig::build()` with diagnostic panics whose messages name the chain-spec field path, the supplied byte length, and the maximum permitted length. The Substrate fail-fast genesis convention (`build()` returns `()`; node halts on invalid genesis) is preserved — only the diagnostic prose changes.

🎫 [Ticket](https://shielded.atlassian.net/browse/PM-19896)  📐 [Engineering](https://github.com/shieldedtech/midnight-agent-eng/blob/mike/.engineering/artifacts/planning/2026-05-06-node-issue-365-panic-risk-inherent-genesis/README.md)

GitHub Issue: shieldedtech/shielded-security-engineering#365

---

## Motivation

External audit flagged three panic sites in the `cnight-observation` pallet (formerly `native-token-observation`):

1. `get_data_from_inherent_data` — already remediated upstream via PM-21799 / PR #1234.
2. `set_mapping_validator_contract_address` — already remediated via the existing `try_into().map_err(...)` pattern.
3. `BuildGenesisConfig::build()` — **residual scope** for this work package.

The four remaining `expect()` calls in `build()` panic with four-word messages ("Mapping Validator address longer than expected", "Policy ID too long", "Asset name too long", "Auth Token asset name longer than expected") that don't say which field was wrong, what length was supplied, or what the cap was. Replacing them with descriptive errors that name the offending field and the maximum permitted length turns a noisy crash into a one-line, self-explanatory diagnostic, shortening operational recovery and closing the audit finding.

---

## Planned Changes

**Round 1 (landed — `8b3b8976`, `6bd01131`).** Named length constants in `primitives/cnight-observation`; four `expect()` → `unwrap_or_else(|_| panic!(...))` in `BuildGenesisConfig::build()` naming the field, supplied length, and cap; three panic-shape tests; mock helpers; change fragment.

**Round 2 (landed — `e0f64900`).** Correct the field path in panic messages to the runtime's actual chain-spec key (`cNightObservation.config.addresses.<field>`); replace the four near-duplicate panic blocks with a single generic helper `bounded_or_panic_with_field_path<S: Get<u32>>(field_path, bytes)` that reads the cap via `BoundedVec::<u8, S>::bound()`; update three round-1 tests; add two helper-level tests.

**Round 3 (landed — `134b4da8`, on top of merge `15e310b3`).** Addresses Klapeyron's overcomplication critique on `e0f64900`: remove the helper and inline the four panics, restore the mock crate to its pre-PR shape, reduce the five panic-message tests to one representative regression guard on `mapping_validator_address`, slim the change file. Net diff vs PR base shrinks from `+288/−46` to `+112/−21` (≈176 LOC reduction). Panic strings, field paths, and the operator-facing contract are unchanged.

**Out of scope:** sibling pallets' genesis builds; `serde_valid::Validate` at the genesis boundary; typing `CNightAddresses` fields as `BoundedVec` directly (deferred — captured as follow-up issue per LGLO's suggestion); asset-name minimum-length check (CDDL allows zero-length); storage migrations; runtime metadata regeneration.

Detailed plan: [`05-work-package-plan.md`](https://github.com/shieldedtech/midnight-agent-eng/blob/mike/.engineering/artifacts/planning/2026-05-06-node-issue-365-panic-risk-inherent-genesis/05-work-package-plan.md). Test plan: [`05-test-plan.md`](https://github.com/shieldedtech/midnight-agent-eng/blob/mike/.engineering/artifacts/planning/2026-05-06-node-issue-365-panic-risk-inherent-genesis/05-test-plan.md).

---

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [ ] I have included a change file, or skipped for this reason: [reason]
- [ ] No new todos introduced

---

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other
- [x] N/A

---

## 🗹 TODO before merging

- [ ] Ready for review


Klapeyron:
I am not sure if we can make significant improvement on that, be design `build` genesis calls are meant to fail and existing expects are quite descriptive, do you plan to provide the length details for `Policy ID too long` or `Asset name too long`?

m2ux:
> I am not sure if we can make significant improvement on that, be design `build` genesis calls are meant to fail and existing expects are quite descriptive, do you plan to provide the length details for `Policy ID too long` or `Asset name too long`?

yes

pallets/cnight-observation/src/lib.rs
```diff
@@ -295,42 +296,60 @@ pub mod pallet {
...
-					.as_bytes()
-					.to_vec()
-					.try_into()
-					.expect("Asset name too long"),
+				// Defence-in-depth: the source field `cnight_policy_id` is a fixed-size
+				// `[u8; 28]`, so this branch is unreachable from chain-spec deserialization
+				// (serde rejects mismatched lengths before reaching genesis build). The
+				// diagnostic panic is retained so a future change to the source type does
+				// not silently lose the operator-facing failure detail.
+				cnight_policy_id_bytes.try_into().unwrap_or_else(|_| {
```
LGLO: It's a shame that we store BoundedVec instead of [u8; 28]. This is a lot of defensive code that doesn't cover error happening in the other direction: when serialization allows to provide too short bytes.

pallets/cnight-observation/src/lib.rs
```diff
@@ -295,42 +296,60 @@ pub mod pallet {
...
+			let mapping_validator_len = mapping_validator_bytes.len();
 			MainChainMappingValidatorAddress::<T>::set(
-				self.config
-					.addresses
-					.mapping_validator_address
-					.as_bytes()
-					.to_vec()
-					.try_into()
-					.expect("Mapping Validator address longer than expected"),
+				mapping_validator_bytes.try_into().unwrap_or_else(|_| {
```
Klapeyron: there exists `bound` function on BoundedVec, so having that may allow us to stay without constants defined for number of elements.
Also TryFrom/TryInto error type is just the source vec that was provided, so we have an option to check the length of the source collection, which may allow us to provide a generic function for it

pallets/cnight-observation/src/lib.rs
```diff
@@ -295,42 +296,60 @@ pub mod pallet {
...
+					panic!(
+						"genesis: cnight_observation.addresses.mapping_validator_address \
```
LGLO: I think that the first segment depends on the name of the pallet in the runtime. IOW: this is too early to know it.
In case of midnight-runtime this name is not `cnight_observation` but `cNightObservation`.

LGLO (commented):
TBH I would prefer passing `BoundedVec`s in `CNightAddresses`. Now we have some parsing there and some parsing in genesis build.

pallets/cnight-observation/src/lib.rs
m2ux: I've kept the storage shape this round — switching `CNightAddresses` to `BoundedVec` directly is broader (it changes the chain-spec JSON encoding for operators who currently pass hex / ASCII), so it's deferred to a follow-up issue per your summary suggestion. On the too-short direction: zero-length asset names are valid per Cardano CDDL (`asset_name = bytes .size (0..32)`); for `cnight_policy_id` the source type is `[u8; 28]`, so serde rejects mismatched lengths before genesis runs.

pallets/cnight-observation/src/lib.rs
m2ux: Done in `e0f64900`: the four call sites now go through `bounded_or_panic_with_field_path<S: Get<u32>>(field_path, bytes)`. The helper uses `BoundedVec::<u8, S>::bound()` to discover the cap and reads the source length from the `TryInto` error path. The named-constant parentheticals are gone; messages name the numeric bound directly (e.g., `maximum 108`).

pallets/cnight-observation/src/lib.rs
m2ux: Confirmed and fixed in `e0f64900`: the panic now prints `cNightObservation.config.addresses.<field>` (lowerCamelCase, with the `.config` segment from `GenesisConfig<T>::config: CNightGenesis`). Verified against `runtime/src/lib.rs:913`, `res/perfnet/chain-spec-abridged.json:207-214`, and `scripts/genesis/genesis-construction.sh:540`.

m2ux:
@LGLO re your summary suggestion to type `CNightAddresses` fields as `BoundedVec` directly: agreed in principle and now listed under 'Out of scope' in the PR description as a follow-up. Pulling parsing into the deserialiser is the right end state, but it changes the chain-spec JSON encoding (operators currently pass a hex string for the policy ID and ASCII for the bech32 / asset names), so it deserves its own work package. Round 2 (`e0f64900`) keeps parsing in `BuildGenesisConfig::build()` but compresses the four sites through a single helper.

pallets/cnight-observation/src/lib.rs
```diff
@@ -288,50 +289,71 @@ pub mod pallet {
...
+	/// bound.
+	///
+	/// Used by `BuildGenesisConfig::build` to enrich the genesis fail-fast panic so that
+	/// an operator reading the startup-failure log can locate and correct the offending
+	/// chain-spec field. The cap is read from the destination type via
+	/// `BoundedVec::<u8, S>::bound()`, so the panic message and the storage type cannot
+	/// drift apart on a future bound change.
+	pub fn bounded_or_panic_with_field_path<S: Get<u32>>(
+		field_path: &'static str,
+		bytes: Vec<u8>,
```
Klapeyron: if you prefer to use this template function, you can take `payload: impl AsRef<[u8]>,` instead

pallets/cnight-observation/src/lib.rs
```diff
@@ -288,50 +289,71 @@ pub mod pallet {
...
+			// Each panic message names the chain-spec field path (matching the JSON keys an
+			// operator edits), the supplied byte length, and the maximum permitted length
+			// read from the destination `BoundedVec` type, so a startup-failure log points
+			// directly at the offending field.
+			MainChainMappingValidatorAddress::<T>::set(bounded_or_panic_with_field_path::<
+				ConstU32<CARDANO_BECH32_ADDRESS_MAX_LENGTH>,
+			>(
+				"cNightObservation.config.addresses.mapping_validator_address",
+				self.config.addresses.mapping_validator_address.as_bytes().to_vec(),
+			));
```
Klapeyron: I am not sure if amount of code here is reflecting the problem we are trying to solve, we have around 300 LOC to print size of the few vectors, in all this places we can add simple panic:
```rust
					.unwrap_or_else(|v| {
						panic!(
							"cNIGHT asset name too long: got length {}, expected at most {}",
							v.len(),
							BoundedCnightAssetName::bound() as usize,
						);
					}),
```
I am not sure if set of testcases, block comments, constants and mock helpers is making it more readable

LGLO:
> @LGLO re your summary suggestion to type `CNightAddresses` fields as `BoundedVec` directly: agreed in principle and now listed under 'Out of scope' in the PR description as a follow-up. Pulling parsing into the deserialiser is the right end state, but it changes the chain-spec JSON encoding (operators currently pass a hex string for the policy ID and ASCII for the bech32 / asset names), so it deserves its own work package. Round 2 (`e0f64900`) keeps parsing in `BuildGenesisConfig::build()` but compresses the four sites through a single helper.

Culprit of my comment was parsing in two places, we could replace
```
	#[serde(with = "hex")]
	pub cnight_policy_id: [u8; 28],
```
with
```
	#[serde(with = "our_function_that_outputs_bounded_vec_of_exactly_28_bytes")]
	pub cnight_policy_id: BoundedVec<...>,
```

and/or
```
	#[cfg_attr(feature = "std", validate(pattern = r"^(addr|addr_test)1[0-9a-z]{1,108}$"))]
	pub mapping_validator_address: String,
```	
with
```
	#[serde(with = "proper_serde_for_cardano_address")]
	pub mapping_validator_address: BoundedCardanoAddress,
```
(BTW validate pattern above allows for total 9+1+108=118 ascii chars, but runtime allows only 108 bytes) => again problem with having this logic in two places.

Anyway, I don't see a bug introduced in this PR. I think that if we touch this code (and add substantial number of lines), we can do it better.

pallets/cnight-observation/src/lib.rs
m2ux: `134b4da8` removes the helper, both mock helpers, and four of the five tests. The four call sites now inline `unwrap_or_else(|v: Vec<u8>| panic!(...))` reading the cap via `<BoundedType>::bound()`, with the field-path literals preserved from R3. One regression test on `mapping_validator_address` remains as a guard on the shared format-string shape; the other three sites are textually identical. Net diff vs PR base shrank from `+288/−46` to `+112/−21` (≈176 fewer LOC).

pallets/cnight-observation/src/lib.rs
m2ux: Moot after the simplification in `134b4da8` — the helper is removed and each site is inlined. The `impl AsRef<[u8]>` shape would have been the right move if we'd kept it; recording it in case the helper ever comes back in a follow-up.

m2ux:
The "parsing in two places" framing makes sense — a serde-side `BoundedVec` with a custom `serde(with = "…")` adapter would resolve both the policy-id and address sites the same way, and the 118-vs-108 regex/runtime mismatch would fall out naturally once the regex is replaced by a typed parse.

Round 3 (`134b4da8`) took the smaller-footprint path: the round-2 `+288/−46` is now `+112/−21`, with the mock crate byte-identical to the PR base. The deeper `CNightAddresses → BoundedVec` rework is captured in the work package's deferred-scope list as a follow-up.

LGLO (dismissed):

m2ux:
Blocked on Docker login issue

LGLO (approved):

Issue #1391, opened 2026-04-23
Db-sync lag causes Midnight network outages via banning


ozgb:
### Context & versions

**Affects all released Midnight Node versions**

**PR started in the partner-chains repo**: Needs cleaning up and migrating to the `midnight-node` repo: https://github.com/input-output-hk/partner-chains/pull/1139

We've run into several issues with nodes when db-sync falls behind the tip - this can be due to rollbacks or infra-outages. In this scenario, the node with the outdated db-sync will be unable to validate main-chain hash references in imported blocks. 

As a result, the BlockImport function returns `VerifcationFailed` - this causes the node with the lagging db-sync to penalise and disconnect from the peers that sent them the block. When reputation recovers and the peers reconnect, the node requests the same block range, resulting in a ban. 

This bi-directional banning has caused outages recently on qanet ([Midnight Slack link](https://shielded.slack.com/archives/C0A2JMF563H/p1767822349431509)), with the most recent outage happening 16/01/2026 (last Friday)

Related tickets:
- https://shielded.atlassian.net/browse/PM-16745
- https://shielded.atlassian.net/browse/PM-15838

<details><summary>Logs and technical details</summary>

> When verification fails, Substrate penalizes the peer who sent the block:
> 
> ```rust
> // chain_sync.rs:738-752
> Err(BlockImportError::VerificationFailed(peer_id, e)) => {
>     self.actions.push(SyncingAction::DropPeer(BadPeer(peer, rep::VERIFICATION_FAIL)));
>     self.restart();
> }
> ```
> 
> **Penalties applied:**
> - `VERIFICATION_FAIL` = -536,870,912 reputation
> - Peer is disconnected (`DropPeer`)
> - Peer enters `DisconnectedPeers` backoff

And the banning mechanism:

> When verification fails, the boot node calls `restart()` and re-requests the same block range. Peers interpret this as spam and ban the requester:
> 
> ```
> node-06: Report 12D3KooWHdiAxVd8uMQR1hGWXccidmfCwLqcMpGwR6QcTP6QRMuD: -2147483648
>          Reason: Same block request multiple times. Banned, disconnecting.
> 
> node-07: Report 12D3KooWHdiAxVd8uMQR1hGWXccidmfCwLqcMpGwR6QcTP6QRMuD: -2147483648
>          Reason: Same block request multiple times. Banned, disconnecting.
> 
[...]

PR #1472, opened 2026-05-07
Prevent db-sync lag make node banned from the network.

LGLO:
# Overview

### Do not ban peers when local Cardano observation lags during block verification

Problem being solved: when a node's local Cardano observation (db-sync) lags,
the node is being banned by its peers.
When block verification fails due db-sync lag, node re-requests the block from peers,
and doing so leads to reputation decrease ending up with a ban.

Solution to this problem is to hold up when the verified block is not stable or unknown.
That are separate cases.
If the block reference is not found at all, node triggers await loop only if our cardano view is certainly bad according to the Praos rules.
If the block is known but misses confirmations (but its timestamp is valid), the node enters a loop awaiting a fresh Cardano block to appear before checking the block again.
If the Cardano block timestamp is out of range in relation to the substrate block timestamp, then the block is immediately rejected as invalid.
This last condition narrows possibility of referencing blocks without enough confirmations, to trigger await loop in the rest of network.
## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A

## Links

https://github.com/midnightntwrk/midnight-node/issues/1391


partner-chains/toolkit/sidechain/sidechain-mc-hash/src/lib.rs
```diff
@@ -402,9 +461,23 @@ impl McHashInherentDataProvider {
...
+			parent_timestamp,
+			block_source,
+		)
+		.await?
+		{
+			ReferenceClassification::Stable(info) => info,
+			ReferenceClassification::AwaitingCardano | 	ReferenceClassification::Invalid => {
+				// Something out-of-assumptions happened: we can't find main chain block info of the block we have already accepted
+				// during verification of the previous substrate block.
+				panic!("Parent block MC Hash is not found and stable when current block MC Hash is stable.");
```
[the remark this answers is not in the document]
LGLO: This happens for a parent hash, that as comment says, was OK. Stable block cannot became unstable under assumptions of this program.
The only scenarios I can think of, leading to this errors involve removing db-sync state, that should surface prior to executing this panic.

node/src/inherent_data.rs
```diff
@@ -351,6 +381,13 @@ pub(crate) struct CreateInherentDataConfig {
 	pub mc_epoch_config: MainchainEpochConfig,
 	pub sc_slot_config: ScSlotConfig,
 	pub time_source: Arc<dyn TimeSource + Send + Sync + 'static>,
+	/// How long block verification waits before re-checking Cardano when our local Cardano
+	/// observation is lagging and cannot yet classify a referenced main-chain block.
+	#[new(value = "DEFAULT_CARDANO_CHECK_BACKOFF")]
+	pub cardano_check_backoff: Duration,
+	/// How many times verification re-checks Cardano before giving up and propagating the error.
+	#[new(value = "DEFAULT_CARDANO_CHECK_MAX_RETRIES")]
+	pub cardano_check_max_retries: u32,
```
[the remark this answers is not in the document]
LGLO: Done

partner-chains/toolkit/sidechain/sidechain-mc-hash/src/lib.rs
```diff
@@ -442,24 +510,67 @@ impl McHashInherentDataProvider {
...
+			}
+		},
+		// Unknown reference. If our Cardano view is healthy it is surely a dishonest reference;
+		// otherwise hold off until our Cardano observation recovers.
+		StableBlockByHashResult::BlockNotFound => {
+			if data_source
+				.is_cardano_ok()
+				.await
+				.map_err(McHashInherentError::DataSourceError)?
+			{
```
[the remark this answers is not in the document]
LGLO: I added 'chain quality rule' check in `is_cardano_ok`, to validate near tip conditions.

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
```diff
@@ -75,15 +79,23 @@ pub struct BlockDataSourceImpl {
 	mainchain_epoch_config: MainchainEpochConfig,
 	/// Additional offset applied when selecting the latest stable Cardano block
 	///
-	/// This parameter should be 0 by default and should only be increased to 1 in networks
-	/// struggling with frequent block rejections due to Db-Sync or Cardano node lag.
+	/// Block producers SHOULD take this parameter into account.
+	/// Block verification uses this value to determine if their observed Cardano tip is recent enough,
+	/// when verified block references unstable Cardano block.
+	/// Bigger value of this parameter makes attack on partner-chain liveness not feasible.
```
ozgb: Bigger value for block producers, or bigger value for block verifiers? 

Before this PR, a bigger value for block producers improved network stability (less likely that a verifier will reject the block)
Having a bigger value as a verifier will protect against a liveness attack?

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
```diff
@@ -75,15 +79,23 @@ pub struct BlockDataSourceImpl {
...
+	/// Recommended value is 10, that add 200 seconds of lag in observing Cardano, but is also safe unless Cardano tip is not older then 200 seconds.
 	block_stability_margin: u32,
+	/// Maximum tolerated age of the latest observed Cardano block before our view of
+	/// Cardano is considered stale by [Self::is_cardano_ok].
+	max_latest_block_age_seconds: u32,
```
ozgb, line 91: This should be set to < the block stability margin of the block producer (in practice, less than the default block stability margin for the chain)

e.g. if this max_latest_block_age_seconds is 100s, and the stability margin is 10 (20s * 10 = 200s), then a validator will still be able to validate blocks if its db-sync is lagging by 100s - and it will never reject a block as invalid if it's db-sync is lagging.

If the block producer has block stability margin set to 5 (100s), and max_latest_block_age_seconds is 200s, then if the verifiers db-sync is lagging by 120s, it will consider it's cardano tip not-stale, and reject the incoming block, causing peer bans

We should document this restriction - 

This could be an attack vector - if a block producer maliciously produced a block with stability margin 0 during a Cardano rollback (some nodes db-sync will lag), it could cause peers bans between the nodes depending on their db-sync state

Mitigation suggestion: we do a soft-reject of the block in [this case only](https://github.com/midnightntwrk/midnight-node/blob/9eddb58a9e6f1c0465e0c0dbf6040f6ba92a2eb3/partner-chains/toolkit/sidechain/sidechain-mc-hash/src/lib.rs#L553-L556)
Soft reject == return `ImportResult::MissingState` rather than `std::io::ErrorKind::InvalidData`, to avoid peer-banning effects.

Would require our own impl. of `BlockImport`, as in [this PR](https://github.com/input-output-hk/partner-chains/pull/1139/changes#diff-c2f8086c44bf8a530cbd2987f0c8e1fe6c3cd19edd5fc0e76039c0d56f84a502)

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
LGLO: The first sentence, regarding block producers, is fix for obsolete comment about defaulting to 0.

Having bigger value as verifier will cause less false-positives of "is cardano stale" due to real Cardano old tip.
So, if there is a malicious node that would like to trigger loop awaiting new Cardano tip in other nodes, would have to send unstable block reference when there is "hole" at Cardano tip that bigger than BLOCK_STABILITY_MARGIN / active_slots_coefficient.

For BLOCK_STABILITY_MARGIN = 10, it is 200s (active slots coeff = 0.05)

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
LGLO: The first observation: db-sync lags are caused either by some restart and catch up - in this db-sync catches up with "minutes of Cardano" per second. In such a case Cardano will be in "not ok" state until the latest synced block is 12 hrs old. Then it is "stale" until it gets the block 200s old.
In the "stale" period it should already resolve any honest stable block reference (because these are at least 12hrs old).

No rejection of honest block (stable and offset by margin) should not happen if algorithm works as I expected.

The attack vector you describe is basically "propose very recent stable block (ignoring margin param)". This attack is a proposers gamble: if its peers don't know as many Cardano blocks as attacker, this block will be rejected and the attacker will lose reputation. But it can be that some node that are "not stale" see one Cardano block less and will reject it.

Before "BLOCK_STABILITY_MARGIN" parameter we had this situation "business as usual" and networks weren't stopping immediately.
My opinion on this is that this vector of attack could succeed only against nodes whose observability is constantly lagging by margin less than "staleness" threshold.

It is not the primary scenario we defend against in this PR: transient cardano-node/db-sync problems.

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
```diff
@@ -139,6 +151,49 @@ impl BlockDataSourceImpl {
...
+			None => return Ok(false),
+			Some(block) => block,
+		};
+		// Praos chain quality rule check.
+		if current_time - self.min_slot_boundary_as_seconds > latest.time {
+			log::debug!(
+				"is_cardano_ok current time: {current_time}, latest: {latest:?}, the latest block is too old."
+			);
+			return Ok(false);
+		}
```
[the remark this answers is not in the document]
LGLO: When verifier is minutes behind producer, then honest stable block (min 12hrs old) will be found as `NotEnoughConfirmations`

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
ozgb: Ah, I see - I missed that the value of `max_latest_block_age_seconds` is derived from the `block_security_parameter` - could we add that to the doc string as a suggestion? ("should be set to `block_stability_margin * expected block time`")

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
ozgb: Got it - for clarity, can we edit the doc string to something like:

```suggestion
	/// Bigger value of this parameter across both block producers and validators makes attacks on partner-chain liveness not feasible.
```

ozgb (dismissed):
Good solution - leaves behaviour of validators with up-to-date db-sync instances in-tact, while intelligently pausing validation when db-sync is lagging/Cardano is having issues

partner-chains/toolkit/sidechain/sidechain-mc-hash/src/lib.rs
```diff
@@ -442,24 +515,70 @@ impl McHashInherentDataProvider {
...
+	{
+		StableBlockByHashResult::BlockStable { info } => Ok(BlockHashClassification::Stable(info)),
+		StableBlockByHashResult::BlockTimestampOutRange { .. } => {
+			Ok(BlockHashClassification::Invalid)
+		},
+		StableBlockByHashResult::NotEnoughConfirmations { .. } => {
+			if data_source
+				.is_cardano_tip_fresh()
+				.await
+				.map_err(McHashInherentError::DataSourceError)?
```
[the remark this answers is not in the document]
LGLO: > while a verifier whose db-sync is 11 blocks behind still has a latest block less than the 200s freshness window old; db-sync then returns

11 blocks is 220s on average. This is why this PR binds `max_latest_block_age_seconds` to `block_stability_margin` - in order to avoid problems when nodes are honest. Other than that, it `block_stability_margin = 10` means that we accept situations when someones db-sync lags up to 10 blocks. If it lags more, than it's on the node operator. 10 blocks lag (~200s) is already very handsome margin.

ozgb (dismissed):

node/src/inherent_data.rs
```diff
@@ -266,19 +275,40 @@ where
...
+							"Still awaiting fresh Cardano data to verify MC reference {hash} after {max_retries} retries; giving up and rejecting the block"
+						);
+						return Err(McHashInherentError::AwaitingCardanoData(hash).into());
+					}
+					cardano_check_attempts += 1;
+					let backoff = config.cardano_check_backoff;
+					log::warn!(
+						"Awaiting fresh Cardano data to verify MC reference {hash}; retry {cardano_check_attempts}/{max_retries} in {backoff:?}"
+					);
+					tokio::time::sleep(backoff).await;
```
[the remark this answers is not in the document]
LGLO: Block hash is bound with block slot/timestamp.

ozgb (approved):

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
LGLO: Done

PR #1475, opened 2026-05-07
fix(node): propagate ContractNotPresent to RPC layer

### LINKED ISSUE #1166: [Node] Issue AD: Return ContractNotPresent Instead of Default State

`do_get_contract_state` should return an explicit `ContractNotPresent` error when queried for a non-existent contract address, instead of returning an empty/default state.

## User Need

Callers of the contract state query need to distinguish between "contract exists with empty state" and "contract does not exist." Currently, downstream logic that trusts the returned state may operate on invalid assumptions.

## Expected Benefit

Clear error signaling for missing contracts, preventing silent incorrect data from propagating through the RPC layer.

## Details

**Current state:**
- `do_get_contract_state` returns a default/empty state when the contract address is not found in storage
- No error variant exists for the "contract not present" case in the return type

**Desired state:**
- `do_get_contract_state` returns an explicit `ContractNotPresent` error when the contract address is not found
- The error is propagated through the RPC layer with a meaningful error code and message

**Scope:**
- `do_get_contract_state` implementation
- Error type addition (`ContractNotPresent` variant)
- RPC error propagation for the missing-contract case

**Acceptance Criteria:**
- [ ] `do_get_contract_state` returns a `ContractNotPresent` error for non-existent contracts
- [ ] The error is propagated through the RPC layer with a descriptive message
- [ ] Callers can distinguish between empty state and missing contract
- [ ] Existing tests pass and new tests cover the missing-contract path

**Constraints:**
- Security: Audit finding rated Low severity (Least Authority Issue AD)
- Compatibility: This is a behavioral change; callers relying on the default-state return will need updating

**References:**
- Least Authority - Shielded Midnight Network Node Initial Audit Report (October 24, 2025), Issue AD, pp. 45-46
- Audit git revision: `d204679fea653a34fa6dccd0e6792c34a2e9937b`

---

_Migrated from [PM-19973](https://shielded.atlassian.net/browse/PM-19973)_

[PM-19973]: https://shielded.atlassian.net/browse/PM-19973?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

rsporny:
# Overview
PR #916 added LedgerApiError::ContractNotPresent in the runtime, but the
  midnight_contractState RPC still collapsed every LedgerApiError into
the
  generic UnableToGetContractState — so callers could not distinguish
  "empty state" from "no such contract", which is why audit Issue AD was
  re-opened as #1166.

  Add StateRpcError::ContractNotPresent and route the matching ledger
  error to it. All other variants still fall through to
  UnableToGetContractState, so this is a strict superset of prior
  behaviour. Re-export LedgerApiError from pallet-midnight so the RPC
  crate can match on it without a new dependency.

  Lock in the audit invariants with three e2e tests in tests/e2e:
  - ContractNotPresent for an undeployed (but well-formed) address
  - BadContractAddress and ContractNotPresent stay distinct paths
  - same address yields ContractNotPresent at block 1 and the deployed state at the post-deploy best block

  Bump NUM_PRE_DEPLOY_TESTS / NUM_DEPLOY_TESTS 2→3 and the Earthfile
  test-threads 4→6 so the new deploy gate doesn't deadlock.

  Closes: #1166

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [x] N/A

### DIFF
--- a/pallets/midnight/rpc/src/lib.rs
+++ b/pallets/midnight/rpc/src/lib.rs
@@ -21,7 +21,7 @@ use jsonrpsee::{
 	types::error::{ErrorObject, ErrorObjectOwned, INVALID_PARAMS_CODE},
 };
 
-use pallet_midnight::MidnightRuntimeApi;
+use pallet_midnight::{LedgerApiError, MidnightRuntimeApi};
 use sc_client_api::{BlockBackend, BlockchainEvents};
 use sp_api::{ApiExt, ProvideRuntimeApi};
 use sp_blockchain::HeaderBackend;
@@ -78,6 +78,7 @@ pub trait MidnightApi<BlockHash> {
 pub enum StateRpcError {
 	BadContractAddress(String),
 	BadAccountAddress(String),
+	ContractNotPresent,
 	UnableToGetContractState,
 	UnableToGetZSwapChainState,
 	UnableToGetZSwapStateRoot,
@@ -134,6 +135,9 @@ impl Display for StateRpcError {
 			},
 			StateRpcError::BadAccountAddress(malformed_address) => {
 				write!(f, "Unable to decode account address: {}", malformed_address)
+			},
+			StateRpcError::ContractNotPresent => {
+				write!(f, "Contract not present at the requested address")
 			},
 			StateRpcError::UnableToGetContractState => {
 				write!(f, "Unable to get requested contract state")
@@ -293,6 +297,9 @@ where
 			.map_err(|_| StateRpcError::UnableToGetContractState)?;
 
 		let result = if api_version < 2 {
+			// Legacy path: v1 of the RPC contract predates ContractNotPresent,
+			// so callers on api_version < 2 must continue to see the generic
+			// UnableToGetContractState. Do not surface ContractNotPresent here.
 			#[allow(deprecated)]
 			api.get_contract_state_before_version_2(at, dehexed)
 				.map_err(|_e| StateRpcError::UnableToGetContractState)?
@@ -300,7 +307,10 @@ where
 			api.get_contract_state(at, dehexed)
 				.map_err(|_e| StateRpcError::UnableToGetContractState)
 				.and_then(|inner_res| {
-					inner_res.map_err(|_| StateRpcError::UnableToGetContractState)
+					inner_res.map_err(|e| match e {
+						LedgerApiError::ContractNotPresent => StateRpcError::ContractNotPresent,
+						_ => StateRpcError::UnableToGetContractState,
+					})
 				})?
 		};
 
--- a/tests/e2e/README.md
+++ b/tests/e2e/README.md
@@ -7,4 +7,11 @@ To execute these tests locally, run `cargo test --test e2e_tests --no-default-fe
 alias: `cargo test-e2e-local`
 
 To run test in parallel use `--test-threads N` argument, e.g.
-`cargo test --test e2e_tests --no-default-features --features local -- --test-threads 4 --no-capture`
+`cargo test --test e2e_tests --no-default-features --features local -- --test-threads 6 --no-capture`
+
+`--test-threads` must be `>= NUM_PRE_DEPLOY_TESTS + NUM_DEPLOY_TESTS` (currently 6) — see
+the gate constants in `tests/e2e/tests/lib.rs`. Lower values can deadlock the deploy gate.
+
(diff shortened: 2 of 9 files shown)


LGLO (approved):

pallets/midnight/rpc/src/lib.rs
```diff
@@ -300,7 +304,10 @@ where
 			api.get_contract_state(at, dehexed)
 				.map_err(|_e| StateRpcError::UnableToGetContractState)
 				.and_then(|inner_res| {
-					inner_res.map_err(|_| StateRpcError::UnableToGetContractState)
+					inner_res.map_err(|e| match e {
```
m2ux, line 310: Optional / nit: Consider a single-line comment in the `api_version < 2` branch above (line 299) noting that the legacy path intentionally does not surface `ContractNotPresent` — the v1 RPC contract predates the variant, so a caller on API v1 must continue to receive the generic error. A line of code documentation there helps the next reader avoid wondering whether this branch was missed by the audit fix.

tests/e2e/tests/lib.rs
```diff
@@ -1516,29 +1528,206 @@ async fn valid_deploy_transaction_succeeds_via_rpc() {
...
+    use midnight_node_toolkit::commands::generate_txs::{self, GenerateTxsArgs};
+    use midnight_node_toolkit::tx_generator::builder::{Builder, ContractCall, ContractDeployArgs};
+    use midnight_node_toolkit::tx_generator::destination::Destination;
+    use midnight_node_toolkit::tx_generator::source::Source;
+
+    let settings = Settings::default();
+    let client = MidnightClient::new(settings.node_client.clone()).await;
+
+    let pre_deploy_hash = client
+        .get_block_hash_at_height(1)
```
m2ux: Optional / test robustness: Consider adding a brief retry on `get_block_hash_at_height(1)` to defend against a CI runner that starts this test before AURA has produced the first block. The other deploy tests don't need this because `wait_before_deploying()` waits on the gate, which is released by tests that touch chain state — but this test queries height 1 directly. Low-priority since AURA produces the first block in ~6s in practice; raising it as a small CI-stability hedge.

tests/e2e/README.md
```diff
@@ -7,4 +7,7 @@ To execute these tests locally, run `cargo test --test e2e_tests --no-default-fe
 alias: `cargo test-e2e-local`
 
 To run test in parallel use `--test-threads N` argument, e.g.
-`cargo test --test e2e_tests --no-default-features --features local -- --test-threads 4 --no-capture`
+`cargo test --test e2e_tests --no-default-features --features local -- --test-threads 6 --no-capture`
+
+`--test-threads` must be `>= NUM_PRE_DEPLOY_TESTS + NUM_DEPLOY_TESTS` (currently 6) — see
+the gate constants in `tests/e2e/tests/lib.rs`. Lower values can deadlock the deploy gate.
```
m2ux, line 13: Optional / discoverability: The README documents the `--test-threads` lower bound, which is great. Consider also documenting `E2E_SKIP_DEPLOY_GATE` here so contributors who want to run a single deploy test (`cargo test <name>`) discover the escape hatch without having to read `wait_before_deploying`'s inline comment.

m2ux (approved):
Approving review (work-package review-mode pass).

The PR completes the audit fix correctly: it surfaces `LedgerApiError::ContractNotPresent` through the `midnight_contractState` RPC as a dedicated `StateRpcError::ContractNotPresent` variant. Layering is preserved (translation lives at the boundary; `LedgerApiError` is re-exported rather than introducing a new dep), the change is additive (strict superset of prior outcomes), and the e2e suite enforces all three audit invariants end-to-end (well-formed-but-undeployed, malformed-vs-not-present distinctness, historical-vs-current-block).

The two scope-adjacent changes (`DEPLOY_SERIAL` mutex and `E2E_SKIP_DEPLOY_GATE` env-var) are defensible — the new third deploy test surfaces the deploy-phase race they prevent, and bundling them avoids a forced ordering between this PR and a separate concurrency PR.

No blocking concerns. Three optional comments inline (C1–C3) for quality-of-life only; they should not delay merge.

PR #1486, opened 2026-05-11
fix(toolkit): contract-address --untagged handling and stdout cleanliness (#1402)

### LINKED ISSUE #1402: [node-toolkit] contract-address command rejects --untagged option even if documented
> **Migrated from Jira:** [PM-19934](https://shielded.atlassian.net/browse/PM-19934)
> Content below preserved verbatim from the Jira ticket. Further updates should happen here.

## Description

Toolkit command contract-address returns the following when using `--untagged` option

```
{
  "output": "Warning: `--untagged` flag is deprecated (now default)\nc6b2fc9455741c9ca7806520d993d12ac8a2ef243bcbfb545820356388f91a72\n",
  "stdout": "c6b2fc9455741c9ca7806520d993d12ac8a2ef243bcbfb545820356388f91a72\n",
  "stderr": "Warning: `--untagged` flag is deprecated (now default)\n",
  "exitCode": 0
}
```

although this doesn’t appear in the help

```
Show and save in a file the Contract Address included in a DeployContract tx

Usage: midnight-node-toolkit contract-address [OPTIONS] --network <NETWORK> --src-file <SRC_FILE>

Options:
      --network <NETWORK>    Target network
  -s, --src-file <SRC_FILE>  Serialized Transaction
      --tagged               Serialize Tagged
      --untagged             Serialize Untagged
  -h, --help                 Print help
```

Having the full json object when no options are specified was good, was consistent with the show address behaviour

## Original Jira metadata

| Field | Value |
|---|---|
| **Jira key** | [PM-19934](https://shielded.atlassian.net/browse/PM-19934) |
| **Type** | Bug |
| **Status at migration** | To Do |
| **Priority** | Low |
| **Assignee (Jira)** | unassigned |
| **Reporter** | Giuseppe Salvatore |
| **Created** | 2025-10-09T20:52:09.007+0100 |
| **Updated** | 2026-03-05T16:02:20.783+0000 |
| **Jira labels** | `2026-Q2-SOW`, `midnight-toolkit`, `toolkit`, `triaged` |
| **Components** | — |
| **Fix versions** | — |

## Attachments

—

### Comment by Sinead Plunkett — 2025-10-13T09:33:39.096+0100

Triage Comments

Lowering Priority - due to being just a doc issue

---

### Comment by Oscar Bailey — 2025-10-13T09:52:39.108+0100

> Having the full json object when no options are specified was good, was consistent with the show address behaviour

This is why I set it to that to begin with - the motivation behind the change was that 99% of the time, a user will want the untagged version since contract address is untagged across the stack (and spec-ed as untagged)

For the TS wrapper though, having a JSON output is better because you only have to run the command once - I can add a `--json` option to the command that outputs the full JSON? 

[PM-19934]: https://shielded.atlassian.net/browse/PM-19934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-19934]: https://shielded.atlassian.net/browse/PM-19934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

m2ux:
## Summary

Align `midnight-node-toolkit contract-address` help output with runtime behaviour. Hide the deprecated `--untagged` flag from `--help`, move its deprecation warning from `eprintln!` to `log::warn!` so it participates in the toolkit's logging contract (`RUST_LOG`, `--quiet`, `--verbose`), and reject `--tagged --untagged` at parse time via `conflicts_with`. Success path now emits empty stderr, matching the `show-address` sister-command pattern.

🐛 [Issue](https://github.com/midnightntwrk/midnight-node/issues/1402)  📐 [Engineering](https://github.com/shieldedtech/midnight-agent-eng/blob/mike/.engineering/artifacts/planning/2026-05-11-toolkit-contract-address-untagged-flag/README.md)

_Jira: [PM-19934](https://shielded.atlassian.net/browse/PM-19934)_

---

## Motivation

The toolkit `contract-address` command advertises an `--untagged` option in its `--help` output, but the implementation already moved to untagged-by-default. Passing `--untagged` emits `Warning: \`--untagged\` flag is deprecated (now default)` via `eprintln!` on stderr while still returning the same address `--no-flag` would have returned. Help text and runtime behaviour disagree, and the warning channel bypasses the toolkit's `tracing_subscriber` logging contract — so it is not suppressible via `RUST_LOG`, not gated by `--quiet`, and does not pick up the standard log prefix. Downstream consumers that merge stdout/stderr observe the warning mixed in with the address.

The sister command `show-address` already handles its own deprecated alias (`--unshielded-user-address-untagged`) cleanly by pairing `conflicts_with` declarations on the current and deprecated flags and emitting `log::warn!` only when the deprecated alias is selected. This PR brings `contract-address` into alignment with that pattern.

---

## Changes

- Mark `ContractAddressArgs::untagged` with `hide = true` so it disappears from `--help` while remaining parseable for existing callers
- Add `conflicts_with` declarations between `--tagged` and `--untagged` so passing both is a clap parse error rather than a silent "tagged wins" outcome
- Replace `eprintln!` deprecation warning with `log::warn!` matching the `show-address` idiom; the warning is now suppressible via `RUST_LOG=error` and picks up the standard toolkit log prefix
- Rewrite the doc-comment on the deprecated field to explicitly state "kept for backward compatibility; will be removed in a future major version"
- Add `trycmd` snapshot tests under `util/toolkit/tests/cmd/` covering: (a) `--help` does not advertise `--untagged`, (b) success path emits empty stderr, (c) `--untagged` alone still works but logs a warning, (d) `--tagged --untagged` is rejected by clap
- Add `changes/toolkit/changed/contract-address-untagged-flag-cleanup.md` change-file entry

The `--tagged` alternative output mode and the default-no-flag output are unchanged. Path B (full flag removal) was considered and rejected for this Low-priority UX bug on external-caller-risk grounds; see the linked Engineering README for the design rationale.

---

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [ ] I have included a change file, or skipped for this reason: [reason]
- [ ] No new todos introduced

---

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other
- [x] N/A

---

## 🗹 TODO before merging

- [ ] Ready for review

---

Closes #1402

[PM-19934]: https://shielded.atlassian.net/browse/PM-19934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


util/toolkit/tests/cmd/contract-address-tagged.toml
```diff
@@ -0,0 +1,12 @@
...
+args = [
+    "contract-address",
+    "--src-file", "../../res/test-contract/contract_tx_1_deploy_undeployed.mn",
+    "--tagged",
+]
+status.code = 0
+stderr = ""
+
+stdout = """
+6d69646e696768743a636f6e74726163742d616464726573735b76325d3add76bcd08403b3a9af64485b1f3960437e7c5872c269d6d06d32b2f298d71577
```
ozgb: Having the exact contract address in the test file makes this test fragile

m2ux: Addressed in a961a8d8.

Replaced the hardcoded address bytes with trycmd's `[..]` wildcard across the three snapshots that carried them:

- `contract-address-tagged.toml`: kept the `midnight:contract-address[v2]:` prefix so the tagged framing is still asserted, replaced only the post-prefix payload with `[..]`
- `contract-address-default.toml` and `contract-address-untagged-deprecated.toml`: replaced the bare untagged address line with `[..]`

ozgb (dismissed):

ozgb (approved):

PR #1489, opened 2026-05-11
#233: input validation error reported as fatal

m2ux:
# Overview

Stop `pallet-cnight-observation` from emitting a "Fatal" log line for Dust registrations whose `DustPublicKey` byte payload is the wrong scalar value; validate at the inherent boundary (consensus-gated by a `CNightObservationApi` v2 → v3 bump) and narrow the legacy log severity in the pallet so misleading "Fatal" lines disappear on both new and old runtimes.

🐛 [Issue](https://github.com/shieldedtech/shielded-security-engineering/issues/233)  📐 [Engineering](https://github.com/shieldedtech/midnight-agent-eng/blob/mike/.engineering/artifacts/planning/2026-05-11-node-issue-233-dust-public-key-fatal-validation/README.md)

---

## Motivation

A Dust registration submitted on Cardano can carry a public key that is the right length but whose value falls outside the valid cryptographic range. Nothing rejects it on the way in: the data source only checks the length, and the code that records the registration never checks the value, so it is stored unchanged. The bad value only surfaces much later, when a related asset event tries to use it. At that point the ledger reports a deserialization failure and the observation pallet logs a line labelled "Fatal".

In practice the chain is unaffected: the invalid registration is skipped, that one token movement is ignored, and block production continues. The problem is purely the wording. "Fatal" tells on-call engineers to escalate and trips severity-based alerting, even though the situation is harmless and recoverable. This change rejects invalid registrations before they ever enter chain state, and softens the existing log line so the older code path stops mislabelling the same case while the network is mid-upgrade.

---

## Changes

- Invalid Dust public keys are now caught the moment a registration arrives from Cardano, so they never reach chain state. The new check is gated behind a runtime API version bump, so it only activates once the network upgrades and every node agrees on the behaviour; older nodes keep the previous pass-through during the upgrade window.
- The misleading log line no longer says "Fatal". The recoverable "skip this one and continue" case now logs quietly at low severity, and other unexpected failures log as warnings rather than errors.
- Registrations already stored on existing networks are deliberately left in place; there is no storage migration.
- New tests and a change file cover the early validation, the version-gated filtering on both old and new runtimes, and the corrected logging.

Out of scope: removing invalid registrations already in storage, extending the early validation to other kinds of decode failures, and rate-limiting for spam.

---

## 🗹 TODO before merging

- [x] Ready for review

---

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] I have included a change file, or skipped for this reason: [reason]
- [x] No new todos introduced

---

## 🧪 Testing Evidence

New unit and integration tests landed with the change and run in CI:

- **Key validation** — a valid Dust public key round-trips, while an out-of-range value, an empty input, and a too-short input are all rejected.
- **Version-gated filtering** — the old runtime passes registrations through unchanged, the new runtime drops the invalid ones, only registration entries are affected, and the order of the remaining entries is preserved.
- **Pallet behaviour and batch tolerance** — an invalid key is skipped, and a mixed batch of one valid and one invalid registration still applies the valid one exactly once without disturbing the rest.
- **Lint and format gate** — formatting, build, and lint checks all pass; test execution runs in CI.

---

## 🔱 Fork Strategy

- [x] Node Runtime Update
- [ ] Node Client Update
- [ ] Other
- [ ] N/A

The `CNightObservationApi` v2 → v3 bump is consensus-affecting: the inherent payload changes once the IDP filter activates. Legacy pass-through is preserved at `api_version <= 2` so old-runtime + new-binary pairings remain consensus-equivalent across the upgrade window.

---

## Links

- Closes shieldedtech/shielded-security-engineering#233
- Jira: [PM-22301](https://shielded.atlassian.net/browse/PM-22301)
- Engineering planning folder: [`2026-05-11-node-issue-233-dust-public-key-fatal-validation`](https://github.com/shieldedtech/midnight-agent-eng/blob/mike/.engineering/artifacts/planning/2026-05-11-node-issue-233-dust-public-key-fatal-validation/README.md)

[PM-22301]: https://shielded.atlassian.net/browse/PM-22301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


m2ux (commented):
## Review feedback

The added test surface in `pallets/cnight-observation/tests/tests.rs` (commit `18527afc`) is disproportionate to the change set:

- one `handle_create` arm narrowed
- one `handle_spend` arm narrowed
- one IDP filter free function
- one runtime API version bump
- log severity downgrades on the dust-public-key branch

Please prune the two integration tests added in commit `18527afc` (`handle_spend_returns_none_for_invalid_dust_public_key`, `process_tokens_preserves_batch_tolerance_with_invalid_registration`) and any supporting helpers down to the high-value cases. Keep coverage for the actual invalid-dust-public-key fatal-validation branch; drop redundant or low-signal scaffolding.

Scope: single file in `pallets/cnight-observation/tests/`, no behaviour change, no plan revision required.

ledger/src/versions/common/api/mod.rs
```diff
@@ -211,3 +211,78 @@ impl Api {
...
+/// out-of-Fr-range inputs. Callers that source bytes from `DustPublicKeyBytes`
+/// (`BoundedVec<u8, ConstU32<33>>`) already have the wire-length envelope
+/// enforced upstream; this helper additionally performs the Fr-range check,
+/// since values whose 33-byte encoding sits above the Bls12-381 Fr modulus
+/// pass the length check but fail downstream circuit use. The check is
+/// performed without emitting the `log::error!` line that `Api::deserialize`
+/// produces on failure — call sites that filter inputs upstream (e.g. the
+/// cNight-observation inherent-data provider) treat invalid registrations as
+/// a per-UTXO non-fatal outcome, so the deserialise failure must not surface
+/// as an `error`-severity log line.
```
gilescope: The doc seems surprisingly long for the one line function?

ledger/src/versions/common/api/mod.rs
```diff
@@ -211,3 +211,78 @@ impl Api {
...
+	fn dust_public_key_is_valid_rejects_high_byte_set() {
+		// A 33-byte vector with the leading byte 0xff forces the encoded value
+		// above the Bls12-381 Fr modulus (~2^254), so deserialisation must fail.
+		//
+		// L-5 (plan): no `log::error!` is emitted on this rejection — the
+		// validator calls the `Deserializable` trait directly, bypassing the
+		// `Api::deserialize` wrapper that produces the `error`-severity log
+		// line. Pinned by inspection rather than a log-capture assertion
+		// (the ledger crate's test target does not install a subscriber).
+		let bytes = vec![0xffu8; 33];
```
gilescope, line 258: A one line comment would be enough

pallets/cnight-observation/tests/tests.rs
```diff
@@ -110,6 +110,19 @@ fn dust_public_key() -> DustPublicKeyBytes {
...
 
+/// Onchain dust address that fails the IDP-level Fr-range validator.
+///
+/// A length-32 vector of `0xFF` bytes round-trips through the
+/// `DustPublicKeyBytes` envelope (`BoundedVec<u8, ConstU32<33>>`) but encodes
+/// a scalar above the Bls12-381 Fr modulus, so
+/// `<DustPublicKey as Deserializable>::deserialize` rejects it. The pallet
+/// writer admits this shape at the v2 api (legacy semantics), and
+/// `handle_create` / `handle_spend` route it through the variant-narrowed
+/// `Deserialization(DustPublicKey)` arm of `LedgerApiError`.
```
gilescope:
```suggestion
/// Invalid dust address (fails IDP-level Fr-range validator).
```

PR #1522, opened 2026-05-15
Local env node release 1.0.0 support

justinfrevert:
# Overview

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] All commits are signed off (`git commit -s`) for the DCO
- [x] Changes are backward-compatible (or flagged if breaking)
- [ ] Pull request description explains why the change is needed
- [ ] Self-reviewed the diff
- [ ] I have included a change file, or skipped for this reason: 
- [ ] If the changes introduce a new feature, I have bumped the node minor version
- [ ] Update documentation (if relevant)
- [ ] Updated AGENTS.md if build commands, architecture, or workflows changed
- [ ] No new todos introduced

## 🧪 Testing Evidence

Please describe any additional testing aside from CI:

- [ ] Additional tests are provided (if possible)

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [x] N/A

PR #1513, opened 2026-05-14
Fix bridge reserve transfer condition

LGLO:
# Overview

Fixes bug that causes wrong classification of some bridge transfers.

In order to properly classify transfer: User vs Reserve we ought to look at origin of ICS tokens surplus.

Runtime has to know reserve validator address to be able to pass it to observability. Because bridge is not deployed yet, we don't add migration here. Addresses will have to be overwritten, so they can be read with the new format.

SQL query returns tokens in and out for reserve validator and ICS validator.

Assumption for Reserve is that it allows only to transfer out to ICS. So, negative delta of Reserve address is "Reserve Transfer". Then positive delta at ICS is what user has added. This is corner case. We expect that there won't be such combined transfer, but the code is ready for such calculation.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] All commits are signed off (`git commit -s`) for the DCO
- [x] Changes are backward-compatible (or flagged if breaking)
- [x] Pull request description explains why the change is needed
- [x] Self-reviewed the diff
- [x] I have included a change file, or skipped for this reason: 
- [x] If the changes introduce a new feature, I have bumped the node minor version
- [ ] Update documentation (if relevant)
- [ ] Updated AGENTS.md if build commands, architecture, or workflows changed
- [x] No new todos introduced

## 🧪 Testing Evidence

Please describe any additional testing aside from CI:

- [ ] Additional tests are provided (if possible)

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A

## Links

Fixes: #1479

PR #1523, opened 2026-05-15
feat(node): add in-tree try-runtime dry-run subcommand and manual CI workflow

jacek-kurkowski-shielded:
# Overview

Two things, one found the other.

**`midnight-node try-runtime`** (`node/src/try_runtime.rs`, behind `--features try-runtime`) dry-runs a runtime upgrade against a state snapshot: `Migrations` plus every pallet's `on_runtime_upgrade`, then the multi-block migration queue, then `try_decode_entire_state` and every pallet's `try_state`. In-tree rather than the standalone `try-runtime-cli` because that links only `sp_io::SubstrateHostFunctions` and cannot resolve our `ledger_*_bridge` host functions, so the upgrade traps before any check runs. Snapshot *creation* still uses `try-runtime-cli`, which needs none.

**Three migrations it surfaced** against live `preview`/`preprod`/`mainnet` state:

| gap | on-chain today | consequence if shipped as-is |
|---|---|---|
| `Session` storage version | 0, pallet declares 1 | version-gated migrations mislead later |
| `Bridge::MainChainScriptsConfiguration` | pre-#1513 3-field value | fails to decode, **silently** (`OptionQuery`) — bridge looks unconfigured |
| `Babe::EpochConfig` | unset | fails babe's `try_state`; panics the pallet on the AURA→BABE flip |

`pallet_session::historical` needs nothing — its prefix holds no keys on any live network, so `BeforeAllRuntimeMigrations` initializes its version.

## Notes for review

**Why `TryRuntime::on_runtime_upgrade` is custom.** Upstream's `Executive::try_runtime_upgrade` asserts every pallet's on-chain storage version right after `on_runtime_upgrade`, which a pallet carried by an MBM cannot satisfy yet — cnight-observation is at in-code version 2 and only gets there once its MBMs drain. Same steps as upstream in the same order, with the drain where the chain would do it and the pallet checks deferred past it. The pallet hooks therefore run twice; the only `Hooks::on_runtime_upgrade` in this runtime is pallet-midnight's `ensure_storage_initialized()`, which is idempotent.

**Why the snapshot reader is hand-rolled.** Adding `frame-remote-externalities` re-resolves `bip39 2.2.2` from `rand_core 0.6.4` onto `rand_core 0.4.2`, and `pallas-wallet` then fails to build with E0277 `CryptoRng` mismatches. Verified, not assumed.

**Scope limit — ledger state.** A snapshot carries substrate storage only; the ledger arena lives in its own parity-db. So on this release train the live-network dry-run gets as far as `pallet_midnight::migrations::v2` (the v8→v9 translation) and stops there with an actionable message. That migration belongs to the fork-testing flow (#1522, `docs/fork-testing.md`), which restores a whole node data directory; `--ledger-db <data>/ledger_storage` points this command at the same directory for the full run locally.

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [x] If the changes introduce a new feature, I have bumped the node minor version — N/A, the version is set by the release PR (#2085)
- [x] No new todos introduced

## 🧪 Testing Evidence

**Unit** — 7 tests in `runtime/src/migrations.rs`, including the exact `MainChainScriptsConfiguration` bytes mainnet holds today: the legacy value does not decode as the new type, the migration defaults the new field, it is idempotent, it leaves a current-layout value alone, and it no-ops when unset.

**End to end, local dev chain with a real ledger arena** — exit 0:

```
$ midnight-node try-runtime --snap dev.snap --runtime midnight_node_runtime.compact.compressed.wasm \
    --ledger-db <base-path>/ledger_storage --disable-spec-version-check --checks all
🔁 re-running pallet hooks with checks
🚚 MBM queue drained in 1 + 1 step(s)
🔍 decoded 3579 storage keys
🩺 Running "System" try-state checks        ... (33 pallets)
Migration consumed Weight { ref_time: 46149442000, proof_size: 290526 } of max block weight ...
```

**Live snapshots (preview / preprod / mainnet)** — identical on all three: the migrations here apply, then the run stops at the ledger translation as described above.

```
🐥 New pallet "Historical" detected ... Initializing the on-chain storage version to ... StorageVersion(1)
🚚 Pallet "Session" VersionedMigration migrating storage version from 0 to 1.
🚚 Bridge::MainChainScriptsConfiguration migrated to the post-#1513 layout
🚚 Babe::EpochConfig initialized to BABE_GENESIS_EPOCH_CONFIG
```

## 🔱 Fork Strategy

- [x] Node Runtime Update — three one-shot migrations, removable once the upgrade carrying them has landed on every network.
- [x] Node Client Update — new subcommand, compiled only under `--features try-runtime`.
- [ ] Other:
- [ ] N/A

## Links

Earlier review comments are addressed: the per-network dry-runs now share one build and run in parallel, `with_allow_missing_host_functions` is gone, and the `pallets/throttle` test change is no longer in the diff.


gilescope (dismissed):

.github/workflows/try-runtime-upgrade.yml
```diff
@@ -0,0 +1,79 @@
...
+        # TODO: Remove before merge.
+        if: ${{ !cancelled() && (github.event_name == 'push' || inputs.preprod) }}
+        run: |
+          . ./.envrc && earthly -P +try-runtime-dry-run-preprod
+
+      - name: Dry-run on-runtime-upgrade - mainnet
+        # TODO: Remove before merge.
+        if: ${{ !cancelled() && (github.event_name == 'push' || inputs.mainnet) }}
+        run: |
+          . ./.envrc && earthly -P +try-runtime-dry-run-mainnet
```
Klapeyron, line 70: we can try slightly parallelize it, so once we finish the build with `try-runtime` and upload the artifact, then execute this 3 in parallel

node/src/try_runtime.rs
```diff
@@ -0,0 +1,213 @@
...
+	#[arg(long, default_value_t = false)]
+	pub disable_spec_name_check: bool,
+}
+
+impl TryRuntimeCmd {
+	pub fn run(&self) -> sc_cli::Result<()> {
+		let executor: WasmExecutor<HostFunctions> = WasmExecutor::builder()
+			.with_onchain_heap_alloc_strategy(DEFAULT_HEAP_ALLOC_STRATEGY)
+			.with_offchain_heap_alloc_strategy(DEFAULT_HEAP_ALLOC_STRATEGY)
+			.with_allow_missing_host_functions(true)
```
Klapeyron: is this line necessary?

pallets/throttle/src/tests.rs
```diff
@@ -127,6 +135,28 @@ fn account_usage_is_independent_per_account() {
...
+				.reads_writes(2, 2)
+		);
+		assert_eq!(Pallet::<Test>::on_chain_storage_version(), 1);
+		assert!(sp_io::storage::get(&key).is_none());
+	});
+}
+
 // ---------------------------------------------------------------------------
 // TransactionExtension::validate() tests
 // ---------------------------------------------------------------------------
```
Klapeyron, line 132: this changes can be removed just in case

jacek-kurkowski-shielded:
Blocked due to: https://github.com/shieldedtech/shielded-sre/issues/238

Klapeyron:
@jacek-kurkowski-shielded I see https://github.com/shieldedtech/shielded-sre/issues/238 is closed, does it unblock this PR?

jacek-kurkowski-shielded:
> @jacek-kurkowski-shielded I see [shieldedtech/shielded-sre#238](https://github.com/shieldedtech/shielded-sre/issues/238) is closed, does it unblock this PR?

Oh thanks! Completely forgot about it.

PR #1538, opened 2026-05-19
fix: remove dockerhub

gilescope:
Remove dockerhub for now.


ozgb (approved):
OK for now - we'll need to add this back in once we've fixed our dockerhub auth issues

PR #1540, opened 2026-05-19
fix: align weight mechanisms

gilescope:
Ensure weight mechanism is not assuming it has to fill 100% of block.


gilescope:
@codex review

justinfrevert (approved):

PR #1541, opened 2026-05-19
feat: add governance batching via vendored pallet-batch

### LINKED ISSUE #1143: [Governance] Add `pallet-utility` to the Runtime

ozgb:
# Overview

Adds a new pallet, `pallet_batch`, a subset of `pallet_utility` that adds only batching functionality. 

**Rationale:** `pallet_utility` allows callers to `dispatch_as` a different origin, allowing governance to side-step restrictions on what root can/cannot call. We want to provide batching functionality, but don't want the increased call surface provided by the utility pallet.

The branch contains three layers of work:

1. **`pallet-batch` (this PR's final commit).**
  - A minimal in-tree pallet at runtime index 20 
  - Exposes only `batch` and `batch_all`, copied verbatim from upstream `pallet-utility` (`polkadot-stable2603`)
  - Restricted to Root origin only
  - Removed: `dispatch_as`, `as_derivative`, `force_batch`, `with_weight`, `if_else`, and the `PalletsOrigin` associated type that exists solely to support `dispatch_as`.

3. **Toolkit additions.** 
  - `update-ledger-parameters` and `runtime-upgrade` gain `--encode-only`, which emits the SCALE-encoded inner call as hex and exits without contacting governance. 
  - A new `batch` subcommand wraps one or more `--encoded-call` blobs in `Batch::batch_all` (or `Batch::batch` with `--allow-partial-failure`) and forwards through `root_call::execute`. The apply-step of `runtime-upgrade` is factored into a new `apply-authorized-upgrade` subcommand. End-to-end the composition is:

   ```
   HEX1=$(toolkit update-ledger-parameters --encode-only ...)
   HEX2=$(toolkit runtime-upgrade --encode-only --wasm-file foo.wasm ...)
   toolkit batch --encoded-call \"\$HEX1\" --encoded-call \"\$HEX2\" -t //Alice -c //Dave
   toolkit apply-authorized-upgrade --wasm-file foo.wasm
   ```

5. **E2E test.** `scripts/tests/toolkit-batch-governance-e2e.sh` boots a dev node, fetches the live runtime WASM via JSON-RPC (so the upgrade is a no-op same-bytes swap that still exercises the authorize → apply → `CodeUpdated` path), runs the four-step composition, and asserts both that ledger params changed and the upgrade applied. Wired into the `just` recipe set and a new CI job.

Issue: https://github.com/midnightntwrk/midnight-node/issues/1143

## 🗹 TODO before merging

- [ ] Rebuild runtime metadata (`/bot rebuild-metadata`) to capture the
      `Utility` → `Batch` pallet rename and the reduced call set
- [x] Update PR link in `changes/runtime/added/add-batch-pallet.md` once
      this PR is opened
- [ ] CI must pass the new `toolkit-batch-governance-e2e` job at least once
      against a built node + toolkit image

## 📌 Submission Checklist

- [x] All commits are signed off (\`git commit -s\`) for the DCO
- [ ] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

- [x] Additional tests are provided (if possible): new unit and integration tests

## 🔱 Fork Strategy

- [x] Node Runtime Update
- [ ] Node Client Update
- [ ] Other:
- [ ] N/A

## Links

Closes #1143


pallets/batch/src/weights.rs
```diff
@@ -0,0 +1,73 @@
...
+	weights::{Weight, constants::RocksDbWeight},
+};
+
+/// Weight functions needed for `pallet_batch`.
+pub trait WeightInfo {
+	fn batch(c: u32) -> Weight;
+	fn batch_all(c: u32) -> Weight;
+}
+
+/// Weights for `pallet_batch` using the Substrate node and recommended hardware.
```
ozgb, line 36: Need to run benchmarking - these weights are placeholder

m2ux:
## PR Review Summary

**PR**: #1541 — feat: add governance batching via vendored pallet-batch
**Date**: 2026-07-01

### Executive Summary

This is a lean, well-scoped change that delivers atomic governance batching by vendoring a hardened, Root-only two-call subset of `pallet-utility` rather than adopting the full pallet. That design choice is sound and, in fact, safer than the issue's literal ask: the origin-forging surface (`dispatch_as` and friends) is closed by construction, not merely filtered. The change is held back from approval by one confirmed blocker the author already self-flagged (placeholder weights wired live into the runtime), a set of release-readiness and correctness gates, and a red CI. None of the findings is chain-halting, and all are addressable without a redesign — but three of them are gates that must close before merge.

**Overall Rating**: Request Changes

---

### Prior Feedback Triage

Disposition of every prior comment and review on the PR, determined before independent analysis. No external human reviewers and no bot findings are present; the only substantive prior signal is the author's own placeholder-weights note.

| # | Prior Comment | Author | Disposition | Reasoning |
|---|---------------|--------|-------------|-----------|
| [1](https://github.com/midnightntwrk/midnight-node/pull/1541#discussion_r3267369031) | Need to run benchmarking — these weights are placeholder | ozgb | Confirmed | Self-flagged and unaddressed at the head commit. The placeholder weights ship wired into the runtime config; this caps the overall rating at Request Changes. |
| [2](https://github.com/midnightntwrk/midnight-node/pull/1541) | Empty PR-level review (state COMMENTED, no body) | ozgb | Refuted | No content — carries no reviewable concern. |

<details>
<summary>Finding Details</summary>

#### PF-1. Placeholder weights not benchmarked (Confirmed — blocker)
The author flagged on the weights file that the batch/`batch_all` weights are placeholders needing benchmarking. This remains true at the head commit and is confirmed independently below as CR-1. Extrinsic weights govern block resource accounting, so shipping placeholder values wired into the live runtime config is a production-readiness gap.
**Disposition**: Confirmed — caps the Overall Rating at Request Changes.

#### PF-2. Empty PR-level review (Refuted)
An author self-review with an empty body and no inline threads. Nothing to address.
**Disposition**: Refuted.

</details>

---

### Code Review Findings

Line references validated against the source at head `a486b81c`. Only non-passing findings are listed; the positives are summarized after the Action Items.

| # | Finding | Severity |
|---|---------|----------|
| [1](https://github.com/midnightntwrk/midnight-node/blob/a486b81c91ed8cd11f465ceac41f7ffcb7660a8e/pallets/batch/src/weights.rs#L17-L19) | Placeholder weights not re-benchmarked; wired live into the runtime config | Blocker |
| [2](https://github.com/midnightntwrk/midnight-node/blob/a486b81c91ed8cd11f465ceac41f7ffcb7660a8e/pallets/midnight/src/lib.rs#L501) | `batch_all` is not atomic over ledger effects — a partially-applied `send_mn_root_transaction` returns Ok and commits | High |
| [3](https://github.com/midnightntwrk/midnight-node/blob/a486b81c91ed8cd11f465ceac41f7ffcb7660a8e/util/toolkit/src/commands/batch.rs#L108) | Best-effort `batch` is reachable by governance; its atomicity default is a toolkit convention only, not runtime-enforced | High |
| [4](https://github.com/midnightntwrk/midnight-node/blob/a486b81c91ed8cd11f465ceac41f7ffcb7660a8e/runtime/src/lib.rs#L281) | `spec_version` not bumped for the new pallet, calls, and events (Metadata Check CI is red as a result) | High |
| [5](https://github.com/midnightntwrk/midnight-node/blob/a486b81c91ed8cd11f465ceac41f7ffcb7660a8e/util/toolkit/src/fetcher/runtimes.rs#L150) | Root-event decoder uses an unchecked slice; a short or malformed event body panics instead of returning None | Medium |
| [6](https://github.com/midnightntwrk/midnight-node/blob/a486b81c91ed8cd11f465ceac41f7ffcb7660a8e/pallets/batch/src/lib.rs#L104-L134) | Nested batching has no explicit depth guard and the behavior is undocumented (bounded and Root-gated) | Medium |
| [7](https://github.com/midnightntwrk/midnight-node/blob/a486b81c91ed8cd11f465ceac41f7ffcb7660a8e/runtime/src/lib.rs#L986) | Misplaced `// Batch` section comment — sits above the Preimage pallet at index 15, not the Batch pallet at index 20 | Low |
| [8](https://github.com/midnightntwrk/midnight-node/blob/a486b81c91ed8cd11f465ceac41f7ffcb7660a8e/util/toolkit/src/commands/batch.rs#L87-L88) | Unreachable empty-calls guard and its error variant (the clap arg is already required with at least one value) | Low |

<details>
<summary>Finding Details</summary>

#### CR-1. Placeholder weights not benchmarked (Blocker)
The four weight literals in the batch pallet's `SubstrateWeight` are copied verbatim from upstream `pallet-utility` and have not been re-benchmarked against this runtime; the module header says so and the author self-flagged it (PF-1). They are wired live into the runtime config. The severity is bounded — these numbers govern only the per-batch overhead term, not the inner calls' own benchmarked weights, which dominate a real governance batch — and the benchmarking harness is already wired into the runtime's benchmarking feature, so only the generated output is missing. But it is a hard gate.
**Suggestion**: Run the standard benchmarking flow over `batch`/`batch_all` and replace the four literals. No code change is needed — the weight expressions and the unit-test weight assertions reference `WeightInfo::batch(..)` symbolically, so regenerated numbers will not break the tests. (Benchmarking was not run in this review — the environment cannot compile the workspace; it is deferred to the author/CI.)

#### CR-2. `batch_all` is not atomic over ledger effects (High)
`batch_all` rolls back only when an inner call returns an error. But `send_mn_root_transaction` returns Ok even when the ledger only partially applied — it advances the state root and emits `RootTxPartialSuccess` rather than returning an error. So a `batch_all` containing a partially-applying root transaction commits that partial ledger state; the atomic wrapper cannot see an Ok-encoded partial failure. This is reachable on the recommended, atomic path, which is what makes it the non-obvious case. There is a genuine design tension here: the partial-success events exist precisely so the off-chain indexer can recover the transaction regardless of how far it applied, so making this path roll back would forfeit that observability.
**Suggestion**: Make an explicit decision. Either `send_mn_root_transaction` returns an error on partial application (forfeiting the partial-success observability), or the guarantee is documented as "atomic over inner-call dispatch results, not over ledger effects" so operators do not read `batch_all` as end-to-end atomic. Either way, add the integration test noted in TR-1.

#### CR-3. Best-effort `batch` reachability is a toolkit convention only (High)
`batch` (best-effort) and `batch_all` (atomic) are both first-class Root calls reachable by the identical governance path; nothing in the runtime distinguishes them. The atomic default is expressed only in the toolkit, by a call-name string chosen from `--allow-partial-failure`. A motion composed outside the toolkit can select `batch` directly, and on a mid-batch inner failure `batch` keeps the prior writes and returns Ok — leaving exactly the half-applied state this feature set out to retire, while the governance log still records the motion as dispatched successfully.
**Suggestion**: Record an explicit decision on whether governance should be able to reach best-effort `batch` at all. If not, drop the `batch` call and keep only `batch_all`, or gate it with a runtime call filter. If it should remain, document the operator responsibility to choose `batch_all` for changes that must be atomic.

#### CR-4. `spec_version` not bumped (High)
The PR adds pallet index 20 with new calls and events, and `pallet-midnight` gains a new call and two event variants — all of which change the runtime's generated metadata. The runtime version constant is untouched, so `spec_version` stays at its current value. FRAME's `set_code` requires the incoming `spec_version` to be strictly greater than the on-chain value, so this runtime would be rejected at upgrade time. This is exactly what turns the Metadata Check CI job red. The transaction version does not strictly need to change (the change is additive and no existing call index is renumbered).
**Suggestion**: Bump `spec_version` before the runtime ships as an on-chain upgrade. The draft status makes deferral to the release cut plausible, but the failure is silent until `set_code`, so it is a gate rather than a nicety.

#### CR-5. Unchecked slice in the Root-event decoder (Medium)
The new `root_tx_applied` decoder recovers the payload by slicing off the 32-byte transaction-hash prefix and decoding the rest. The slice is unchecked: if the event body is shorter than 32 bytes (a truncated or malformed event, or a future runtime whose variant name matches but whose layout differs), it panics out of bounds instead of yielding None — contradicting the function's own documented contract that it returns None for runtimes lacking these events. This is off-chain toolkit/indexer code, so a panic aborts the fetcher, not the chain.
**Suggestion**: Use a checked slice, e.g. `field_bytes.get(32..)?`, so a short or unexpected body degrades to None.

#### CR-6. Nested-batch depth is unguarded and undocumented (Medium)
A batch may contain a batch — `Batch` is a legal inner call and each nesting re-runs `ensure_root` on the same Root origin. Depth is bounded only by the per-level call limit and the decode/size ceiling; there is no explicit depth guard. This is inherited faithfully from upstream `pallet-utility` and is bounded and Root-gated, so it is not a denial-of-service vector, but governance is the only caller and the property is unstated.
**Suggestion**: Add an integrity note (and optionally a depth cap) documenting the nesting behavior on this Root-reachable path.

#### CR-7. Misplaced section comment (Low)
The diff retitles a section comment from `// Utility` to `// Batch`, but that comment sits directly above the Preimage pallet at index 15, not above the Batch pallet at index 20. Cosmetic only — comments do not affect metadata or dispatch — but a reader scanning the runtime for the batch pallet is pointed at the wrong index.
**Suggestion**: Move `// Batch` to the index-20 registration and restore or drop the stray comment above Preimage.

#### CR-8. Unreachable empty-calls guard (Low)
The toolkit batch command opens with a guard returning an error on an empty call list, but the clap argument is declared required with at least one value, so the guard and its error variant are unreachable dead code.
**Suggestion**: Remove the guard and the unused error variant.

</details>

---

### Test Review Findings

| # | Gap | Severity |
|---|-----|----------|
| 1 | No test exercises `batch_all` over a partially-applying ledger transaction — the one case that tests the ledger-level atomic guarantee | High |
| [2](https://github.com/midnightntwrk/midnight-node/blob/a486b81c91ed8cd11f465ceac41f7ffcb7660a8e/scripts/tests/toolkit-batch-governance-e2e.sh) | The single end-to-end test drives only the happy `batch_all` path; failure and live root-transaction paths are not exercised end-to-end | Medium |

<details>
<summary>Finding Details</summary>

The unit tests are clean and high-value — 11 tests across the batch and midnight pallets, no anti-patterns, covering Root-only entry, the filter bypass, best-effort versus atomic semantics, weight refunds, the call limit, and the unsigned-only fencing of `send_mn_transaction`. The gaps are coverage boundaries, not broken tests.

#### TR-1. Ledger-level atomicity is untested (High)
No test exercises `batch_all` (or a bare `send_mn_root_transaction`) over a ledger transaction that only partially applies. Because that path returns Ok and emits the partial-success event, `batch_all` commits the partial state — so this is the one test that falsifies the intuitive "`batch_all` equals atomic" reading (see CR-2). It is currently unreachable because the two mocks are tested in isolation: the batch mock uses balance and system inner calls, and the midnight mock never composes a batch, so neither harness can reach the composed path.
**Suggestion**: Add an integration test with a mock (or runtime harness) that composes the batch pallet over `pallet-midnight`: a `batch_all` containing a single partially-applying `send_mn_root_transaction`, asserting the extrinsic returns Ok, the partial-success event is deposited, the state advances, and the storage layer commits (i.e. no rollback).

#### TR-2. Failure and root-transaction paths untested end-to-end (Medium)
The single end-to-end test drives only a two-call `batch_all` happy path through one motion. Not exercised end-to-end: a best-effort `batch` partial failure through a live motion; a `batch_all` inner-error rollback on a live node; and a live `send_mn_root_transaction` with the payload recovered by the fetcher (which would also exercise the CR-5 decoder path).
**Suggestion**: Extend the end-to-end test to cover a `batch_all` inner-failure rollback and, ideally, a best-effort `batch` partial failure, asserting on-chain state and the observability difference between an interrupted batch and a failed motion. Add a unit test feeding the root-event decoder a body shorter than 32 bytes so the CR-5 panic becomes a caught regression.

</details>

---

### Validation Findings

No local compilation was run — the review environment cannot compile the workspace (it exhausts memory on cargo), so the four workspace checks (check, clippy, test, fmt) are deferred to CI/author, not asserted here. The authoritative suite signal is the PR's own CI at head `a486b81c`. Four jobs are red; each traces to a finding above rather than to a broken unit test. The pass set (Run tests, both node-and-image builds, Test Pallet Fixtures, formatting/linting, and every other toolkit end-to-end job) establishes that the workspace compiles, lints, and passes unit tests at head.

| # | Check | Severity |
|---|-------|----------|
| [1](https://github.com/midnightntwrk/midnight-node/actions/runs/26578830652/job/78308977527) | Metadata Check fails — runtime metadata changed but `spec_version` not bumped | High |
| [2](https://github.com/midnightntwrk/midnight-node/actions/runs/26578830652/job/78308977354) | Toolkit Batch Governance E2E fails — the PR's own new end-to-end test | Medium |
| [3](https://github.com/midnightntwrk/midnight-node/actions/runs/26578830652/job/78308977450) | Test Toolkit killed (exit 137, ~2h) — an infrastructure/timeout signal, not an assertion failure | Warning |

<details>
<summary>Finding Details</summary>

#### VF-1. Metadata Check fails (High)
The runtime metadata built from the PR head differs from the committed reference, and `spec_version` was not bumped. The metadata-diff gate fires exactly on this drift. This is CR-4 confirmed by CI — not a test defect; the check correctly reports an un-regenerated, un-versioned metadata change.
**Suggestion**: Bump `spec_version` and rebuild the reference metadata (the PR's own TODO notes the metadata-rebuild step). Resolves with CR-4.

#### VF-2. Toolkit Batch Governance E2E fails (Medium)
The PR's new batch-governance end-to-end test fails. It exercises the toolkit Root-event decoder (the CR-5 unchecked-slice path) and depends on the runtime upgrade that the un-bumped `spec_version` (CR-4) would gate. Reproduction needs a built node image plus toolkit, so it was not run here; statically, the un-bumped `spec_version` and the decoder slice are the candidate triggers.
**Suggestion**: Re-run after addressing CR-4 and CR-5; if it still fails, it points to a live-path defect the happy-path unit coverage does not reach (see TR-2).

#### VF-3. Test Toolkit killed (Warning)
Exit 137 means the process was killed (out-of-memory or timeout) after roughly two hours, rather than failing an assertion. This looks like an infrastructure/timeout signal, not a logic defect, but it is distinct from a clean pass.
**Suggestion**: Re-run the job on CI to distinguish an infrastructure flake from a real hang.

</details>

---

### Branch Hygiene

| # | Item | Severity |
|---|------|----------|
| [1](https://github.com/midnightntwrk/midnight-node/pull/1541/commits) | DCO check fails — commits lack a sign-off; the PR checklist claims commits are signed off | High |
| 2 | Branch is 69 commits behind main (32 ahead) | Warning |

<details>
<summary>Finding Details</summary>

#### BH-1. DCO fails and the checklist overstates it (High)
The DCO check fails because the commits lack a `Signed-off-by` line. Separately, the PR's submission checklist ticks "all commits are signed off," which contradicts the failing gate. This is the author's to resolve on their own commits (a sign-off amend and force-push) — flagged here, not acted on.
**Suggestion**: Sign off the commits (`git commit --amend -s` / rebase with sign-off) and correct the checklist to match the actual state.

#### BH-2. Branch freshness (Warning)
The branch is 69 commits behind `origin/main` (32 ahead). Not blocking, but a rebase before merge reduces integration risk.
**Suggestion**: Rebase onto the latest main before merge.

</details>

---

### Action Items

**Must Address (Blocking)**:
- [ ] Re-benchmark `batch`/`batch_all` and replace the placeholder weights (PF-1, CR-1)
- [ ] Bump `spec_version` and rebuild reference metadata before shipping as an on-chain upgrade (CR-4, VF-1)
- [ ] Make an explicit decision on ledger-level atomicity of `batch_all` over a partially-applying root transaction — return an error on partial apply, or document the boundary (CR-2)
- [ ] Record an explicit decision on best-effort `batch` reachability by governance — drop it, gate it, or document the operator responsibility (CR-3)
- [ ] Add the integration test for `batch_all` over a partially-applying `send_mn_root_transaction` (TR-1)
- [ ] Sign off the commits so the DCO check passes, and correct the submission checklist (BH-1)

**Should Address (Recommended)**:
- [ ] Use a checked slice in the Root-event decoder so a short body degrades to None (CR-5)
- [ ] Extend end-to-end coverage to failure and live root-transaction paths; add a short-body decoder unit test (TR-2, VF-2)
- [ ] Re-run the Test Toolkit job to distinguish an infrastructure flake from a real hang (VF-3)

**Could Address (Suggested)**:
- [ ] Document nested-batch behavior (and optionally add a depth cap) on the Root-reachable path (CR-6)
- [ ] Rebase onto the latest main before merge (BH-2)

**Nice to Have (Optional)**:
- [ ] Move the `// Batch` section comment to the index-20 registration (CR-7)
- [ ] Remove the unreachable empty-calls guard and its error variant (CR-8)

---

### What This PR Gets Right

Worth stating alongside the findings, because it frames the verdict:

- The origin-forging surface is closed by construction. The vendored pallet exposes only `batch` and `batch_all`, both Root-only; inner calls are dispatched under the caller's unchanged Root origin; and `dispatch_as`, `as_derivative`, and the `PalletsOrigin` type are absent, so there is no site that could forge a signed or unsigned origin. `batch` is also not whitelisted as a top-level submitted call. The security claim holds by enumeration, not merely by filtering.
- Vendoring a hardened, Root-only subset instead of adopting the full `pallet-utility` is a sound, security-motivated refinement of the issue's literal ask. FRAME offers no per-call exposure switch, so vendor-and-delete is the only design-time-safe way to drop the high-surface calls; a runtime filter over the full pallet would be weaker. The one carried cost is that the vendored copy is frozen against upstream and must be watched by hand — worth surfacing as an owned responsibility.
- The change is lean and focused: the shared apply-body extraction and the composition-driven encode refactor are reuse at the right level, and every changed file maps to the feature. The unit tests are clean and high-value.

---

*This review was prepared by an AI agent (Claude) operating under the work-package review workflow, on behalf of mike.clay@shielded.io. Builds, tests, and benchmarks were deferred to CI — nothing was compiled or run locally, and no benchmark numbers are asserted here.*

Issue #1531, opened 2026-05-18
Preview node stuck at block #623104 on version 1.0.0-rc.8


TranHuyHiep:
**Description**

- Preview node is currently stuck at block #623104 and is no longer syncing properly.
- The node occasionally shows syncing messages, but it keeps returning to idle state with 0 peers and never progresses past block #623306. Finalized block remains stuck at #623104.

**Version**: 1.0.0-rc.8

**Logs**
```
2026-05-18 21:37:33 💤 Idle (0 peers), best: #623306 (0x2957…8512), finalized #623104 (0x9e87…ebc4), ⬇ 0.2kiB/s ⬆ 0.3kiB/s

2026-05-18 21:37:38 ⚙️ Syncing 0.0 bps, target=#789621 (1 peers), best: #623306 (0x2957…8512), finalized #623104 (0x9e87…ebc4), ⬇ 2.7kiB/s ⬆ 0.7kiB/s

2026-05-18 21:37:43 💤 Idle (0 peers), best: #623306 (0x2957…8512), finalized #623104 (0x9e87…ebc4), ⬇ 0.4kiB/s ⬆ 0.4kiB/s

2026-05-18 21:37:48 ⚙️ Syncing 0.0 bps, target=#789622 (1 peers), best: #623306 (0x2957…8512), finalized #623104 (0x9e87…ebc4), ⬇ 1.5kiB/s ⬆ 0.6kiB/s

2026-05-18 21:37:53 💤 Idle (0 peers), best: #623306 (0x2957…8512), finalized #623104 (0x9e87…ebc4), ⬇ 0.2kiB/s ⬆ 0.2kiB/s
```
**Expected behavior**

- The node should continue syncing normally past block #623104.

**Actual behavior**

- The node remains stuck and continuously switches between Idle (0 peers) and Syncing state without making progress.

justinfrevert:
We'll do some digging into any potential version incompatibility. Thanks for the report.

In the meantime, if you do need to sync, I recommend trying the version that's currently reported `0.22.5`.

TranHuyHiep:
@justinfrevert  Thank you for your reply.

I tried updating to version `0.22.5` before switching to version `1.0.0-rc.8`, because I thought the newer version would fix this issue. However, the bug still persists.

Is there any temporary workaround or solution for this issue?

justinfrevert:
> [@justinfrevert](https://github.com/justinfrevert) Thank you for your reply.
> 
[...]

PR #1436, opened 2026-04-28
Sync Perf Part 2

gilescope:
# Overview

feat: syncs to tip of mainnet from genesis in 3h 19m (572k blocks, 48bps). (needs tidying)

## 🗹 TODO before merging

- [x] Tidy up so not referring to bulk files any more.
- [x] Rather than one large bulk snapshot, can we do a series of them?
- [x] Refactor code
- [x] Try additional speedups now ledger has less bugs. ( Saved in `giles-pre-verifier` branch, so far the juice doesn't warrant the squeeze)

## 📌 Submission Checklist

- [ ] All commits are signed off (`git commit -s`) for the DCO
- [x] Changes are backward-compatible (or flagged if breaking)
- [ ] Pull request description explains why the change is needed
- [ ] Self-reviewed the diff
- [x] I have included a change file, or skipped for this reason: 
- [ ] If the changes introduce a new feature, I have bumped the node minor version
- [ ] Update documentation (if relevant)
- [ ] Updated AGENTS.md if build commands, architecture, or workflows changed
- [ ] No new todos introduced

## 🧪 Testing Evidence

- [x] Synced mainnet from genesis to tip

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [x] Other: client-side perf only — no runtime/protocol changes
- [ ] N/A

PR #1546, opened 2026-05-20
Improve query behavior and sync speed

justinfrevert:
# Overview

Addresses slow syncing behavior in https://github.com/midnightntwrk/midnight-node/issues/1531. ~Synced from ~ 622000-624000(past the reported problematic 623104) to test.~(Outdated comment - reduced some query changes)

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [x] Node Client Update
- [ ] Other: 
- [ ] N/A

### DIFF
--- a/primitives/mainchain-follower/src/data_source/candidates_data_source/db_model.rs
+++ b/primitives/mainchain-follower/src/data_source/candidates_data_source/db_model.rs
@@ -286,23 +286,36 @@ pub(crate) async fn get_token_utxo_for_epoch(
 	ident: i64,
 	epoch: EpochNumber,
 ) -> Result<Option<TokenTxOutput>, SqlxError> {
-	let sql = "SELECT
-			origin_tx.hash        AS origin_tx_hash,
-        	tx_out.index          AS utxo_index,
-        	origin_block.epoch_no AS tx_epoch_no,
-        	origin_block.block_no AS tx_block_no,
-        	origin_block.slot_no  AS tx_slot_no,
-        	origin_tx.block_index AS tx_block_index,
-        	datum.value           AS datum
-        FROM ma_tx_out
-        INNER JOIN tx_out               ON ma_tx_out.tx_out_id = tx_out.id
-        INNER JOIN tx origin_tx         ON tx_out.tx_id = origin_tx.id
-        INNER JOIN block origin_block   ON origin_tx.block_id = origin_block.id
-        LEFT JOIN datum                 ON tx_out.data_hash = datum.hash
-        WHERE ma_tx_out.ident = $1
-        AND origin_block.epoch_no <= $2
-        ORDER BY tx_block_no DESC, origin_tx.block_index DESC
-        LIMIT 1";
+	// Force PostgreSQL to start from the highly-selective `ma_tx_out.ident` predicate before
+	// sorting, instead of walking blocks backwards and probing every candidate tx_out.
+	let sql = "WITH candidate_outputs AS MATERIALIZED (
+			SELECT
+				origin_tx.hash        AS origin_tx_hash,
+        		tx_out.index          AS utxo_index,
+        		origin_block.epoch_no AS tx_epoch_no,
+        		origin_block.block_no AS tx_block_no,
+        		origin_block.slot_no  AS tx_slot_no,
+        		origin_tx.block_index AS tx_block_index,
+        		tx_out.data_hash      AS data_hash
+        	FROM ma_tx_out
+        	INNER JOIN tx_out             ON ma_tx_out.tx_out_id = tx_out.id
+        	INNER JOIN tx origin_tx       ON tx_out.tx_id = origin_tx.id
+        	INNER JOIN block origin_block ON origin_tx.block_id = origin_block.id
+        	WHERE ma_tx_out.ident = $1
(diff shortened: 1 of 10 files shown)


primitives/mainchain-follower/src/data_source/cnight_observation.rs
```diff
@@ -138,18 +158,14 @@ impl MidnightCNightObservationDataSource for MidnightCNightObservationDataSource
...
-			)
-			.await?;
-
-		let cnight_ident = self
-			.multi_asset_cache
-			.resolve_ident(&config.cnight_policy_id, cnight_asset_name)
-			.await?;
+			),
+			self.multi_asset_cache.resolve_ident(&config.cnight_policy_id, cnight_asset_name),
+		)?;
```
LGLO: Besides the first call it will be access to memory. Should this really be parallelized? Seems a bit of overkill.

primitives/mainchain-follower/src/data_source/cnight_observation.rs
```diff
@@ -113,6 +117,22 @@ impl MidnightCNightObservationDataSource for MidnightCNightObservationDataSource
 	) -> Result<ObservedUtxos, Box<dyn std::error::Error + Send + Sync>> {
 		let cnight_asset_name = config.cnight_asset_name.as_bytes();
 
+		// Memoize by (current_tip, start_block, start_tx_index). config and tx_capacity are
+		// effectively constant per node, so they aren't part of the key. Hot path for sync: many
+		// Midnight blocks reference the same Cardano tip and we observed 8-11 redundant calls per
```
LGLO: Each "redundant" call should be a call for an empty range. It should not be possible to have queries for different Substrate blocks to return any overlapping `ObservedUtxos`.
Looking at query bounds we could say that it is a query of effectively empty range.
I think that adding cache here hides the problem of making SQL queries that cannot return a UTXO.
Using cache suggests that we are returning same set of `ObservedUtxos` for different substrate blocks.

primitives/mainchain-follower/src/data_source/candidates_data_source/db_model.rs
```diff
@@ -286,23 +286,36 @@ pub(crate) async fn get_token_utxo_for_epoch(
...
-        INNER JOIN tx_out               ON ma_tx_out.tx_out_id = tx_out.id
-        INNER JOIN tx origin_tx         ON tx_out.tx_id = origin_tx.id
-        INNER JOIN block origin_block   ON origin_tx.block_id = origin_block.id
-        LEFT JOIN datum                 ON tx_out.data_hash = datum.hash
-        WHERE ma_tx_out.ident = $1
-        AND origin_block.epoch_no <= $2
-        ORDER BY tx_block_no DESC, origin_tx.block_index DESC
-        LIMIT 1";
+	// Force PostgreSQL to start from the highly-selective `ma_tx_out.ident` predicate before
+	// sorting, instead of walking blocks backwards and probing every candidate tx_out.
```
LGLO: I can't see how this helps. If anything changed here, I see a risk that `LIMIT 1` is not taken into account when creating `candidate_outputs` and it makes join through whole tx_out table (epoch_no applies).

Fact that I don't see how extracting `candidate_outputs` before join with `datum` helps doesn't mean it doesn't help of course. So, I won't block this extraction but please add `LIMIT 1` to `candidate_outputs`.

primitives/mainchain-follower/src/data_source/cnight_observation.rs
LGLO: I've seen competing proposal for solving CNight observation in another PR that queries the stable part of Cardano and keeps UTXOs in memory, serving them during the sync period and I think it is the way to go.
That approach was used in partner-chains with a success (bridge queries IIRC).

LGLO (commented):
I would love to see this PR trimmed to epoch_nonce and technical committee and council caching - these 3 are good to go for me.

I think that for sync there is a better solution than proposed caching and also fix could be proposed to not make SQL queries that are bound to return empty result (I'll take a look at this one).

I don't see how proposed SQL query change helps.

gilescope:
Is there any crossover with https://github.com/midnightntwrk/midnight-node/pull/1436 ?

justinfrevert:
Reduced the scope 

> I would love to see this PR trimmed to epoch_nonce and technical committee and council caching - these 3 are good to go for me.
> 
> I think that for sync there is a better solution than proposed caching and also fix could be proposed to not make SQL queries that are bound to return empty result (I'll take a look at this one).
> 
> I don't see how proposed SQL query change helps.

Reduced the scope to the caching - no query, so we can approach that work separately. That makes the speedup of this pr very modest. 

It should also be simpler to reconcile with your branch, @gilescope

justinfrevert:
Also, I assume that the priority of this can be lowered, given the plausible workaround for the issue noted. Perhaps this change can just be made to `main`, and addressed maybe even after Giles' sync changes. That PR is here https://github.com/midnightntwrk/midnight-node/pull/1551 FYI: @gilescope @LGLO .

PR #1552, opened 2026-05-21
feat: extended observability of McHashDataSource with additional errors

### LINKED ISSUE #1391: Db-sync lag causes Midnight network outages via banning

Klapeyron:
# Overview

Follow-up on @LGLO comment https://github.com/input-output-hk/partner-chains/pull/1139#issuecomment-3768106604

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A

### DIFF
--- a/node/src/main_chain_follower.rs
+++ b/node/src/main_chain_follower.rs
@@ -24,7 +24,7 @@ use partner_chains_mock_data_sources::{
 	SidechainRpcDataSourceMock, TokenBridgeDataSourceMock,
 };
 use sc_service::error::Error as ServiceError;
-use sidechain_mc_hash::McHashDataSource;
+use sidechain_mc_hash::{BlockByHash, McHashDataSource};
 use sp_partner_chains_bridge::TokenBridgeDataSource;
 use sqlx::{Pool, Postgres};
 
@@ -177,8 +177,9 @@ async fn log_db_sync_startup_probe(block_data_source: &BlockDataSourceImpl) {
 		Err(_) => "query_failed",
 	};
 	let block_lookup_state = match &block_lookup {
-		Some((_, Ok(Some(_)))) => "confirmed",
-		Some((_, Ok(None))) => "missing",
+		Some((_, Ok(BlockByHash::Found(_)))) => "confirmed",
+		Some((_, Ok(BlockByHash::NotFound { .. }))) => "missing",
+		Some((_, Ok(BlockByHash::LocalDataUnavailable { .. }))) => "cardano_is_stale",
 		Some((_, Err(_))) => "query_failed",
 		None => "skipped",
 	};
--- a/partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
+++ b/partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
@@ -34,8 +34,6 @@ enum StableBlockByHashError {
 	BlockNotFound { hash: McBlockHash },
 	#[error("Latest block info was unavailable while checking block hash {0}.")]
 	LatestBlockUnavailable(McBlockHash),
-	#[error("Local Cardano tip {latest_block:?} is too old to classify the referenced hash.")]
-	LocalTipTooOld { latest_block: Block },
 	#[error(
 		"Block with hash {hash} is not stable yet: block {block_no} requires latest block >= {required_latest_block_no}, but latest block is {latest_block_no}."
 	)]
@@ -426,10 +424,41 @@ impl BlockDataSourceImpl {
 				},
 				Err(err) => {
 					warn!("Get stable block by hash failed: {err}");
-					Ok(self.classify_stable_block_error(err, reference_timestamp))
+					let latest_block =
+						db_model::get_latest_block_info(&self.pool).await.ok().flatten();
+					Ok(self.classify_stable_block_lookup_error(
+						err,
+						latest_block,
+						reference_timestamp,
+					))
 				},
 			}
 		}
+	}
+
+	fn classify_stable_block_lookup_error(
+		&self,
+		err: StableBlockByHashError,
+		latest_block: Option<Block>,
+		reference_timestamp: NaiveDateTime,
+	) -> StableBlockForHash {
+		if let Some(latest_block) = latest_block {
+			let is_ambiguous = matches!(
+				err,
+				StableBlockByHashError::BlockNotFound { .. }
+					| StableBlockByHashError::NotStableYet { .. }
+			);
+			if is_ambiguous && !self.is_local_cardano_tip_recent(&latest_block, reference_timestamp)
+			{
+				let max_allowed_timestamp = self.max_allowed_block_time(reference_timestamp);
+				return self.local_tip_too_old(
+					latest_block,
+					max_allowed_timestamp,
+					reference_timestamp,
+				);
+			}
+		}
+		self.classify_stable_block_error(err, reference_timestamp)
 	}
 
 	fn classify_stable_block_error(
@@ -437,15 +466,11 @@ impl BlockDataSourceImpl {
 		err: StableBlockByHashError,
 		reference_timestamp: NaiveDateTime,
 	) -> StableBlockForHash {
-		let max_allowed_timestamp = self.max_allowed_block_time(reference_timestamp);
 		match err {
 			StableBlockByHashError::LatestBlockUnavailable(_hash) => {
 				StableBlockForHash::LocalDataUnavailable {
 					reason: LocalDataUnavailableReason::LatestBlockUnavailable,
 				}
-			},
-			StableBlockByHashError::LocalTipTooOld { latest_block } => {
-				self.local_tip_too_old(latest_block, max_allowed_timestamp, reference_timestamp)
 			},
 			StableBlockByHashError::BlockNotFound { hash } => {
 				StableBlockForHash::BlockNotFound { hash }
@@ -535,16 +560,11 @@ impl BlockDataSourceImpl {
 			return Err(StableBlockByHashError::LatestBlockUnavailable(hash));
 		};
 		self.observe_latest_cardano_block_metrics(&latest_block);
-		let is_local_tip_recent =
-			self.is_local_cardano_tip_recent(&latest_block, reference_timestamp);
 
 		let block = db_model::get_block_by_hash(&self.pool, hash.clone()).await.map_err(|err| {
 			StableBlockByHashError::Database { hash: hash.clone(), error: format!("{err:?}") }
 		})?;
 		let Some(block) = block else {
-			if !is_local_tip_recent {
-				return Err(StableBlockByHashError::LocalTipTooOld { latest_block });
-			}
 			return Err(StableBlockByHashError::BlockNotFound { hash });
 		};
 		self.observe_referenced_cardano_block_metrics(&block);
@@ -556,9 +576,6 @@ impl BlockDataSourceImpl {
 		let is_time_valid = self.is_block_time_valid(&block, reference_timestamp);
 
 		if !is_stable {
-			if !is_local_tip_recent {
-				return Err(StableBlockByHashError::LocalTipTooOld { latest_block });
-			}
 			return Err(StableBlockByHashError::NotStableYet {
 				hash,
 				block_no: block.block_no.0,
--- a/partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
+++ b/partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
@@ -126,9 +126,12 @@ impl BlockDataSourceImpl {
 		Ok(is_block_time_valid)
 	}
 
-	/// Tests if our Cardano view matches Praos chain-density requirements: at least one
-	/// block must exist in the latest `security_parameter` blocks within the allowed
-	/// time window relative to current wall-clock time.
+	/// Checks whether our Cardano view meets Praos chain-density requirements.
+	///
+	/// Over the latest `security_parameter` blocks, Praos requires chain density of at
+	/// least one third of the expected density. That requirement is enforced indirectly:
+	/// we look up the block `security_parameter` below tip and verify its timestamp
+	/// falls within the allowed time window relative to current wall-clock time.
 	///
 	/// Unlike [Self::is_cardano_tip_fresh], this is not a heuristic.
 	pub async fn is_cardano_ok(&self) -> Result<bool, Box<dyn std::error::Error + Send + Sync>> {
@@ -144,6 +147,11 @@ impl BlockDataSourceImpl {
 	/// Returns the latest _stable_ Cardano block from the Db-Sync database that is within
 	/// acceptable bounds from `reference_timestamp`, accounting for the additional stability
 	/// offset configured by [block_stability_margin][Self::block_stability_margin].
+	///
+	/// When no block fits the timestamp window, this returns [LatestStableBlockForTimestamp::NoStableBlockInRange]
+	/// rather than inferring local db-sync lag from tip age. The tip-age heuristic
+	/// ([Self::is_local_cardano_tip_recent]) is reserved for hash lookups during verification,
+	/// where another signal already suggests observability may be incomplete.
 	pub async fn get_latest_stable_block_for(
 		&self,
 		reference_timestamp: Timestamp,
@@ -171,19 +179,6 @@ impl BlockDataSourceImpl {
 		let block = self.get_latest_block(stable, reference_timestamp).await?;
 		Ok(match block {
 			Some(block) => LatestStableBlockForTimestamp::Found(block.into()),
-			None if !self.is_local_cardano_tip_recent(&latest, reference_timestamp) => {
-				LatestStableBlockForTimestamp::LocalDataUnavailable {
-					reason: LocalDataUnavailableReason::LocalTipTooOld {
-						latest_block: latest.into(),
-						max_allowed_timestamp: Self::db_timestamp_to_sp_timestamp(
-							self.max_allowed_block_time(reference_timestamp),
-						),
-						reference_timestamp: Self::db_timestamp_to_sp_timestamp(
-							reference_timestamp,
-						),
-					},
-				}
-			},
 			None => LatestStableBlockForTimestamp::NoStableBlockInRange {
 				max_stable_block_number: McBlockNumber(stable.0),
 				min_allowed_timestamp: Self::db_timestamp_to_sp_timestamp(
@@ -208,7 +203,11 @@ impl BlockDataSourceImpl {
 		self.get_stable_block_by_hash(hash, reference_timestamp).await
 	}
 
-	/// Finds a block by its `hash` and returns its info
+	/// Finds a block by its `hash` and returns its info.
+	///
+	/// Unlike [Self::get_stable_block_for], this does not check whether the block is stable or
+	/// whether its timestamp falls within the allowable window for a Partner Chain reference.
+	/// Callers that need import-time MC reference validation must use `get_stable_block_for`.
 	pub async fn get_block_by_hash(
 		&self,
 		hash: McBlockHash,
@@ -568,9 +567,6 @@ impl BlockDataSourceImpl {
 			});
 		}
 		if !is_time_valid {
-			if !is_local_tip_recent {
-				return Err(StableBlockByHashError::LocalTipTooOld { latest_block });
-			}
 			return Err(StableBlockByHashError::TimestampOutOfRange {
 				hash,
 				block_time: block.time,
--- a/partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
+++ b/partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
@@ -207,17 +207,7 @@ impl BlockDataSourceImpl {
 		};
 		let block_opt = match from_cache {
 			Some(block) => Some(block),
-			None => match db_model::get_block_by_hash(&self.pool, hash.clone()).await {
-				Ok(block) => block,
-				Err(err) => {
-					return Ok(BlockByHash::LocalDataUnavailable {
-						reason: LocalDataUnavailableReason::DatabaseError {
-							context: format!("querying Cardano block {hash}"),
-							error: format!("{err:?}"),
-						},
-					});
-				},
-			},
+			None => db_model::get_block_by_hash(&self.pool, hash.clone()).await?,
 		};
 		Ok(block_opt
 			.map(|block| BlockByHash::Found(block.into()))
--- a/partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
+++ b/partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
@@ -155,17 +155,7 @@ impl BlockDataSourceImpl {
 		reference_timestamp: Timestamp,
 	) -> Result<LatestStableBlockForTimestamp, Box<dyn std::error::Error + Send + Sync>> {
 		let reference_timestamp = BlockDataSourceImpl::timestamp_to_db_type(reference_timestamp)?;
-		let latest = match db_model::get_latest_block_info(&self.pool).await {
-			Ok(latest) => latest,
-			Err(err) => {
-				return Ok(LatestStableBlockForTimestamp::LocalDataUnavailable {
-					reason: LocalDataUnavailableReason::DatabaseError {
-						context: "querying latest Cardano block".into(),
-						error: format!("{err:?}"),
-					},
-				});
-			},
-		};
+		let latest = db_model::get_latest_block_info(&self.pool).await?;
 		let Some(latest) = latest else {
 			return Ok(LatestStableBlockForTimestamp::LocalDataUnavailable {
 				reason: LocalDataUnavailableReason::LatestBlockUnavailable,
--- a/partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
+++ b/partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
@@ -12,24 +12,30 @@ use log::{debug, info, warn};
 use serde::Deserialize;
 use sidechain_domain::mainchain_epoch::{MainchainEpochConfig, MainchainEpochDerivation};
 use sidechain_domain::*;
+use sidechain_mc_hash::{
+	BlockByHash, LatestStableBlockForTimestamp, LocalDataUnavailableReason, StableBlockForHash,
+};
 use sp_timestamp::Timestamp;
 use sqlx::PgPool;
 use std::{
 	error::Error,
 	sync::{Arc, Mutex},
 };
+use time_source::{SystemTimeSource, TimeSource};
 
 #[cfg(test)]
 mod tests;
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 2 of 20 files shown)


Klapeyron:
@codex review

partner-chains/toolkit/sidechain/sidechain-mc-hash/src/lib.rs
```diff
@@ -253,6 +262,157 @@ pub enum McHashInherentError {
 	StableBlockNotFoundByHash(McBlockHash),
 }
 
+/// Result of querying the latest stable Cardano block for a Partner Chain block timestamp.
+#[derive(Clone, Debug, PartialEq)]
+pub enum LatestStableBlockFor {
```
LGLO: Rename to `LatestStableBlockForTimestamp` because we have `LatestStableBlockFor` and  `StableBlockFor`. One "For" is timestamp and other is block hash.

partner-chains/toolkit/sidechain/sidechain-mc-hash/src/lib.rs
```diff
@@ -253,6 +262,157 @@ pub enum McHashInherentError {
...
+	/// A latest stable block was found.
+	Found(MainchainBlock),
+	/// Local Cardano observability is not caught up enough to decide yet.
+	LocalDataUnavailable {
+		/// Why local Cardano data cannot support a final decision yet.
+		reason: LocalDataUnavailableReason,
+	},
+	/// Local Cardano observability is recent enough, but no stable block fits the timestamp range.
+	NotFound {
```
LGLO: Rename to `NoStableBlockInRange`.

partner-chains/toolkit/sidechain/sidechain-mc-hash/src/lib.rs
```diff
@@ -253,6 +262,157 @@ pub enum McHashInherentError {
...
+		/// Latest Cardano block seen locally when the lookup was classified.
+		latest_block: MainchainBlock,
```
LGLO: IMO full block here is too much (if we print block hash as bytes array then for sure it is too much), we only care about this block number. Alternatively we care about the highest block number we were filtering (`latest_block.block_no - security_param`)

partner-chains/toolkit/sidechain/sidechain-mc-hash/src/lib.rs
```diff
@@ -253,6 +262,157 @@ pub enum McHashInherentError {
...
+			Self::Found(block) => Some(block),
+			_ => None,
+		}
+	}
+}
+
+/// Result of checking whether a main-chain block hash is a stable reference for a Partner Chain
+/// block timestamp.
+#[derive(Clone, Debug, PartialEq)]
+pub enum StableBlockFor {
```
LGLO: Please rename to `StableBlockForHash`

partner-chains/toolkit/sidechain/sidechain-mc-hash/src/lib.rs
```diff
@@ -253,6 +262,157 @@ pub enum McHashInherentError {
...
+	LocalDataUnavailable {
+		/// Why local Cardano data cannot support a final decision yet.
+		reason: LocalDataUnavailableReason,
+	},
+	/// Local Cardano observability is recent enough and the referenced hash does not exist.
+	BlockNotFound {
+		/// Referenced Cardano block hash.
+		hash: McBlockHash,
+		/// Latest Cardano block seen locally when the lookup was classified.
+		latest_block: MainchainBlock,
```
LGLO: I think that block_number would carry enough information here.

partner-chains/toolkit/sidechain/sidechain-mc-hash/src/lib.rs
```diff
@@ -253,6 +262,157 @@ pub enum McHashInherentError {
...
+	},
+	/// The referenced block exists, but it does not have enough blocks on top.
+	BlockFoundButNotStable {
+		/// Referenced Cardano block hash.
+		hash: McBlockHash,
+		/// Referenced Cardano block number.
+		block_number: McBlockNumber,
+		/// Latest Cardano block seen locally when the lookup was classified.
+		latest_block: MainchainBlock,
```
LGLO: I think we don't need this piece either. We know its number from the field below.

partner-chains/toolkit/sidechain/sidechain-mc-hash/src/lib.rs
```diff
@@ -253,6 +262,157 @@ pub enum McHashInherentError {
...
+		/// Minimum latest Cardano block number required for this reference to be stable.
+		required_latest_block: McBlockNumber,
+	},
+	/// The referenced block exists, but its timestamp is outside the allowed range.
+	BlockTimestampOutOfRange {
+		/// Referenced Cardano block hash.
+		hash: McBlockHash,
+		/// Referenced Cardano block.
+		block: MainchainBlock,
```
LGLO: We care about timestamp only

partner-chains/toolkit/sidechain/sidechain-mc-hash/src/lib.rs
```diff
@@ -253,6 +262,157 @@ pub enum McHashInherentError {
...
+		/// Latest Cardano block seen locally when the lookup was classified.
+		latest_block: MainchainBlock,
```
LGLO: I hope we don't need to print the whole latest block

partner-chains/toolkit/data-sources/db-sync/src/block/tests.rs
```diff
@@ -64,10 +64,18 @@ async fn test_get_stable_block_at(pool: PgPool) {
 	let security_parameter = 2;
 	let source = mk_datasource(pool, security_parameter);
 	let exact_ts = BLOCK_4_TS_MILLIS.into();
-	let block = source.get_stable_block_for(block_2().hash, exact_ts).await.unwrap();
+	let block = source
+		.get_stable_block_for(block_2().hash, exact_ts)
+		.await
+		.unwrap()
+		.into_option();
```
LGLO: Instead of `.into_option()` match to `Found` variant.

Klapeyron:
@codex review

partner-chains/toolkit/data-sources/db-sync/src/governed_map/mod.rs
```diff
@@ -211,11 +213,12 @@ impl GovernedMapDataSource for GovernedMapDataSourceCachedImpl {
...
-			match self.blocks.get_latest_stable_block_for(latest_block_timestamp.into()).await? {
-				Some(block) => BlockNumber(block.number.0),
-				None => up_to_block_number,
-			};
+		let latest_stable_block_lookup =
+			self.blocks.get_latest_stable_block_for(latest_block_timestamp.into()).await?;
+		let latest_stable_block = latest_stable_block_lookup
+			.into_option()
+			.map(|block| BlockNumber(block.number.0))
+			.unwrap_or(up_to_block_number);
```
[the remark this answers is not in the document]
Klapeyron: removed `into_option` from the code, was misleading https://github.com/midnightntwrk/midnight-node/pull/1552/changes/ea28b6628bdb602894e940d0713e0cbcf6e47281

partner-chains/toolkit/data-sources/db-sync/src/bridge/cache.rs
```diff
@@ -241,6 +241,7 @@ impl CachedTokenBridgeDataSourceImpl {
 			.blocks
 			.get_latest_stable_block_for(latest_block_timestamp.into())
 			.await?
+			.into_option()
 			.map(|block| block.number.into()))
```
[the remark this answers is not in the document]
Klapeyron: removed `into_option` from the code, was misleading https://github.com/midnightntwrk/midnight-node/pull/1552/changes/ea28b6628bdb602894e940d0713e0cbcf6e47281

node/src/main_chain_follower.rs
```diff
@@ -169,8 +169,9 @@ async fn log_db_sync_startup_probe(block_data_source: &BlockDataSourceImpl) {
 		Err(_) => "query_failed",
 	};
 	let block_lookup_state = match &block_lookup {
-		Some((_, Ok(Some(_)))) => "confirmed",
-		Some((_, Ok(None))) => "missing",
+		Some((_, Ok(BlockByHash::Found(_)))) => "confirmed",
+		Some((_, Ok(BlockByHash::NotFound { .. }))) => "missing",
+		Some((_, Ok(BlockByHash::LocalDataUnavailable { .. }))) => "query_failed",
```
LGLO: maybe "cardano_is_stale" ?

partner-chains/toolkit/data-sources/db-sync/src/block/tests.rs
```diff
@@ -150,19 +185,30 @@ async fn test_get_latest_stable_block_with_stability_margin(pool: PgPool) {
...
 	let ts_too_high_for_block_2 = max_ts_for_block_2 + 1;
 	let block = source
 		.get_latest_stable_block_for(ts_too_high_for_block_2.into())
 		.await
 		.unwrap();
-	assert_eq!(block, None);
+	assert_eq!(
+		block,
+		LatestStableBlockForTimestamp::LocalDataUnavailable {
+			reason: LocalDataUnavailableReason::LocalTipTooOld {
```
LGLO: I think this point that logic is invalid.
Does this query look at tip timestamp at all? 
AFAIK it get reference timestamp and checks if there is a block that has `k` blocks on top of it and has timestamp in some range derived from `reference_timestamp`.
We can't judge that `LocalTipIsTooOld` without a current clock.

Klapeyron: Migrated wall-clock check and logic from https://github.com/midnightntwrk/midnight-node/pull/1472/

Klapeyron:
@codex review

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
```diff
@@ -97,19 +108,93 @@ impl BlockDataSourceImpl {
...
+		let is_block_time_valid = self.is_latest_block_time_valid(&block);
+		log::debug!(
+			"is_cardano_tip_fresh block: {block:?}, is_block_time_valid: {is_block_time_valid}"
+		);
+		Ok(is_block_time_valid)
+	}
+
+	/// Tests if our Cardano view matches Praos chain-density requirements: at least one
+	/// block must exist in the latest `security_parameter` blocks within the allowed
+	/// time window relative to current wall-clock time.
```
LGLO: This isn't expressed in words I would use. Praos requires that recent `security_parameters` blocks chain-density is at least 1/3 of expected density. This chain density check is hidden in the check of the timestamp of the block that is `k` below tip.

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
```diff
@@ -97,19 +108,93 @@ impl BlockDataSourceImpl {
...
+	///
+	/// Unlike [Self::is_cardano_tip_fresh], this is not a heuristic.
+	pub async fn is_cardano_ok(&self) -> Result<bool, Box<dyn std::error::Error + Send + Sync>> {
```
LGLO, line 135: I'm wondering if this shouldn't be implemented as `get_latest_stable_block_for(<current_timestamp>).map(Result::is_ok)`

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
```diff
@@ -364,6 +568,9 @@ impl BlockDataSourceImpl {
 			});
 		}
 		if !is_time_valid {
+			if !is_local_tip_recent {
+				return Err(StableBlockByHashError::LocalTipTooOld { latest_block });
+			}
```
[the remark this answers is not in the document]
LGLO: I agree with codex

partner-chains/toolkit/sidechain/sidechain-mc-hash/src/block_announce_validator.rs
```diff
@@ -0,0 +1,438 @@
...
+where
+	B: BlockT,
+{
+	fn validate(
+		&mut self,
+		header: &B::Header,
+		data: &[u8],
+	) -> Pin<Box<dyn Future<Output = Result<Validation, Box<dyn Error + Send>>> + Send>> {
+		if !data.is_empty() {
+			log::debug!("Received unknown data alongside the MC hash block announcement.",);
```
LGLO, line 59: Why is it important? What is the reason to prevent `data` to be non-empty?

Klapeyron: SDK allows to send additional metadata, but we won't utilize that. In that scenario, we have to ensure that noone will spam us with some garbage there, so keep the existing behavior by ensuring it is empty

partner-chains/toolkit/sidechain/sidechain-mc-hash/src/block_announce_validator.rs
```diff
@@ -0,0 +1,438 @@
...
+			Ok(mc_hash) => mc_hash,
+			Err(err) => {
+				log::debug!("Failed to retrieve MC hash from announced block digest: {err}");
+				return Box::pin(async { Ok(Validation::Failure { disconnect: true }) });
+			},
+		};
+		let block_source = self.block_source.clone();
+
+		Box::pin(async move {
+			match block_source.get_block_by_hash(mc_hash.clone()).await {
```
LGLO: This has to be check for a **stable** block in relation to a timestamp, the same that Verification IDP was using. Right?

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
```diff
@@ -97,19 +108,93 @@ impl BlockDataSourceImpl {
...
+		};
+		self.observe_latest_cardano_block_metrics(&latest);
 		let offset = self.security_parameter + self.block_stability_margin;
-		let stable = latest.number.saturating_sub(offset).into();
+		let stable = latest.block_no.saturating_sub(offset).into();
 		let block = self.get_latest_block(stable, reference_timestamp).await?;
-		Ok(block.map(From::from))
+		Ok(match block {
+			Some(block) => LatestStableBlockForTimestamp::Found(block.into()),
+			None if !self.is_local_cardano_tip_recent(&latest, reference_timestamp) => {
```
LGLO: Originally, Cardano tip age **heuristic check** was used only during block verification and when there was clue that there is a problem with it.

I think this branch is redundant and in case of block production we can use `NoStableBlockInRange`, which isn't tainted with heuristic nature. In other words: we are sure Praos conditions are not met and we don't guess that the problem is near the tip.

partner-chains/toolkit/sidechain/sidechain-mc-hash/src/block_announce_validator.rs
Klapeyron: yes it should, without it we do not make full-covered check. But it requires to provide timestamp of the block based only on its header, we can try some slot based recalculation, or maybe it is time to utilize BlockAnnounceValidator metadata to carry this information

Klapeyron: I will try with parent timestamp as well

Klapeyron: I gave a try with additional timestamp source derived from the slot, https://github.com/midnightntwrk/midnight-node/pull/1552/changes/f4d671ecca08cb34a533083df88af39912e566bc

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
```diff
@@ -12,24 +12,30 @@ use log::{debug, info, warn};
...
-	#[error("Block with hash {0} was not found.")]
-	BlockNotFound(McBlockHash),
+	#[error("Database query failed while checking block hash {hash}: {error}")]
+	Database { hash: McBlockHash, error: String },
+	#[error("Block with hash {hash} was not found.")]
+	BlockNotFound { hash: McBlockHash },
 	#[error("Latest block info was unavailable while checking block hash {0}.")]
 	LatestBlockUnavailable(McBlockHash),
+	#[error("Local Cardano tip {latest_block:?} is too old to classify the referenced hash.")]
+	LocalTipTooOld { latest_block: Block },
```
LGLO: IMO local tip should not be used for StableBlockByHashError because local tip check is heuristic and not Praos thing.

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
```diff
@@ -97,19 +108,93 @@ impl BlockDataSourceImpl {
...
+	///
+	/// Unlike [Self::is_cardano_tip_fresh], this is not a heuristic.
+	pub async fn is_cardano_ok(&self) -> Result<bool, Box<dyn std::error::Error + Send + Sync>> {
+		let timestamp = Timestamp::new(self.time_source.get_current_time_millis());
+		let timestamp = BlockDataSourceImpl::timestamp_to_db_type(timestamp)?;
+		let latest = self.get_latest_block_info().await?;
+		let stable = latest.number.saturating_sub(self.security_parameter).into();
+		let block = self.get_latest_block(stable, timestamp).await?;
+		log::debug!("is_cardano_ok timestamp: {timestamp}, stable: {stable:?}, block: {block:?}");
+		Ok(block.is_some())
```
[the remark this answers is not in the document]
LGLO: I think codex raises false positive here. This isn't any block in the window, it has to have number high enough as well.

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
```diff
@@ -97,19 +108,88 @@ impl BlockDataSourceImpl {
...
 		&self,
 		reference_timestamp: Timestamp,
-	) -> Result<Option<MainchainBlock>, Box<dyn std::error::Error + Send + Sync>> {
+	) -> Result<LatestStableBlockForTimestamp, Box<dyn std::error::Error + Send + Sync>> {
 		let reference_timestamp = BlockDataSourceImpl::timestamp_to_db_type(reference_timestamp)?;
-		let latest = self.get_latest_block_info().await?;
+		let latest = match db_model::get_latest_block_info(&self.pool).await {
+			Ok(latest) => latest,
+			Err(err) => {
+				return Ok(LatestStableBlockForTimestamp::LocalDataUnavailable {
```
LGLO: I'm not convinced to "LocalDataUnavailable" if postgres query fails. Later on, if postgres query fails during `get_latest_block` call, error will be returned.

partner-chains/toolkit/data-sources/db-sync/src/block/mod.rs
```diff
@@ -118,26 +198,42 @@ impl BlockDataSourceImpl {
...
+	/// Finds a block by its `hash` and returns its info.
+	///
+	/// Unlike [Self::get_stable_block_for], this does not check whether the block is stable or
+	/// whether its timestamp falls within the allowable window for a Partner Chain reference.
+	/// Callers that need import-time MC reference validation must use `get_stable_block_for`.
 	pub async fn get_block_by_hash(
 		&self,
 		hash: McBlockHash,
-	) -> Result<Option<MainchainBlock>, Box<dyn std::error::Error + Send + Sync>> {
+	) -> Result<BlockByHash, Box<dyn std::error::Error + Send + Sync>> {
```
LGLO, line 202: Can this function still return `Err`?

partner-chains/toolkit/sidechain/sidechain-mc-hash/src/block_announce_validator.rs
LGLO: Understood. This message could be extended or a comment added.

Klapeyron:
I will park it for now, maybe we will make another try when https://github.com/paritytech/polkadot-sdk/pull/12506 will be addressed

PR #1556, opened 2026-05-21
sec: firewall packages

gilescope:
socket firewall does cargo and JS packages.

we've also install most things in the base ci image so we should be able to use one base image everywhere which should reduce number of downloads.


gilescope:
AL2023 certs / openssl really doesn't like socket firewall's method.

Issue #1530, opened 2026-05-18
[toolkit] Toolkit consumes all available memory and crashes


chrispalaskas:
### Context & versions

On Friday the toolkit was working fine. Over the weekend I loaded the network with about 80K txs, bringing the total to 170K. Now, the toolkit cannot handle simple operations, such as `show wallet` for 100 wallets (usually works for 1000). The memory consumption maxes out without performing the task.
I experienced this once before, it's not a glitch.
Will try to add more data as I'm finding.

ozgb:
From @chrispalaskas 

> toolkit stops working at some point and I can't do any work. Eventually mainnet will get there (in utxo set size) so it will be unusable if we don't find out what is going on. I suspect we are reallocating in memory the same data.
> 
> Maybe we can migrate the functionality from my scripts to the toolkit, that can manage memory better. E.g. I use show-wallets for 1000 wallets, it goes by number of available threads every batch. Once the chain gets bloated above a threshold, even 100 wallets consume 300GB of RAM and crash. If we change --seed <SEED> to --seed-range maybe it would work beyond that point?

Migrating functionality for working on multiple wallets into the toolkit sounds like a good first-step mitigation

More comprehensive solutions:
- Add indexer support #1186 
- Support disk-backend for ledger storage (Toolkit uses default InMemory for all Ledger items at the moment)
- More generally, identify large allocations and cache to disk (using ledger storage?)

gilescope:
The toolkit replays all blocks into a `LedgerState<DefaultDB>` (in-memory storage from the ledger crate) and holds the full UTXO set in RAM. At 170K transactions the UTXO set is large enough that even a single `show-wallet` cold run exhausts available memory, and repeated invocations (e.g. 100 wallets in a shell loop) each performed full replay because the ledger snapshot was being tagged at `block_height = 0` instead of the real head - confirmed at `util/toolkit/src/tx_generator/builder/mod.rs`.

PR #1574 (merged 2026-05-28) fixed the snapshot-height tagging bug when `dust_warp=true`, so warm runs now restore correctly from the cached ~49MB snapshot rather than replaying from genesis every time. This should meaningfully reduce steady-state memory use after the first cold run.

The root cause - no disk-backed `LedgerState`, no `--seed-range` batching - is still present in the codebase. None of the three architectural mitigations proposed in the issue comments (indexer support #1186, disk-backend ledger storage, `--seed-range`) have been implemented.

[...]

Issue #1186, opened 2026-04-01
[Toolkit] Add support for fetching data from the indexer


ozgb:
## User Need:

This allows users to interact with Midnight via the node-toolkit without having to download every block locally. 

Using the toolkit is painful with long-running chains - being able to use the indexer (optimised for this use-case) will improve UX greatly. 

We'll still keep the toolkit cache system as a backup, and running in environments where the indexer isn't present.

## Expected Benefit:

The Midnight Toolkit will then be able to work as a general CLI for Midnight, and allow some testing of Indexer functionality.

## Details:

Needs investigating

PR #1605, opened 2026-05-28
refactor(toolkit): add new BuilderContext trait (indexer support prep)

ozgb:
# Overview

Encapsulate the methods toolkit transaction builders use behind a new `BuilderContext<D>` trait, so builders no longer depend on a concrete `Arc<LedgerContext<D>>` that has to replay every block locally.

The trait exposes only the queries builders actually need: `latest_block_context`, `ledger_parameters`, `network_id`, `unshielded_utxos`, `zswap_state`, `contract_state`, `resolver`/`update_resolver`, `well_formed`, and wallet access (`with_wallet_from_seed`, `with_wallets_from_seeds`). `LedgerContext` implements the trait, so runtime behaviour is unchanged. A stub `IndexerContext` proves the trait is implementable by a non-local backend.

A follow-up PR will add the real indexer-API implementation of `BuilderContext` (issue #1186), letting the toolkit drive transaction construction against a node via indexer queries instead of a fully replayed local ledger.

The `batches` command still uses the concrete `LedgerContext` backend - it needs a full ledger state to pre-generate and apply transactions to create future chain states.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist
- [x] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

Existing tests via CI should not show regression

## 🔱 Fork Strategy
- [ ] Node Runtime Update
- [ ] Node Client Update
- [x] Other: Toolkit-only refactor; no on-chain change.
- [ ] N/A

## Links

- Issue #1186 — indexer-backed toolkit
- Once #1186 is resolved, #1530 will be mitigated
- Follow-up: `IndexerContext` implementation against the indexer GraphQL API


ledger/helpers/src/versions/common/transaction.rs
```diff
@@ -78,21 +78,21 @@ impl DustRegistrationBuilder {
 	}
 }
 
-pub struct StandardTrasactionInfo<D: DB + Clone> {
-	pub context: Arc<LedgerContext<D>>,
-	pub intents: HashMap<SegmentId, Box<dyn BuildIntent<D>>>,
-	pub guaranteed_offer: Option<OfferInfo<D>>,
-	pub fallible_offers: HashMap<u16, OfferInfo<D>>,
+pub struct StandardTrasactionInfo<D: DB + Clone, C: BuilderContext<D>> {
```
Klapeyron, line 81:
```suggestion
pub struct StandardTransactionInfo<D: DB + Clone, C: BuilderContext<D>> {
```

changes/toolkit/changed/builder-context-trait.md
```diff
@@ -0,0 +1,21 @@
...
+concrete `Arc<LedgerContext<D>>`, which forced every builder to replay the whole
+chain locally. They are now generic over a new `BuilderContext<D>` trait that
+exposes only the queries builders actually need (latest block context, ledger
+parameters, network id, unshielded UTXOs, zswap state, contract state, resolver,
+well-formedness check, and wallet access). `LedgerContext` implements the trait,
+so runtime behaviour is unchanged. A stub `IndexerContext` proves the trait can
+be satisfied by a non-local backend, preparing the toolkit to talk to a node via
+indexer queries (issue #1186) without touching the builders again.
+
+No user-visible behaviour change. The `batches` builder retains its previous CLI
```
Klapeyron, line 14: will it be worth to introduce additional trait that will contain the mutable operations only? Looks like we need functions like `update_from_tx` or `block_context`

ledger/helpers/src/versions/common/context/builder_context.rs
```diff
@@ -0,0 +1,79 @@
...
+	async fn zswap_state(&self) -> ZswapChainState<D>;
+
+	/// The on-chain state of the contract at `address`, if it exists.
+	async fn contract_state(&self, address: ContractAddress) -> Option<ContractState<D>>;
+
+	/// The resolver currently used for proving.
+	async fn resolver(&self) -> &'static Resolver;
+
+	/// Replace the resolver used for proving.
+	async fn update_resolver(&self, resolver: &'static Resolver);
```
Klapeyron, line 68: some of the functions are shared among LedgerContext and BuilderContext like this one or `with_wallet_from_seed`. `latest_block_context` for instance is sometimes used via trait as async function or directly without async, can we try to unify that?

ledger/helpers/src/versions/common/context.rs
```diff
@@ -529,6 +534,95 @@ impl<D: DB + Clone> LedgerContext<D> {
...
+		self.with_ledger_state(|ledger_state| (*ledger_state.parameters).clone())
+	}
+
+	async fn network_id(&self) -> String {
+		self.with_ledger_state(|ledger_state| ledger_state.network_id.clone())
+	}
+
+	async fn unshielded_utxos(&self, seed: WalletSeed) -> Vec<(Utxo, Timestamp)> {
+		self.with_ledger_state(|ledger_state| {
+			self.with_wallet_from_seed(seed, |wallet| {
```
Klapeyron, line 572: here we are making two mutex locks (wallet and ledger_state), which may cause deadlock if somewhere else we make it in opposite direction. Maybe we can try to extract addresses first before second lock is aquired?

gilescope (approved):

changes/toolkit/changed/builder-context-trait.md
ozgb: The indexer backend is unlikely to support mutable operations - unless we expect that to change, I don't see benefit in creating the additional trait

ledger/helpers/src/versions/common/transaction.rs
ozgb: Will apply in a follow-up

ledger/helpers/src/versions/common/context/builder_context.rs
ozgb: I think this is OK and expected - some of the trait API will have a 1:1 mapping with the backend

EDIT: Can you point to the specific usage?

ledger/helpers/src/versions/common/context.rs
ozgb: Good catch, thanks - will fix this in a follow-up

ozgb: I wasn't able to neatly remove this double-locking - I've document the correct order, and verified that all current usage is the same order: https://github.com/midnightntwrk/midnight-node/pull/1626

ledger/helpers/src/versions/common/transaction.rs
ozgb: Fixed here: https://github.com/midnightntwrk/midnight-node/pull/1627

PR #1630, opened 2026-06-02
ci(local-env): Rather than use host ports for local env, always use the same ports but use Docker in Docker

gilescope:
# Overview

Runs the local-environment stack + finality check **inside earthly's nested dockerd** (`+local-env-ci`) instead of `LOCALLY` using non overlapping port ranges. Each invocation gets its own network namespace, so concurrent PRs no longer collide on host ports. I think this is a better approach as we're A) testing with more usual setup in terms of ports, and B) it's more isolated so we'd expect the test runs to be more robust.

Isolation was proven locally: two distinct `WITH DOCKER --compose` stacks both bound host `:8080` concurrently, isolated, exit 0.

Tracking issue (SRE): shieldedtech/shielded-sre#281.

## 🗹 TODO before merging

- [x] Confirm `+local-env-ci` goes green on x86 CI

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] I have included a change file, or skipped for this reason: change only affects CI
- [x] If the changes introduce a new feature, I have bumped the node minor version — N/A
- [x] Updated AGENTS.md if build commands, architecture, or workflows changed — N/A (experimental job, no workflow contract change)
- [x] No new todos introduced

## 🧪 Testing Evidence

Local (arm64): `WITH DOCKER --compose` netns isolation proven — two distinct stacks both bound host `:8080` simultaneously, each saw only its own container, both exit 0, no "port already allocated". Note: the full stack can't run natively on arm64 (cardano-db-sync is amd64-only — IntersectMBO/cardano-db-sync#2070), so `+local-env-ci` itself is validated by **this PR's x86 CI run** (the experimental job).

## 🔱 Fork Strategy

- [x] N/A — CI/build tooling only

## Links

- shieldedtech/shielded-sre#281 (SRE tracking)
- IntersectMBO/cardano-db-sync#2070 (arm64 images — blocks native arm64 local runs)


Earthfile
```diff
@@ -631,14 +620,25 @@ node-ci-image-single-platform:
         tar \
         gzip \
         docker \
-        jq && \
+        jq \
+        nodejs \
```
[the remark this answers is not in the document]
gilescope: fixed

justinfrevert (dismissed):

Earthfile
```diff
@@ -1587,8 +1598,202 @@ start-local-env-with-indexer-ci:
...
+    ARG CHAIN_INDEXER_IMAGE
+    ARG WALLET_INDEXER_IMAGE
+    ARG TOOLKIT_IMAGE
+    ARG USERARCH
+    # Fail early + kindly if any image ref is empty — otherwise `WITH DOCKER --pull` below
+    # gets an empty arg and dies with the opaque "invalid reference format".
+    RUN test -n "$NODE_IMAGE" && test -n "$TOOLKIT_IMAGE" && test -n "$INDEXER_API_IMAGE" \
+          && test -n "$CHAIN_INDEXER_IMAGE" && test -n "$WALLET_INDEXER_IMAGE" || { \
+        echo "+local-env-ci needs all five image refs, e.g.:"; \
+        echo "  earth -P +local-env-ci \\"; \
```
LGLO: Is `earth` some alias for `earthly`?

LGLO (commented):
I think QA team should see this as well

Earthfile
gilescope: Ah good spot. No it's not an alias, it's the maintained fork called EarthBuild. For now referring to it as `earthly` is the right call. I'll update that.

rsporny:
LGTM!

rsporny (approved):

scottbuckel (approved):

PR #1635, opened 2026-06-03
feat(c-to-m bridge): benchmarks for the bridge

LGLO:
# Overview

Adds benchmarks and proper weights for bridge calls. Configuration calls are benchmarked. PC Bridge handler call in the Midnight runtime environment has been benchmared as well. Weights are added to the runtime.
## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A


res/dev/reserve-config.json
```diff
@@ -4,6 +4,13 @@
     "policy_id": "d2dbff622e509dda256fedbd31ef6e9fd98ed49ad91d5c0e07f68af1",
     "asset_name": "NIGHT"
   },
-  "utxos": [],
-  "total_amount": 0
+  "_comment": "Dev/test reserve: 5B NIGHT (5e15 STARS) covers default faucet funding (~2B NIGHT). The remainder of MAX_SUPPLY lands in locked_pool so c2m-bridge UnlockToTreasury/DistributeNight(CardanoBridge) inflows have something to draw from. Without this, locked_pool=0 (because fallback for reserve is MAX_SUPPLY, that leaves 0 for locked pool) and any bridge transfer fails with SystemTransactionError::IllegalPayout.",
```
LGLO, line 7: Also present as separate change in #1634. This enables playing with bridge. Otherwise Midnight Reserve and Locked pools are empty.

runtime/src/lib.rs
```diff
@@ -905,9 +907,50 @@ impl pallet_partner_chains_bridge::Config for Runtime {
...
+				let bytes = i.to_le_bytes();
+				let mut buf = [0u8; 32];
+				buf[0..4].copy_from_slice(&bytes[0..4]);
+
+				let mc_tx_hash = sidechain_domain::McTxHash(buf);
+
+				pallet_c2m_bridge::ApprovedMcTxHashes::<Runtime>::insert(mc_tx_hash, ());
+
+				// UserTransfer is the most expensive (and common) case, so it is used for the benchmark
+				let recipient = TransferRecipient::Address {
```
justinfrevert, line 940: Do we need any validation it's a valid recipient? I'm conscious the pallet can execute txes and silently fail, which would exercise a less complex path.

justinfrevert (approved):

runtime/src/lib.rs
LGLO: `TransferRecipient::Address` can't be invalid, it is protected by type system.

But I see your point. We would rather have to add benchmark tests to see that transactions were executed, but  I think it is a bit too much, because failures aren't silent. If anyone runs benchmarks and transactions fail, they are spamming the screen. I know, because they were failing before adding ledger initialization.

Issue #1648, opened 2026-06-05
Warp sync doesn't work


gilescope:
It seems current midnight consensus is not supporting warp sync.
(I think this should work regardless of if we're using Babe or Aura for consensus - It normally works with Babe but I don't know of any reason it shouldn't also work for Aura)

gilescope:
Warp sync blocks arrive with `StateAction::ApplyChanges(State)` (`with_state() == true`), which the verifier must short-circuit without running inherent checks. The old custom Aura fork (`sc-partner-chains-consensus-aura`) did have a `with_state()` early return, but the entire fork was replaced by #1700 (merged 2026-07-17).

The new `PartnerChainsVerifier` in `partner-chains/toolkit/consensus/src/verifier.rs` delegates to the upstream `sc_consensus_aura` verifier for warp sync blocks (`block.with_state() || block.state_action.skip_execution_checks()`) and has a dedicated canary test (`skip_execution_checks_bypasses_partner_chains_check_and_delegates_to_aura`) that pins this exact path - along with a comment noting "if either side starts requiring a full inherent check here, warp sync for new nodes wedges."

#1700 references issue #1859 rather than this one, so the fix may have been a side effect of the larger refactor rather than a direct fix.

Can @justinfrevert or someone on the node team re-test warp sync on the latest main (post-#1700) and confirm whether this is now resolved?

---
*Claude triage review - an automated pass over open bugs with little recent activity. Conclusions come from reading this issue plus the current code and git history, and may be wrong; please correct anything that is off.*

justinfrevert:
I tested it. We still need the PR, as this doesn't sync the ledger database. 

Claude sums up the run:
> 2. Ledger separate-storage sync — the real work. Warp would need to also fetch/reconstruct the Midnight ledger's separate DB for the warp target (extend the warp/state-sync path to cover midnight-storage-core, or make the ledger lazily backfill its arena from peers). That's a substantial design change in the storage/ledger layer, not a config tweak.

We have a PR to implement this here: https://github.com/midnightntwrk/midnight-node/pull/1650, where we've tested warp sync successfully.

PR #1650, opened 2026-06-05
Add ledger arena for warp sync

justinfrevert:
# Overview

Part 1 / ~3 for https://github.com/midnightntwrk/midnight-node/issues/1648. Prepares ledger arena storage for warp sync. Adds ledger sync process and handlers for ledger data transfer. (This is covered under the upcoming Q3 statement of work regarding sync performance). 

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A


ledger/src/versions/common/mod.rs
```diff
@@ -729,6 +729,82 @@ where
 		api.serialize(&ledger_state.as_typed_key())
 	}
 
+	/// Serialize the full ledger arena snapshot at `state_key` into the canonical, `Ledger`-rooted
+	/// transfer blob used by trustless warp ledger-sync: `derived_tag_prefix ‖
+	/// TopoSortedNodes(Ledger DAG)`.
+	///
+	/// Mirrors the single-pass technique of the toolkit's `serialize_ledger_state_fast`, but roots
+	/// at `Ledger` (the `Sp` from `get_ledger` is an `Sp<Ledger>`) rather than `LedgerState` — see
+	/// `warp-ledger-sync-spec.md` ODD-1. Because the blob is rooted at `Ledger`, its recomputed
```
gilescope: warp-ledger-sync-spec.md - stray comment?

ledger/src/versions/common/mod.rs
```diff
@@ -729,6 +729,82 @@ where
...
+		// node DAG), written directly. Byte-identical to the default impl's output.
+		let nodes: TopoSortedNodes = ledger.serialize_to_node_list();
+		let tag_prefix = format!("{}{}:", GLOBAL_TAG, <Ledger<D> as Tagged>::tag());
+		let mut bytes = Vec::with_capacity(tag_prefix.len() + nodes.serialized_size());
+		bytes.extend_from_slice(tag_prefix.as_bytes());
+		nodes.serialize(&mut bytes).map_err(|e| {
+			log::error!(target: LOG_TARGET, "Failed to serialize ledger snapshot: {e:?}");
+			LedgerApiError::Serialization(SerializationError::LedgerState)
+		})?;
+		Ok(bytes)
```
gilescope, line 810: How big is this snapshot? It might be worth a fast compress to make it half the size?

gilescope (dismissed):

ledger/src/versions/common/mod.rs
justinfrevert: I think with the content being mostly cryptographic, we're not likely to get many compression gains. Let's take it as an issue and look into it if it gets to the point where we really need a solution.

justinfrevert: https://github.com/midnightntwrk/midnight-node/issues/1687

node/src/warp_ledger_sync/oracle.rs
```diff
@@ -0,0 +1,155 @@
...
+
+impl RecoveryGate {
+	/// A fresh gate: nothing pending (full-sync default).
+	pub fn new() -> Arc<Self> {
+		Arc::new(Self::default())
+	}
+
+	/// Arm the gate: the warp path was taken, so authoring + import must wait for verification.
+	pub fn arm(&self) {
+		self.recovery_pending.store(true, Ordering::Release);
```
Klapeyron, line 53: maybe it will simplify the code, if we will be persisting in the database this flag, so we may have support for restart of the node & the same location of data (fetched data & flag), also AuxData::insert_aux supports transactional operations if we put multiple keys together

node/src/warp_ledger_sync/server.rs
```diff
@@ -0,0 +1,170 @@
...
+		let blob = midnight_node_ledger::serialize_ledger_snapshot(self.unified, &state_key)
+			.map_err(HandleError::Serialize)?;
+		log::debug!(
+			target: LOG_TARGET,
+			"Serialized ledger snapshot for {target:?}: {} bytes",
+			blob.len()
+		);
+
+		let blob = Arc::new(blob);
+		self.cache = Some((target, blob.clone()));
```
Klapeyron: what is the average blob size there? Are we safe to store that in-memory?

node/src/warp_ledger_sync/server.rs
```diff
@@ -0,0 +1,170 @@
...
+				result,
+				reputation_changes: Vec::new(),
+				sent_feedback: None,
+			});
+		}
+	}
+
+	fn handle_request(&mut self, payload: &[u8]) -> Result<Vec<u8>, HandleError> {
+		let req = LedgerSyncRequest::<B::Hash>::decode(&mut &payload[..])?;
+		let blob = self.blob_for(req.target_hash)?;
```
Klapeyron: is it expensive to compute the blob for specific hash? Maybe worth to narrow scope of available hashes, like `target_hash` will be available only for blocks that rotate the session, or even only for the last session bundary that is finalized?

Klapeyron: GRANDPA should already follow that path of session-boundary hops, but maybe to not make it dependent on that, there is a chance of running GRANDPA warp-sync first, or retrieve the session boundary block via runtime api like `Session::current_index`

Klapeyron: having that also, maybe we can adjust the protocol to not target a block hash but rather a session index?

node/src/warp_ledger_sync/oracle.rs
```diff
@@ -0,0 +1,155 @@
...
+	/// Wait until recovery is no longer in progress (poll-based; recovery takes seconds-to-minutes,
+	/// so a sub-second poll adds no meaningful latency and keeps the gate free of async machinery).
+	pub async fn wait_until_released(&self) {
+		while self.ledger_recovery_in_progress() {
+			tokio::time::sleep(std::time::Duration::from_millis(250)).await;
+		}
+	}
+}
+
+/// Wraps the node's inner [`SyncOracle`] (the `SyncingService`) so AURA reports "still syncing"
```
Klapeyron, line 77: AURA has internal logic to skip block authorship when it is in major_sync, it is located in the Verifier that is used for block production & block import:
https://github.com/paritytech/polkadot-sdk/blob/8ae2e6a47aef16e392b4a951b7165c87d9f1e75b/substrate/client/consensus/aura/src/import_queue.rs#L157
maybe we can check if that works when warp is active, or set the flag if that is not a case?

ledger/src/versions/common/mod.rs
justinfrevert: I change my mind after seeing a compressed full node snapshot. Huge size reduction. Added a pass at compression/decompression in [524ac65](https://github.com/midnightntwrk/midnight-node/pull/1650/commits/524ac650251585c8522956fbf2a63bdc8afe6c10).

node/src/warp_ledger_sync/oracle.rs
justinfrevert: node/src/warp_ledger_sync/monitor.rs checks the relevant state key on boot, so we can resume partial states.

node/src/warp_ledger_sync/server.rs
justinfrevert: It's a good idea for protection. I think with the scope, maybe we can work on this separately. https://github.com/midnightntwrk/midnight-node/issues/1723.

node/src/warp_ledger_sync/oracle.rs
justinfrevert: We do use the internal aura check ahead of our check on warp sync status
```rust
inner.is_major_syncing() || gate.ledger_recovery_in_progress()
```
Does that make sense for now?

node/src/warp_ledger_sync/server.rs
justinfrevert: I just pulled it for mainnet, and it's a few hundred MiB.

node/src/service.rs
```diff
@@ -552,6 +577,33 @@ pub async fn new_full<Network: sc_network::NetworkBackend<Block, <Block as Block
...
 	net_config.add_request_response_protocol(beefy_req_resp_cfg);
 
+	// Warp ledger-sync: register the request/response protocol that serves / recovers the
+	// Midnight ledger arena after warp+state-sync. The server handler is spawned after the network
+	// is built; `ledger_sync_protocol_name` is reused by the client driver in the monitor.
+	//
+	// Validators do not serve: arena serialization is the protocol's most CPU-expensive operation
+	// and must not compete with authoring/finality (a remote DoS vector). They still register the
+	// protocol as a client, so a fresh validator can warp-sync and recover its own arena.
+	let serve_ledger_sync = !config.role.is_authority();
```
Klapeyron: maybe worth to add a dedicated command line flag that will activate that? Thinking about smaller or local networks with validators only

node/src/warp_ledger_sync/client.rs
```diff
@@ -0,0 +1,199 @@
...
+			{
+				return Err(ClientError::InconsistentResponse);
+			}
+			if next.bytes.is_empty() {
+				// Server returned an empty range before completion: treat as a truncated transfer.
+				// `into_blob` below will surface `Incomplete`.
+				break;
+			}
+			assembler.accept(next.offset, &next.bytes)?;
+		}
```
Klapeyron, line 191: do we need to introduce some protection against peers sensing legit but really small chunks of data?

node/src/backend/custom_parity_db.rs
```diff
@@ -107,20 +107,67 @@ pub fn open<H: Clone + AsRef<[u8]>>(
...
 			Ok((OwnedDb(db), LedgerStorageDb::SeparateDb(storage_config.db_path.clone())))
 		},
 		StorageSeparation::Unified => {
-			midnight_node_ledger::ledger_9::storage::init_storage_paritydb_unified::<
-				_,
-				NUM_COLUMNS_POLKADOT,
-			>(OwnedDb(db.clone()), &storage_config.genesis_state, storage_config.cache_size);
+			let cache = storage_config.cache_size;
+			match genesis_version {
+				Some(16 | 17 | 18) => {
```
Klapeyron: is there maybe some option to get that information from ledger directly or the node? I mean, it may be harder to maintain in longer term if we will have to maintain the list

node/src/warp_ledger_sync/server.rs
Klapeyron: delivering of this PR automatically activates this feature (warp syncing of ledger as general) or this will be in later PR's?

node/src/warp_ledger_sync/server.rs
```diff
@@ -0,0 +1,170 @@
...
+		let header = self.client.header(target)?.ok_or(HandleError::UnknownBlock)?;
+		if *header.number() > self.client.info().finalized_number {
+			return Err(HandleError::NotFinalized);
+		}
+
+		// Read the raw `pallet_midnight::StateKey` at the target block.
+		let state_key = read_state_key::<B, Client, BE>(&self.client, target)?
+			.ok_or(HandleError::NoStateKey)?;
+
+		let blob = midnight_node_ledger::serialize_ledger_snapshot(self.unified, &state_key)
```
Klapeyron, line 168: I am thinking if this operation by blocking the thread is blocking other parts of the system or the request is already spawned as separate task/thread?

node/src/service.rs
```diff
@@ -327,9 +330,16 @@ pub fn new_partial(
...
+	// `no_genesis()` is true, so we skip the commit — leaving `finalized_state == None` so substrate
+	// will actually engage warp sync (it refuses warp on a DB that already has a finalized state).
+	// This mirrors stock substrate (`sc_service::builder`'s `!config.no_genesis()`); hardcoding
+	// `true` here previously meant warp sync silently fell back to full sync. Full-sync behavior is
+	// unchanged (`no_genesis()` is false → commit). The ledger arena genesis init in `open_paritydb`
+	// is independent and still runs, so `default_storage` is set for post-warp recovery.
 	let genesis_block_builder = GenesisBlockBuilder::<Block, _, _>::new(
 		genesis_storage,
-		true,
+		!config.no_genesis(),
```
Klapeyron: this function also says about LightState, does it have impact for us? It is probably also accessible from command line
```rust
	pub fn no_genesis(&self) -> bool {
		matches!(self.network.sync_mode, SyncMode::LightState { .. } | SyncMode::Warp { .. })
	}
```

node/src/service.rs
justinfrevert: Added [86cdcdc](https://github.com/midnightntwrk/midnight-node/pull/1650/commits/86cdcdcb16a709ee85257804c33e5cc563f83027)

node/src/backend/custom_parity_db.rs
justinfrevert: I think that it will take some more work. With warp sync, we are not executing within block, block-by-block, so we do not get the ledger version automatically. https://github.com/midnightntwrk/midnight-node/issues/1816

node/src/warp_ledger_sync/server.rs
justinfrevert: > delivering of this PR automatically activates this feature (warp syncing of ledger as general) or this will be in later PR's?

Non-authority nodes will automatically serve the ledger arena data. Syncing nodes can opt-in to warp-sync ledger with the flag.

node/src/warp_ledger_sync/client.rs
justinfrevert: Added [0f539aa](https://github.com/midnightntwrk/midnight-node/pull/1650/commits/0f539aac5954ffb9e37e9d46772c1c447490348e)

node/src/warp_ledger_sync/server.rs
justinfrevert: That's true. I just changed the new server task to use `spawn_blocking`.

node/src/service.rs
justinfrevert: True. Changed it to just depend on warp sync.

chrispalaskas:
**Field report from perfnet (`d261900b` deployed to 9 nodes: 8 validators + 1 rpc, c7i.xlarge / 4 vCPU / 8 GiB, AL2023):**

The fix works as advertised — we wiped the warp client's `paritydb` + `ledger_storage` and the full pipeline completed end-to-end with the runtime responsive throughout (peers held at 8, block announcements pre-validated during recovery, `NRestarts=0`):

```
00:27:40 Warp sync is complete, continuing with state sync.
00:27:41 State sync is complete, continuing with block sync.
00:27:42 Recovering ledger arena at warp target #20273 (0x8b09…5648)
00:28:02 Verifying + importing ledger arena snapshot … (41421163 bytes); large arenas can take a while…
01:38:58 Recovered + verified ledger arena … (41421163 bytes) in 4256.2s
01:38:58 Ledger arena recovered + verified; authoring + import gate released
```

Two observations that may be worth follow-up:

1. **Verify+import took 4,256 s (71 min) for a 41.4 MB snapshot** (~10 KB/s effective), and the phase appears wait-bound, not compute-bound: sampled ~12 min in, all 39 threads were at 0.0 % CPU with ~3 s *total* accumulated CPU since process start; 37 threads parked in `futex_do_wait`, 2 in `ep_poll`. Whatever the verify path serializes on (lock/channel round-trips per storage node?), it isn't using the machine. On a 20k-block chain this held the authoring gate closed for 71 min — on a validator, that's a missed slot every committee rotation for the duration; at mainnet arena sizes this would extrapolate to much worse.

2. **The 15 s heartbeat (`…still verifying (Ns elapsed)`) is decoupled from progress** — it ticks from a timer regardless of whether the verify task advances. During the stall it was indistinguishable from a hang until we looked at thread states. Logging actual progress (bytes/nodes imported) would make slow-vs-stuck diagnosable from logs alone.

Happy to provide the full journal or re-run with extra logging — the node (perfnet euw1-v8) is wiped/resynced on demand.

gilescope (approved):
Have got some hardnening suggestions. Have put them in a separate PR.

PR #1651, opened 2026-06-08
Fix ledger 8 host function interface

justinfrevert:
# Overview

Repairs the ledger 8 host function interface, as it was altered. Ledger 8 is already on networks, so we cannot change the host functions.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A

### DIFF
--- a/ledger/src/host_api/ledger_8.rs
+++ b/ledger/src/host_api/ledger_8.rs
@@ -429,20 +429,19 @@ pub trait Ledger8Bridge {
 		}
 	}
 
-	// Restored for backwards-compatibility (removed by #1604).
-	// The deployed ledger_8 runtime imports this symbol
-	// (`ext_ledger_8_bridge_construct_distribute_treasury_system_tx_version_1`); removing it from this
-	// frozen interface meant our ledger_9 binary could not instantiate the ledger_8 runtime, blocking
-	// the ledger_8 -> ledger_9 upgrade. Restored verbatim to regenerate the exact same symbol.
+	// Restored for backwards-compatibility (removed by #1604). The deployed ledger_8 runtime imports
+	// this symbol (`ext_ledger_8_bridge_construct_distribute_treasury_system_tx_version_1`); dropping it
+	// from this frozen interface meant our ledger_9 binary could not instantiate the ledger_8 runtime,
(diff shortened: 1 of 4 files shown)


ledger/src/host_api/ledger_8.rs
```diff
@@ -429,6 +429,21 @@ pub trait Ledger8Bridge {
 		}
 	}
 
+	// Restored for backwards-compatibility (removed by #1604). The deployed ledger_8 runtime imports
+	// this symbol (`ext_ledger_8_bridge_construct_distribute_treasury_system_tx_version_1`); dropping it
+	// from this frozen interface meant our ledger_9 binary could not instantiate the ledger_8 runtime,
```
LGLO, line 434: Is it possible to use some more direct test for keeping this requirement? Like having the test trying to instantiate ledger_8 runtime?

gilescope: We have a sync test for testing ledger 8 for example but due to lack of state migration that test has been disabled for example. So once we can upgrade then yes.

gilescope (changes requested):
Can we check with Thomas on this one? Host functions can be added and if we re-release a ledger 8 + 9 that understand a new host function then that would be reasonable. Let's check to see if the correct fix should be to upgrade our ledger 8 version so that it understands this new host function....

justinfrevert:
> Can we check with Thomas on this one? Host functions can be added and if we re-release a ledger 8 + 9 that understand a new host function then that would be reasonable. Let's check to see if the correct fix should be to upgrade our ledger 8 version so that it understands this new host function....

I opened this because we removed one 

> Can we check with Thomas on this one? Host functions can be added and if we re-release a ledger 8 + 9 that understand a new host function then that would be reasonable. Let's check to see if the correct fix should be to upgrade our ledger 8 version so that it understands this new host function....

I think technically, we would still need to do this work, but since the next version won't be used beyond lower environments, it'd just be simpler to do this along with the ledger 9 version + migration we want to put on all envs.

Issue #1657, opened 2026-06-08
ci(nightly-cnight-e2e-qanet): pre-flight check the Cardano faucet balance


rsporny:
## Summary

The nightly `nightly-run-cnight-e2e-qanet` workflow has a fast pre-flight smoke check for the toolkit-cache DB (`psql … -c 'SELECT 1;'`), but no equivalent for the Cardano Preview faucet that primes the e2e worker pool. A drained faucet currently fails inside the cargo test, *after* an 8-minute release build, with:

```
FaucetManager: faucet has 14474 ADA total but needs at least 16005 ADA
to prime 16 workers of 1000 ADA each.
Top up addr_test1vr5vxqpnpl3325cu4zw55tnapjqzzx78pdrnk8k5j7wl72c6y08nd.
```

(See the run that surfaced this: `cnight::observation::*` × 11 panicked in `tests/e2e/src/faucet.rs:238`.)

The panic itself is good — it bails in seconds rather than draining the 3 h stability window — but it costs the wasted build. A balance check at the *start* of the job would shrink that failure from ~8 min to ~3 s and surface the actionable Slack ping ("top up the faucet") far earlier.

## Proposed check

The threshold the FaucetManager enforces is derived from two constants in `tests/e2e/src/faucet.rs`:

```rust
const MIN_WORKER_LOVELACE: u64 = REQUEST_CAP_LOVELACE; // 1000 ADA
const PRIME_FEE_BUFFER:    u64 = 5_000_000;            // 5 ADA
// required = worker_count() * MIN_WORKER_LOVELACE + PRIME_FEE_BUFFER
```

A smoke check should:

1. Resolve the faucet address (same source the e2e config uses today — see `tests/e2e/src/config.rs` and `tests/e2e/src/faucet.rs`).
2. Resolve `worker_count()` from `E2E_FAUCET_WORKERS` (same env-var override the workflow already sets to `16`).
3. Query Ogmios for UTxOs at the address (`tests/e2e/src/api/cardano.rs` already wraps Ogmios via `OgmiosClientSettings`).
4. If `total_lovelace < worker_count * MIN_WORKER_LOVELACE + PRIME_FEE_BUFFER`, fail with the same human-readable message plus a link to https://docs.cardano.org/cardano-testnets/tools/faucet/.

[...]

PR #1578, opened 2026-05-27
ci(nightly-e2e-qanet): integrate shared toolkit fetch cache

skylar-simoncelli:
# Overview

Wire the nightly qanet cNgD e2e workflow into the shared toolkit fetch cache (Postgres RDS in shielded-dev, fronted by an NLB whose security group only allows the Hetzner runner's static IP). Cold-cache sync on qanet costs ~3h per run today; the shared cache eliminates that.

Closes [shieldedtech/shielded-gitops#2152](https://github.com/shieldedtech/shielded-gitops/issues/2152).

## What changes

`.github/workflows/nightly-run-cnight-e2e-qanet.yml` only:

- `runs-on`: `ubuntu-latest` → `[self-hosted, "tier:large"]` (Hetzner fleet). Required because the toolkit-cache NLB is locked to specific static IPs — GH-hosted runners have unallowlistable egress.
- Install `postgresql-client` alongside `protoc`.
- New pre-flight smoke step: `psql -c 'SELECT 1;'` against `TOOLKIT_CACHE_DB_URL`. Fails fast (~1s) if the DB or NLB is broken, before the ~4h cargo test job starts.
- Pass `TOOLKIT_CACHE_DB_URL` to the cargo test step via `env`.

## Verified before opening

End-to-end SQL round-trip from `fsn1-runner-01` through the NLB to the RDS:

- Runner egress IP `178.63.134.209` → NLB SG allow.
- NLB DNS resolves to the three EIPs (`34.253.0.153`, `54.246.98.89`, `52.208.163.70`).
- TCP probe to NLB:5432 succeeds.
- `psql … -c 'SELECT 1;'` returns `1`, DB `toolkit_cache_qanet`, engine PostgreSQL 16.14.

## Prereqs (already done)

- [x] `TOOLKIT_CACHE_DB_URL` repo secret set (rotate via: `pw=$(AWS_PROFILE=shielded-dev aws secretsmanager get-secret-value --secret-id toolkit-cache/master --region eu-west-1 --query SecretString --output text | jq -r .password); printf '%s' "postgresql://toolkit_cache_admin:${pw}@toolkit-cache-nlb-a6d7b387f8880acb.elb.eu-west-1.amazonaws.com:5432/toolkit_cache_qanet?sslmode=require" | gh secret set TOOLKIT_CACHE_DB_URL --repo midnightntwrk/midnight-node`).

## Out of scope

[...]

PR #1658, opened 2026-06-08
ci(nightly-cnight-e2e-qanet): run job in container to drop sudo apt-get

### LINKED ISSUE #1655: ci(nightly-cnight-e2e-qanet): `sudo apt-get` fails on self-hosted runner
## Summary

The nightly `nightly runs` workflow (`.github/workflows/nightly-run-cnight-e2e-qanet.yml`) fails on its first step inside the "Install system dependencies (protoc, postgresql-client)" stage:

```
Run sudo apt-get update
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
```

The job was moved from `ubuntu-latest` to `[self-hosted, "tier:large"]` (Hetzner) in #1571 / d1691aa4 so it could reach the toolkit-cache NLB via the runner's static IP. The Hetzner runner user is not a passwordless sudoer and has no tty, so `sudo apt-get …` can't prompt for a password and the step exits non-zero before any tests run.

The bug was latent until the first scheduled run after #1613 (41578181), which is the first commit that ships the qanet-targeted observation suite the workflow tries to exercise — hence "broken by 41578181" in practice even though the offending step landed earlier.

## Impact

- Nightly cNIGHT observation tests against Cardano Preview have not run since this regression landed.
- Slack notifications are still firing (`if: always()`), so the failure is visible — but no actual test signal is produced.

## Root cause

`sudo` on the self-hosted runner requires either:
- a tty (we have none), or
- `NOPASSWD:` configuration on the runner host (we don't have it, and adding it to a CI host is a security/admin call, not a workflow change).

GitHub-hosted runners (e.g. `ubuntu-latest`) configure passwordless sudo by default; self-hosted runners do not.

## Proposed fix

Run the job inside a `container:` so apt installs happen as root, no sudo needed. Image should pin Rust 1.95 to match `rust-toolchain.toml` (e.g. `rust:1.95-trixie`, same base as the Earthfile's `build-prepare` stage).

Sketch:

```yaml
jobs:
  rust_e2e:
    runs-on: [self-hosted, "tier:large"]
    container:
      image: rust:1.95-trixie
    timeout-minutes: 360
    steps:
      - uses: actions/checkout@... { with: { submodules: true } }
      - name: Install system dependencies
        run: |
          apt-get update
          apt-get install -y --no-install-recommends \
            protobuf-compiler postgresql-client \
            clang pkg-config libssl-dev libpq-dev libsqlite3-dev
      # ...remaining steps unchanged
```

Notes for the implementer:
- The container shares the host's network via Docker's default bridge, so outbound connections still SNAT to the runner's static IP — the toolkit-cache NLB allowlist keeps working.
- The `dtolnay/rust-toolchain@stable` step becomes redundant (the image ships rustup + Rust 1.95; the workspace's `rust-toolchain.toml` pulls any missing components/targets on first cargo invocation).
- `ubuntu-latest` had `clang`, `libssl-dev`, `pkg-config`, etc. by default — `rust:1.95-trixie` does not — so the apt list needs to grow vs. the current workflow.

## Follow-up (out of scope, but worth tracking)

Bake a CI image (e.g. `ghcr.io/midnightntwrk/ci-rust-e2e:1.95`) with the apt deps + rustup components pre-installed, so each nightly run skips the apt-get + first-cargo-invocation toolchain setup cost.

Assisted-by: Claude:claude-4.7-opus

rsporny:
# Overview

The nightly `nightly-run-cnight-e2e-qanet` workflow's `sudo apt-get update` step failed on its first run after moving to the Hetzner self-hosted runner (#1578):

```
sudo: a terminal is required to read the password; either use the
-S option to read from standard input or configure an askpass helper
```

The runner user is not a passwordless sudoer and has no tty, so `sudo apt-get …` cannot prompt for a password. `ubuntu-latest` configures passwordless sudo by default; self-hosted runners do not.

## Approach

Move the job into a `rust:1.95-trixie` container — same Rust version as `rust-toolchain.toml` and the Earthfile's `build-prepare` stage — so apt installs happen as root with no sudo needed.

- The container shares the host's network via Docker's default bridge, so outbound connections still SNAT to the runner's static IP and the toolkit-cache NLB allowlist keeps working.
- The `dtolnay/rust-toolchain@stable` step is dropped: the image already ships rustup + Rust 1.95, and the workspace's `rust-toolchain.toml` pulls any missing components on first cargo invocation.
- Apt install is now `protobuf-compiler`, `postgresql-client`, and `pkg-config` only. `libssl-dev`, `libpq-dev`, and `libsqlite3-dev` are pre-installed in the image. `clang` is intentionally omitted: the e2e test binary depends on `subxt + sqlx + redb` only, none of which pull `rocksdb` or any other `bindgen`-using crate (the Earthfile installs `clang` because it builds the full node + runtime wasm, a heavier dep set than this job).
- Apt caches are cleared after install so the ~4 h cargo build has disk headroom on the runner.

Closes #1655.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

Two trial runs against the self-hosted runner confirmed the fix:

1. **First run** failed during `apt-get install` with `E: You don't have enough free space in /var/cache/apt/archives/`. Investigation showed the failure was a transient host-state issue, but it also revealed that several packages I'd initially listed (`clang`, `libssl-dev`, `libpq-dev`, `libsqlite3-dev`) were either pre-installed in the image or not needed by this crate's dep tree. The squashed commit reflects the trimmed final list.
2. **Second run** passed the workflow steps this PR touches: container started, apt install completed, cargo build finished in 8m 27s, and all 11 cNIGHT observation tests started. The subsequent failure was a separate, unrelated runtime issue — the Cardano Preview faucet at `addr_test1vr5vxqpnpl…q3h90l6` is underfunded by ~1.5k tADA — tracked in #1657 with a proposed pre-flight smoke check to fail fast on this class of failure in future.

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other
- [x] N/A

## Links

- Closes #1655 — `sudo apt-get` failure on self-hosted runner
- Related: #1657 — proposed faucet-balance smoke check (surfaced by the second trial run)
- Related: #1578 — original move to the Hetzner self-hosted runner

Assisted-by: Claude:claude-4.7-opus

### DIFF
--- a/.github/workflows/nightly-run-cnight-e2e-qanet.yml
+++ b/.github/workflows/nightly-run-cnight-e2e-qanet.yml
@@ -100,6 +100,19 @@ jobs:
         run: |
           cargo test --release --test e2e_tests --no-default-features --features "$E2E_FEATURE" -- cnight::observation:: --no-capture --test-threads 16
 
+      # `container:` runs as root, so the bind-mounted workspace (and
+      # target/) end up UID 0. Host-mode workflows on this self-hosted
+      # runner pool (see continuous-integration.yml, main.yml) then can't
+      # rm/replace these files during their own checkout. Chown back to
+      # whoever owns the workspace's parent directory — that's the runner
+      # user, by definition. `if: always()` so cancelled/failed jobs
+      # don't poison the runner either.
+      - name: Restore workspace ownership for next runner job
+        if: always()
+        shell: bash
+        run: |
+          chown -R --reference="$(dirname "$GITHUB_WORKSPACE")" "$GITHUB_WORKSPACE"
+
       - name: Notify Slack
         if: always()
         env:
--- a/.github/workflows/nightly-run-cnight-e2e-qanet.yml
+++ b/.github/workflows/nightly-run-cnight-e2e-qanet.yml
@@ -23,9 +23,10 @@ jobs:
     # Run inside a container so test deps are apt-installed as root.
     # The self-hosted runner user is not a passwordless sudoer and has
     # no tty, so `sudo apt-get` fails with "a terminal is required to
-    # read the password". The image pins Rust 1.95 to match
-    # rust-toolchain.toml (single source of truth, mirrored by the
-    # Earthfile's build-prepare stage).
+    # read the password". The Rust version below is duplicated from
+    # rust-toolchain.toml and must be bumped by hand on toolchain
+    # upgrades — #1661 tracks replacing this with a baked CI image
+    # that closes the drift.
     container:
       image: rust:1.95-trixie
     timeout-minutes: 360
@@ -52,8 +53,10 @@ jobs:
       # any other bindgen-using crate (subxt + sqlx + redb only). The
       # Earthfile installs clang because it builds the full node +
       # runtime wasm, a much heavier dep set than we need here.
-      # Clear apt caches afterwards so the ~4 h cargo build has disk
-      # headroom on this runner.
+      # Clear apt caches afterwards so the cargo test build has disk
+      # headroom on this runner — target/ grows several GB during the
+      # release compile (~8 min) that precedes the multi-hour Cardano
+      # stability waits.
       - name: Install system dependencies
         shell: bash
         run: |
--- a/.github/workflows/nightly-run-cnight-e2e-qanet.yml
+++ b/.github/workflows/nightly-run-cnight-e2e-qanet.yml
@@ -20,6 +20,14 @@ jobs:
     # Hetzner self-hosted; its static IP is on the toolkit-cache NLB
     # allowlist. ubuntu-latest cannot reach the shared cache.
     runs-on: [self-hosted, "tier:large"]
+    # Run inside a container so test deps are apt-installed as root.
+    # The self-hosted runner user is not a passwordless sudoer and has
+    # no tty, so `sudo apt-get` fails with "a terminal is required to
+    # read the password". The image pins Rust 1.95 to match
+    # rust-toolchain.toml (single source of truth, mirrored by the
+    # Earthfile's build-prepare stage).
+    container:
+      image: rust:1.95-trixie
     timeout-minutes: 360
 
     steps:
@@ -34,16 +42,28 @@ jobs:
         with:
           submodules: true
 
-      - name: Install Rust toolchain
-        uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 #v1
-        with:
-          toolchain: stable
-
-      - name: Install system dependencies (protoc, postgresql-client)
+      # Running as root inside the container, so no sudo.
+      # `rust:1.95-trixie` already ships gcc/make/git/libssl-dev/libpq-dev/
+      # libsqlite3-dev, so we only add what's missing for this crate:
+      #   - protobuf-compiler: prost/scale-codec build scripts.
+      #   - postgresql-client: psql smoke test against toolkit-cache.
+      #   - pkg-config:        for -sys crates to discover system libs.
+      # No clang/libclang — the e2e test binary doesn't pull rocksdb or
+      # any other bindgen-using crate (subxt + sqlx + redb only). The
+      # Earthfile installs clang because it builds the full node +
+      # runtime wasm, a much heavier dep set than we need here.
+      # Clear apt caches afterwards so the ~4 h cargo build has disk
+      # headroom on this runner.
+      - name: Install system dependencies
         shell: bash
         run: |
-          sudo apt-get update
-          sudo apt-get install -y protobuf-compiler postgresql-client
+          set -euo pipefail
+          apt-get update
+          apt-get install -y --no-install-recommends \
+            protobuf-compiler \
+            postgresql-client \
+            pkg-config
+          rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*.deb
           protoc --version
 
       # Fail fast (~1 s) if the shared toolkit cache is unreachable,
--- a/changes/node/changed/nightly-cnight-e2e-self-hosted-container.md
+++ b/changes/node/changed/nightly-cnight-e2e-self-hosted-container.md
@@ -30,8 +30,9 @@ the e2e test binary depends on `subxt` + `sqlx` + `redb` only, none of
 which pull `rocksdb` or any other `bindgen`-using crate, so libclang
 is not needed (the Earthfile installs it because it builds the full
 node + runtime wasm, a heavier dep set than this job). Apt caches are
-cleared after install so the ~4 h cargo build has disk headroom on
-the runner.
+cleared after install so the cargo test build has disk headroom on
+the runner — `target/` grows several GB during the release compile
+(~8 min) that precedes the multi-hour Cardano stability waits.
 
 A final `if: always()` step chowns the workspace back to the runner
 user before the job exits. Without it, host-mode workflows landing on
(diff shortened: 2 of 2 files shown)


.github/workflows/nightly-run-cnight-e2e-qanet.yml
```diff
@@ -20,6 +20,14 @@ jobs:
...
     # allowlist. ubuntu-latest cannot reach the shared cache.
     runs-on: [self-hosted, "tier:large"]
+    # Run inside a container so test deps are apt-installed as root.
+    # The self-hosted runner user is not a passwordless sudoer and has
+    # no tty, so `sudo apt-get` fails with "a terminal is required to
+    # read the password". The image pins Rust 1.95 to match
+    # rust-toolchain.toml (single source of truth, mirrored by the
+    # Earthfile's build-prepare stage).
+    container:
+      image: rust:1.95-trixie
```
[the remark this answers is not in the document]
rsporny: Good catch — real bug, not a false positive. Fixed in 02a501f2.

Added an `if: always()` final step that chowns `$GITHUB_WORKSPACE` back to the owner of its parent directory (i.e. the runner user). Using `chown -R --reference="$(dirname "$GITHUB_WORKSPACE")" …` instead of a hard-coded UID/GID so it stays portable if the runner-pool user changes, and `always()` so cancelled or failed runs also clean up after themselves rather than poisoning the runner.

Verified the diff lands the step before the Slack notify so the notification still fires either way.

.github/workflows/nightly-run-cnight-e2e-qanet.yml
```diff
@@ -20,6 +20,14 @@ jobs:
     # Hetzner self-hosted; its static IP is on the toolkit-cache NLB
     # allowlist. ubuntu-latest cannot reach the shared cache.
     runs-on: [self-hosted, "tier:large"]
+    # Run inside a container so test deps are apt-installed as root.
+    # The self-hosted runner user is not a passwordless sudoer and has
+    # no tty, so `sudo apt-get` fails with "a terminal is required to
+    # read the password". The image pins Rust 1.95 to match
+    # rust-toolchain.toml (single source of truth, mirrored by the
```
LGLO: So, it isn't a single source of truth. rust-toolchain.toml is a source of truth, line 30 in this file is another source of truth and Earthfile is another one.

.github/workflows/nightly-run-cnight-e2e-qanet.yml
```diff
@@ -34,16 +42,28 @@ jobs:
...
+      # `rust:1.95-trixie` already ships gcc/make/git/libssl-dev/libpq-dev/
+      # libsqlite3-dev, so we only add what's missing for this crate:
+      #   - protobuf-compiler: prost/scale-codec build scripts.
+      #   - postgresql-client: psql smoke test against toolkit-cache.
+      #   - pkg-config:        for -sys crates to discover system libs.
+      # No clang/libclang — the e2e test binary doesn't pull rocksdb or
+      # any other bindgen-using crate (subxt + sqlx + redb only). The
+      # Earthfile installs clang because it builds the full node +
+      # runtime wasm, a much heavier dep set than we need here.
+      # Clear apt caches afterwards so the ~4 h cargo build has disk
```
LGLO: What `cargo build`? There is `cargo test` that builds some code - does it really take 4h or is majority of it running the tests actually?

LGLO (dismissed):

.github/workflows/nightly-run-cnight-e2e-qanet.yml
rsporny: Fair point — that comment is self-contradictory. `rust-toolchain.toml` is the source of truth for the Earthfile's `build-prepare` (which `grep`s the channel out of it at line 637), but this workflow's image tag and the Earthfile's `subxt` target each hardcode `1.95` separately and have to be bumped by hand. Calling it SOT while duplicating it twice was sleight of hand.

Reworded in 655ee1f7 to acknowledge the duplication outright and point at #1661, which I filed to track collapsing this back to one source via a pre-built CI image (the Dockerfile derives from `rust-toolchain.toml`, the workflow just references the published tag).

.github/workflows/nightly-run-cnight-e2e-qanet.yml
rsporny: You're right, "~4 h cargo build" was wrong — that wall-clock is the `cargo test` invocation, and the majority of it is Cardano stability waits, not compile. The release compile itself is ~8 min (per the second trial run in this PR's testing evidence); the multi-hour tail is the `k=432` finality window the e2e suite waits through.

Reworded in 655ee1f7 to call out compile vs. waits explicitly. The disk-headroom point still stands but now correctly aimed at the ~8 min window where `target/` actually grows.

LGLO (approved):

gilescope:
Why isn't this an Earthfile target? Then you'd have sudo and it would work on that machine as it would be isolated.

gilescope (changes requested):
if we're doing containers let's do them in earthly. We've already got all the local env stack running in DinD in earthly so should be possible

rsporny:
@gilescope you're right — this is architecturally cleaner than what I shipped, and there's working prior art in the repo: `Earthfile:1513` (`local-env-e2e`) already runs `cargo test --test e2e_tests` from a `FROM +prep` target. The qanet variant would be essentially the same recipe with `--features qanet`, `TOOLKIT_CACHE_DB_URL` threaded in via `--secret`, and `--test-threads=16`.

Wins over the current job-level `container:` approach:

- **No dep duplication.** `+prep`/`+build-prepare` already declare every apt dep this test needs (and matches the production build, removing the drift LGLO called out above).
- **No root-owned workspace leakage.** Earthly's buildkit context isn't bind-mounting the runner's `/__w/...` tree, so the cleanup step we added (#1658 / commit `02a501f2`) becomes unnecessary.
- **Same shape as the rest of the repo's CI.** `continuous-integration.yml` and friends already run via Earthly on `tier:large`; this would stop being the odd one out.

Reasons I kept the runtime-install approach for this PR specifically:

- **Minimal blast radius.** One workflow file, easy to revert if the next nightly surfaces something I missed; the Earthly migration touches an Earthfile target plus a workflow rewrite.
- **The bug is in front of us.** The nightly has been red since the original move to self-hosted (#1578), and getting it green at all unblocks operational signal. Larger refactors are easier to do on top of a green baseline.

Things I'd want to verify before doing the Earthly migration as a follow-up:

1. **Log streaming.** Earthly buffers `RUN` output by default. A 4 h test run with periodic Cardano-stability progress logs needs those visible in near-real-time in the GitHub Actions UI; need to confirm whether `--no-output-buffering` (or equivalent) does the right thing for a long RUN.
2. **Egress IP for the NLB allowlist.** The toolkit-cache RDS only accepts the runner's static IP. Need to confirm Earthly's buildkit container egresses through the host's IP (it almost certainly does — Docker bridge SNAT — but worth verifying with a `+smoke-test` target before the migration lands).
3. **`--no-cache` semantics.** We want the test to actually re-run every nightly, not be served from buildkit's cache; needs `RUN --no-cache cargo test ...` or `cargo test`'s output classified as non-cacheable.

I've updated #1661 to fold this in — that issue is now framed as "move test deps out of runtime; evaluate Earthly target (preferred, has prior art) vs. baked CI image (the original suggestion)." Happy to drive that as a separate PR once #1658 lands.

gilescope (approved):

PR #1677, opened 2026-06-11
test: c2m-bridge subminimal-transfer edge cases

### LINKED ISSUE #1248: [C-to-M Bridge] Accumulate too small transactions

rsporny:
# Overview

QA follow-up for #1248 ("Accumulate too small transactions"). The c2m-bridge subminimal-transfer accumulator has a happy-path e2e (PR #1654) and one unit test, but several branches in `handle_subminimal_transfer` were untested. This PR fills those gaps at unit level.

**Unit tests** added to `pallets/c2m-bridge/src/tests.rs`:
- Strict `sum > threshold` boundary — three tests for sum just below / equal / just above the threshold (equality must NOT flush).
- Post-flush state reset — after a flush, the next subminimal restarts the accumulator from zero (verified via storage, not just events).
- Routing precedence over `TransferRecipient::Invalid` and unapproved User — subminimal-Invalid / subminimal-unapproved-User must accumulate, not emit `InvalidTransfer` / `UnapprovedTransfer`.
- Non-interference between regular and subminimal — a regular approved User transfer in between two subminimals must take the regular path and leave the accumulator untouched.

A storage-level e2e assertion (reading `SubminimalTransfers` between submissions) was tried and then reverted — the accumulator's `(count, sum)` is an implementation detail, and the transition-level guarantees are better expressed at unit level. The existing e2e already covers the user-observable behavior via flush events and `UnlockToTreasury`.

Two earlier test cases (`single_subminimal_above_threshold_flushes_immediately`, `governance_lowering_threshold_flushes_existing_accumulator`) were dropped per LGLO review — the first tested a code path only reachable with invalid configuration (`flush_threshold < c_to_m_bridge_min_amount`), the second pinned down accumulator behaviour across mid-flight config changes that isn't a spec requirement.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

Unit tests — `cargo test -p pallet-c2m-bridge`:

```
running 16 tests
test tests::subminimal_no_flush_just_below_threshold ... ok
test tests::subminimal_no_flush_at_exact_threshold ... ok
test tests::subminimal_flushes_just_above_threshold ... ok
test tests::subminimal_state_resets_after_flush_and_accumulates_again ... ok
test tests::subminimal_invalid_recipient_accumulates_not_unlocks ... ok
test tests::subminimal_unapproved_user_accumulates_not_unlocks ... ok
test tests::regular_transfer_does_not_disturb_subminimal_accumulator ... ok
... (existing tests also pass)
test result: ok. 16 passed; 0 failed
```

E2E (existing test, unchanged) — verified against local-env:

```
E2E_SKIP_DEPLOY_GATE=1 cargo test --test e2e_tests --no-default-features --features local \
  subminimal_transfers_accumulate_and_flush_on_threshold_breach -- --test-threads=1 --nocapture
...
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 31 filtered out; finished in 250.80s
```

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other:
- [x] N/A 

## Links

Closes #1248

### DIFF
--- a/pallets/c2m-bridge/src/tests.rs
+++ b/pallets/c2m-bridge/src/tests.rs
@@ -269,30 +269,6 @@ fn subminimal_flushes_just_above_threshold() {
 }
 
 #[test]
-fn single_subminimal_above_threshold_flushes_immediately() {
-	new_test_ext().execute_with(|| {
-		// Threshold below a single subminimal amount → the very first transfer
-		// flushes with count=1, exercising the `count: 0 → 1` saturating add
-		// in the same call as the flush.
-		set_flush_threshold(50);
-		C2MBridge::handle_incoming_transfer(subminimal_transfer());
-		assert_eq!(
-			pallet::SubminimalTransfers::<Test>::get(),
-			SubminimalTransfersState { count: 0, sum: 0 }
-		);
-		assert_eq!(mock_pallet::Transfers::<Test>::get().len(), 1);
-		assert_eq!(
-			subminimal_events(),
-			vec![Event::SubminimalFlushTransfer {
-				amount: 90,
-				count: 1,
-				midnight_tx_hash: [0u8; 32],
-			}],
-		);
-	})
-}
-
-#[test]
 fn subminimal_state_resets_after_flush_and_accumulates_again() {
 	new_test_ext().execute_with(|| {
 		set_flush_threshold(180);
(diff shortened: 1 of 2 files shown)


tests/e2e/tests/c2m_bridge.rs
```diff
@@ -431,6 +431,13 @@ async fn subminimal_transfers_accumulate_and_flush_on_threshold_breach() {
             }
         });
 
+        // Read the pallet accumulator directly — events alone can't catch a
+        // regression where the pallet silently drops a transfer (no flush event,
+        // but storage also not advanced).
```
LGLO: If any of three transfer is silently dropped, then there would be no flush after the 3rd one and no flush event that we assert on.

rsporny: dropping this

pallets/c2m-bridge/src/tests.rs
```diff
@@ -203,6 +203,261 @@ fn subminimal_transfer_handling() {
...
+			pallet::SubminimalTransfers::<Test>::get(),
+			SubminimalTransfersState { count: 1, sum: 90 }
+		);
+		assert!(subminimal_events().is_empty());
+		assert!(mock_pallet::Transfers::<Test>::get().is_empty());
+	})
+}
+
+#[test]
+fn governance_lowering_threshold_flushes_existing_accumulator() {
```
LGLO: This is misleading for me. It suggest the scenario is that (without a transfer after lowering threshold):
* accumulator has N
* config is set to M < N
* flush is observed

There were no requirements about behavior in such scenario. I would drop this test as it makes up requirements.

Anyway, there is new `C2MBridge::handle_incoming_transfer(subminimal_transfer());` in line 401 that triggers flush according to the updated value.

pallets/c2m-bridge/src/tests.rs
```diff
@@ -203,6 +203,261 @@ fn subminimal_transfer_handling() {
...
+				amount: 180,
+				count: 2,
+				midnight_tx_hash: [0u8; 32],
+			}],
+		);
+	})
+}
+
+#[test]
+fn single_subminimal_above_threshold_flushes_immediately() {
```
LGLO: This is a test case for behavior when assumptions are not met. I would drop it. If assumptions (correct configurations) are not met, then every behavior is correct. This is how logic works.

rsporny: fair point, dropping it

pallets/c2m-bridge/src/tests.rs
rsporny: Fair point — the test name read as "lowering threshold alone triggers a flush" but the flush actually came from the follow-up transfer at line 401, and accumulator persistence across mid-flight config changes isn't a spec requirement. Dropped in cead85ca alongside the `single_subminimal_above_threshold_flushes_immediately` case from the other thread.

LGLO (approved):

PR #1731, opened 2026-06-19
fix(toolkit): emit successful txs from batch-single-tx on partial failure

chrispalaskas:
## Problem

`generate-txs batch-single-tx` builds N transfers concurrently, then **discards every successfully-built tx if even one transfer fails**. It returns `PartialFailure` before serializing anything, so `--dest-file` is never written.

This wastes the whole batch. Downstream load appliers (the perf repo's `tx_load_applier.py` → `generate_txs_round_robin.py`) see the non-zero exit, delete the batch directory, and skip submission entirely. In practice a 250-transfer batch with even one failed transfer sends **zero** transactions — so the load test produces no traffic at all:

```
Built tx ... index=250 total=250
failed to build transactions: 3 of 250 transfers failed
Error: BuildTransactions(PartialFailure { succeeded: 247, failed: 3 })
❌ batch-single-tx failed with exit code 1 and produced no transactions
```

All 247 valid txs are thrown away because 3 failed.

## Fix

Add an opt-in `--emit-partial-batch` flag (**off by default**, per review discussion — downstream decides whether a partial batch is acceptable):

- **Flag off (default):** unchanged all-or-nothing behavior — any failed transfer fails the whole batch with `PartialFailure`, nothing is emitted.
- **Flag on:** every tx that built successfully is written and the process exits 0 (with a warning); only a batch where **all** transfers fail errors out (`AllFailed`).

To make partial emission safe, transfers touching the same wallet are now serialized via per-seed locks, and a failed transfer's wallet mutations (spent coins, pending change, DUST) are rolled back to a pre-transfer snapshot — so emitted txs never depend on state from a failed, never-submitted tx.

No wrapper change is needed — the existing `valid_count == 0 && returncode != 0` guard in `generate_txs_round_robin.py` handles the `AllFailed` case, and the applier submits partial batches when the flag is passed.

## Testing

- `cargo check -p midnight-node-toolkit`, `cargo clippy -p midnight-node-toolkit` clean.
- `cargo test -p midnight-node-toolkit --lib`: 97 passed, including new tests:
  - partial failure + `--emit-partial-batch` → `Ok` with exactly the successful txs
  - partial failure without the flag → `PartialFailure` error


HarleysCodes (commented):
Thanks for the fix — the behavior change is correct and the rationale (don't discard 247 valid txs because 3 failed) is exactly right. A few concrete observations from reading the diff:

**Behavior change is sound.** The rename of `PartialFailure` → `AllFailed` is a clean refactor: the variant is only constructed in one place and only matched by the `BuildTxs` `Error` impl in this same file, so there's no other call site to update. Verified via `gh search code 'PartialFailure'` returning only the original construction site. The new `succeeded > 0` short-circuit correctly turns a partial-failure error into a successful return with the partial batch.

**Suggestion: add a test for the new behavior.** The PR description only mentions `cargo check -p midnight-node-toolkit` passing — there's no test that exercises `failed > 0 && succeeded > 0` and verifies the function returns `Ok` with the partial batch. Even a small unit test in `batch_single_tx.rs` (constructing a builder with 5 mock transfers where 2 fail) would prevent regressions and document the new contract. Given the perf-repo load applier in the description relies on the new `Ok` path, locking the behavior in a test is high-value.

**Minor: warn-message arg order.** In the new `tracing::warn!`:

```rust
tracing::warn!(
    succeeded = succeeded,
    failed = failed,
    "Partial batch: {} of {} transfers failed; emitting {} successful txs",
    failed,
    succeeded + failed,
    succeeded
);
```

The format string lists arguments as `failed / total / succeeded` but the structured fields above are `succeeded = succeeded, failed = failed`. I'd swap the format placeholders to `{}/{}, succeeded {}` → `failed/total/succeeded` for consistency with the structured fields, OR move the structured fields below the format string with `failed = {}`-style placeholders. Cosmetic, but the dual layout makes it easy to misread at a glance.

**Pre-existing edge case, not a blocker.** If the batch is empty (0 transfers, `succeeded == 0 && failed == 0`), the new code returns `Ok(SerializedTxBatches { batches: vec![] })`. Pre-existing behavior was the same, so this PR doesn't regress anything — but worth filing a follow-up if the team considers an empty batch an error condition (e.g. missing `--num-transfers` flag). Out of scope for this fix.

**Optional doc nit.** The PR body mentions the perf-repo wrapper (`tx_load_applier.py` → `generate_txs_round_robin.py`) but doesn't link it. A maintainer reviewing this would have to grep for those script paths manually. A couple of `<url>` references would let the reviewer verify the claim that 'no wrapper change is needed.'

Overall the fix is small, focused, and addresses a real waste-of-work bug. The main ask is adding a test; the rest is polish. Happy to follow up on any of these.

util/toolkit/src/tx_generator/builder/builders/common/batch_single_tx.rs
```diff
@@ -249,8 +249,21 @@ impl<C: BuilderContext<DefaultDB>> BuildTxs for BatchSingleTxBuilder<C> {
...
 		progress.finish(format!("batch-single-tx: {} succeeded, {} failed", succeeded, failed));
 
+		// Emit every tx we managed to build, even on partial failure. Discarding the
+		// successful txs because a few transfers failed wastes the whole batch and starves
+		// downstream load appliers. Only treat a batch with zero successes as an error.
 		if failed > 0 {
-			return Err(BatchSingleTxError::PartialFailure { succeeded, failed });
+			if succeeded == 0 {
+				return Err(BatchSingleTxError::AllFailed { failed });
+			}
```
HarleysCodes, line 347: Tiny nit on the warn-message: structured fields above are `succeeded = succeeded, failed = failed` but the format placeholders consume them as `failed, succeeded + failed, succeeded` — i.e. the format string reads `{failed} of {total} transfers failed; emitting {succeeded} successful txs`. Either swap the format placeholders to match the structured-field ordering, or reorder the args. Cosmetic, but the dual layout is easy to misread.

HarleysCodes (commented):
Two small follow-up inline comments after the main PR-level review above.

util/toolkit/src/tx_generator/builder/builders/common/batch_single_tx.rs
```diff
@@ -249,8 +249,21 @@ impl<C: BuilderContext<DefaultDB>> BuildTxs for BatchSingleTxBuilder<C> {
...
+			tracing::warn!(
+				succeeded = succeeded,
+				failed = failed,
+				"Partial batch: {} of {} transfers failed; emitting {} successful txs",
+				failed,
```
HarleysCodes, line 356: This is the new code path worth a test: `failed > 0 && succeeded > 0` must return `Ok` with the partial batch, not `Err`. The perf-repo load applier in the PR description relies on this exact behavior. Even a small test that constructs a `BatchSingleTxBuilder` with 5 transfers where 2 deliberately fail and verifies the function returns `Ok(SerializedTxBatches)` with 3 entries would lock the contract and document the new semantics for future maintainers. Without it, a future refactor could revert to `Err` on partial failure and break the load-test pipeline silently.

HarleysCodes (commented):
Additional inline comment.

util/toolkit/src/tx_generator/builder/builders/common/batch_single_tx.rs
```diff
@@ -246,8 +249,21 @@ impl<C: BuilderContext<DefaultDB>> BuildTxs for BatchSingleTxBuilder<C> {
 
 		progress.finish(format!("batch-single-tx: {} succeeded, {} failed", succeeded, failed));
 
+		// Emit every tx we managed to build, even on partial failure. Discarding the
+		// successful txs because a few transfers failed wastes the whole batch and starves
+		// downstream load appliers. Only treat a batch with zero successes as an error.
 		if failed > 0 {
```
LGLO, line 344: I would make sure that process returns a valid JSON for downstream instead of playing with error codes here.
As pointed by another comment: 0 of 0 is success, 0 of 1 if failure, but in both cases downstream doesn't have much to work on. It seems that return code alone isn't sufficient here.

Also, is 1 okay out of 250 a success? It hardly seems so to me.

It seems simpler to me if downstream makes decision what is enough for it.

chrispalaskas: What if we have a flag? For my case, when I load, I need any tx to be sent, better than failure. For someone else he can have it off.

util/toolkit/src/tx_generator/builder/builders/common/batch_single_tx.rs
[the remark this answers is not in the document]
chrispalaskas: Addressed in 56799cfe4. Transfers touching the same wallet are now serialized with per-seed async locks (source + funding seed, acquired in a canonical order), and the affected wallets are snapshotted before the build — on any failure (including `ProvingFailed` after `build_inputs` has already spent local coins/DUST or created pending change) the wallets are rolled back to the pre-transfer snapshot. So a failed, un-emitted transfer can no longer leak state into later transfers built from the same wallet, and transfers on distinct wallets keep building fully in parallel.

chrispalaskas: Fixed in 56799cfe4 — the warning now uses shorthand structured fields and inline format captures in the same order: `tracing::warn!(failed, total, succeeded, "Partial batch: {failed} of {total} transfers failed; emitting {succeeded} successful txs")`.

util/toolkit/src/tx_generator/builder/builders/common/batch_single_tx.rs
chrispalaskas: Added in 56799cfe4: `test_batch_single_tx_partial_failure_emits_successful_txs_with_flag` builds 3 transfers where the middle one deliberately fails input selection (amount above the genesis balance) and asserts the result is `Ok(SerializedTxBatches)` with exactly the 2 successful txs. Its counterpart `test_batch_single_tx_partial_failure_errors_without_flag` locks the default behavior: the same batch without `--emit-partial-batch` must return the `PartialFailure` error.

util/toolkit/src/tx_generator/builder/builders/common/batch_single_tx.rs
chrispalaskas: Implemented the flag in 56799cfe4: `--emit-partial-batch`, off by default. Default behavior is back to strict all-or-nothing (any failure → `PartialFailure` error, nothing emitted), so downstream decides whether partial batches are acceptable by opting in. The error message also points at the flag: `N of M transfers failed; pass --emit-partial-batch to emit the K successful txs instead`. Whether "1 ok out of 250" is a success is then the caller's call — for the load-test applier any traffic beats none, so it opts in.

chrispalaskas:
No interest from people, closing. Also it's a very specific test use case, a normal user might like the whole batch to fail if one tx fails.

PR #1732, opened 2026-06-20
Chore: protect genesis files

justinfrevert:
# Overview

Protect changes against genesis files of live networks. 

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [x] N/A


scripts/ci/check-frozen-res.sh
```diff
@@ -0,0 +1,38 @@
...
+PROTECTED_RE='^res/(mainnet|qanet|preprod|preview)/'
+
+# --no-renames so a rename *out of* a protected folder surfaces as a deletion of
+# the protected path (and is therefore caught), instead of an R entry we'd have
+# to special-case. The default diff filter already includes A/M/D, so additions
+# of new files inside a protected folder are caught too.
+changed="$(git diff --name-only --no-renames "${BASE_REF}...HEAD" | grep -E "$PROTECTED_RE" || true)"
+
+if [ -n "$changed" ]; then
+  echo "::error::Frozen res/ network config changed. These files are deployed artifacts and must never be modified:"
```
LGLO: This sentence is false in few distinct ways:
* some of the config files were modified or even created **after** the chain-spec was created and network started
* "never" - preview is perhaps going to be restarted
* some of the config file should be changed, because they are just wrong and not consistent with chain-spec

justinfrevert: True. Only updatable in rare cases, which get flagged, and we can review case-by-case. Updated the language to allow for that.

LGLO (approved):
OK, but it makes fixes like #1730 harder to land. Main question for now is if the existing chain-spec files are really the exact same chain-specs that were used for genesis of the 4 networks.

gilescope (approved):

PR #1796, opened 2026-06-30
feat(local-env): bridge-funded `local` network + seed cNIGHT at genesis

### LINKED ISSUE #1778: test(e2e): seed local-env Cardano cNIGHT at genesis + fix stale ICS/Reserve addresses (#1773 Part A)
Part A of #1773. **Separate PR.** Prerequisite for Part B — it establishes the cooperative genesis state the monitor asserts against.

## Problem

On local-env the Cardano side mints **0 cNIGHT** at genesis (`res/local-environment/ics-config.json` and `reserve-config.json` both have `total_amount: 0, utxos: []`), while the Midnight side already carries the full pools via the committed `res/genesis/genesis_state_undeployed.mn` (`M.R ≈ 5B`, `M.L ≈ 16.8B`, `M.U = S − M.R − M.L ≈ 2.2B`). So the cross-chain invariant `M.U ≤ C.L` (2.2B ≤ 0) is violated from genesis — the Cardano side must be seeded to mirror the Midnight pools.

Separately, the static ICS/Reserve **validator addresses** in `res/local-environment/*-config.json` are **stale**. The real validators are `illiquid_circulation_supply.ics_forever` / `reserve.reserve_forever`, compiled fresh each run (one-shot-param'd) and surfaced in `contracts-info.json` as **"ICS Forever" / "Reserve Forever"** — addresses differ from the static files. `midnight-setup` already patches Council/Tech-Auth addresses from `contracts-info.json` but **not** ICS/Reserve.

## Scope

**A1 — Patch ICS/Reserve addresses at runtime.** In `local-environment/src/networks/local-env/configurations/midnight-setup/entrypoint.sh`, mirror the existing Council/Tech-Auth patching: read `"ICS Forever"` / `"Reserve Forever"` addresses from `/runtime-values/contracts-info.json` and `jq`-patch them into `/tmp` copies of `ics-config.json` / `reserve-config.json` (`illiquid_circulation_supply_validator_address` / `reserve_validator_address`), then point `CHAINSPEC_ICS_CONFIG` / `CHAINSPEC_RESERVE_CONFIG` at the `/tmp` copies. Keep `asset.policy_id = d2dbff…`, `asset_name = ""`. Transparent to E1–E4 (they read the address from bridge storage).

**A2 — Seed Cardano cNIGHT at genesis (pre-observation-checkpoint).** At the **end of** `local-environment/.../contract-compiler/compile-contracts.sh` (after deploy → `contracts-info.json` exists; completes before `midnight-setup` via `depends_on: service_completed_successfully`, i.e. pre-`initial_data_checkpoint`): mint **24,000,000,000,000,000** (24B) cNIGHT under policy `d2dbff…` (the `test_cnight_no_audit.tcnight_mint_infinite` infinite-mint script) with the funded-address key, and distribute:

| Target | Amount (STARS) | Notes |
|---|---:|---|
| "Reserve Forever" addr | 5,000,000,000,873,988 | `C.R = M.R`; unit inline datum |
| "ICS Forever" addr | 2,200,000,000,000,000 | `C.L = M.U`; unit inline datum; **no** bridge metadata (label 6500973) |
| funded/faucet addr | 16,799,999,999,126,012 | `C.U = M.L`; circulating pool the flood spends |

Follow the tx patterns in `scripts/cnight-generates-dust/receive_cnight.sh` (mint) and `lock_to_ics.sh` (send-to-ICS + datum), reusing the container's `bun cli` / cardano-cli + `cnight_policy.plutus`. Wait for confirmation before the script exits. Amounts as named constants commented to the committed-genesis baseline.

## Critical correctness invariant

The ICS seeding output **must land before** the Cardano tx that `midnight-setup` selects for `initial_data_checkpoint`, so the bridge treats seeded ICS cNIGHT as pre-existing locked supply (not a transfer to sweep). Part B's genesis quiescence assertion (`C.* == M.*`) is the canary.

## Acceptance criteria

- [x] `midnight-setup` patches ICS/Reserve validator addresses from `contracts-info.json`; bridge storage reflects the real "Forever" validators.
- [x] After a fresh local-env start, kupo reports 24B total cNIGHT under policy `d2dbff…`; ICS/Reserve "Forever" addresses hold 2.2B / 5B; faucet holds 16.8B.
- [x] Seeded ICS cNIGHT is **not** observed/swept by the bridge (no spurious `UnlockToTreasury` at startup).
- [x] E1–E4 (`tests/e2e/tests/c2m_bridge.rs`) still pass with the address patch.

## References

- Plan + model: #1773
- `node/src/chain_spec/mod.rs` (bridge `MainChainScripts` from ics/reserve configs)
- `util/toolkit/src/genesis_generator.rs` (pool formula: `treasury = ics.total`, `reserve_pool = reserve.total`, `locked = MAX_SUPPLY − reserve − treasury`; `DistributeReserve` funds dev seeds ~1B from reserve)

### LINKED ISSUE #1792: Dedicated `local` network for local-environment (decouple from res/dev)

Introduce a first-class **`local`** network with its **own on-chain ledger identity**, so the dockerized **local-environment** has a single, self-contained config namespace and no longer depends on `res/dev`.

Today the local-environment's chain-spec assembly (`local-environment/.../midnight-setup/entrypoint.sh`) pulls config from three places and reuses the `undeployed` identity:
- `res/local-environment/` — cnight/ics/pc-chain/reserve
- `res/dev/` — federated-authority + system-parameters
- `res/genesis/genesis_*_undeployed.mn` — the ledger genesis, itself built from `res/dev/` (`Earthfile:248-261`)

## User Need

Developers working on the local-environment want a dedicated network that owns its configuration end-to-end, rather than a partial override layer (`res/local-environment/`) sitting on top of `res/dev/`. The current split also creates a latent inconsistency: the genesis ledger state is generated from `res/dev/`, while the runtime substrate/observation config comes from `res/local-environment/`.

## Expected Benefit

- Single source of truth: one `res/local/` directory drives both genesis generation and the runtime chain-spec.
- Clean separation: `res/dev` / `CFG_PRESET=dev` / `UndeployedNetwork` are untouched; `local` is additive.
- A genuinely independent network (`mn_addr_local1…` addresses) usable via both `cargo run` (`CFG_PRESET=local`) and the docker stack.

## Details

**Decisions:** new ledger identity (network id `local`); full preset + registered network; named `local`.

**Key constraint:** in `util/toolkit/src/genesis_generator.rs:180`, `--network` is passed straight into `LedgerState::with_genesis_settings(network_id, …)` and wallet address generation — so it is *both* the bech32 address prefix and the genesis-file selector. A new identity therefore changes addresses and invalidates pre-baked `undeployed` test transactions used by `tests/e2e/`.

**Workstreams:**
- **A. `res/local/`** — new config namespace (full file set, mirroring `res/dev/`), seeded by merging `res/dev/*` + `res/local-environment/*`. Must be named `res/local` to match `get_res_preset_dir()` (`node/src/command.rs:347`). Retire `res/local-environment/`.
- **B. `LocalNetwork`** — new struct in `res/src/networks/definitions.rs` (`id()=="local"`, reads `res/local/*` + `genesis_*_local.mn`); split node dispatch `node/src/cfg/mod.rs:305` so `"local" => LocalNetwork`, `"dev" => UndeployedNetwork` (repurposes the `--chain local` alias; no current caller relies on it).
- **C. `res/cfg/local.toml`** — preset (copy of `dev.toml`) with `chainspec_id="midnight_local"`, paths → `res/local/*` + `genesis_*_local.mn`, args select the `local` chain.
- **D. Earthfile** — extend the `undeployed` special-casing (config copy + hardcoded faucet seeds) to `local`; add `rebuild-genesis-state-local`; regenerate `*_local.mn` test-tx artifacts. Keep `util/toolkit/test-data/genesis/` on `undeployed`.
- **E. Docker entrypoint** — repoint all `res/dev`/`res/local-environment` reads to `res/local`, genesis → `genesis_*_local.mn`, `CHAINSPEC_ID=midnight_local`. Node entrypoints need no change (they load `--chain=/shared/chain-spec.json`).
- **F. New-identity blast radius** — add a `local` module to the res crate (`res/src/lib.rs`) exposing regenerated `transactions::*`; migrate `tests/e2e/` (`rpc_abuse`, `operational`, `contract_state`) from `undeployed::transactions::*` to `local::transactions::*`. Standalone `scripts/tests/toolkit-*-e2e.sh` and toolkit unit fixtures stay on `undeployed`.

**Out of scope / untouched:** `res/dev`, `CFG_PRESET=dev`, `UndeployedNetwork`, single-node toolkit e2e scripts, toolkit unit-test fixtures, benchmark genesis (`ledger/src/host_api/ledger_9.rs:522`).

rsporny:
# Overview

Adds a first-class **`local`** network for the dockerized local-environment — its own ledger identity (`mn_addr_local1…`), decoupled from `res/dev` — funded **entirely over the real cNIGHT→mNIGHT bridge**: genesis ships **no** funded wallets, so the cross-chain pool invariants hold from block 0 (fixes #1778). Also revives the e2e RPC tests that no longer depend on funded-genesis fixtures (#1792).

Why: the local-env previously reused the `undeployed` identity and pulled config from three places (`res/dev`, `res/local-environment`, `res/genesis/*undeployed*`), and the Cardano side minted **0 cNIGHT** at genesis while the Midnight side carried the full NIGHT pools — violating `M.U ≤ C.L` from genesis and leaving the bridge nothing consistent to observe.

### `local` network
- `res/cfg/local.toml` preset + `res/local/` config namespace. `build-spec` with `CFG_PRESET=local` sources everything from `res/local/` at runtime via the dynamic `CustomNetwork` path — no compile-time network (`dev` stays the built-in `UndeployedNetwork`). `res/local-environment/` retired.
- Unfunded genesis `res/genesis/genesis_{state,block}_local.mn` (Earthfile `rebuild-genesis-state-local`, `FUND_FAUCET_WALLETS=false`).
- `midnight-setup` patches all `res/local/` config from the **deployed** contract addresses (incl. the cNIGHT policy id from the compiled blueprint) and anchors the bridge `initial_data_checkpoint` to the cNIGHT seeding tx (seeded ICS supply treated as pre-existing, not swept to Treasury).

### cNIGHT bridge funding (cardano-cli + toolkit; no dedicated images, no governance round)
- **Genesis seed** — a `mint-cnight-supply` service (`contract-compiler → mint-cnight-supply → midnight-setup`) mints the full cNIGHT supply in one tx and splits it to mirror the Midnight genesis pools: Reserve `C.R=M.R`, ICS `C.L=M.U` (inline unit datum, no bridge metadata), faucet `C.U=M.L`.
- **Dev wallet `0x..01`** — the same service then sends a c2m bridge transfer locking part of the circulating cNIGHT to ICS for it; `midnight-setup` pre-approves its tx hash in the c2m-bridge genesis config via the new **`approved_txs`** entry threaded into the chainspec (#1809). Post-genesis, `init-mnight-faucet` claims it (`claim-rewards --claim-kind cardano-bridge`, feeless/self-signed) and registers the wallet's DUST address so `0x..01` can transact. Idempotent via a `runtime-values/mnight-faucet-ready` marker.

### e2e tests (#1792)
- Revived the **valid-deploy** and **contract-state** RPC tests (real deploys against the bridge-funded `0x..01`) and ported the toolkit **multi-dest no-hang** check from a bash script into the Rust suite (dropped `scripts/tests/toolkit-multi-dest-e2e.sh`). Hardened against shared-wallet DUST contention under CI's `--test-threads=6`. The suite is gated on `init-mnight-faucet` completing.
- **Pre-dispatch validation** (ContractNotPresent / replay / malformed) is left to `pallet-midnight`'s unit tests, which assert it precisely with mocked state — no live chain, no dust confound.

### Supporting fix
- `ogmios-client`: tolerate absent Plutus cost models in protocol params (the local-env Cardano omits them).

## 🗹 TODO before merging

- [x] Ready
- Node version bump not done — additive test network + local-env tooling + an additive genesis config field (`approved_txs`), no existing-runtime change. Bump if reviewers prefer.

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

- The full `--features local` e2e suite runs the real path end to end in CI (`Local Environment Tests`): dockerized Cardano bring-up → cNIGHT genesis seeding → dev-wallet bridge funding → valid-deploy / contract-state / multi-dest.
- Reproduced locally at the CI concurrency (`--test-threads=6`), exercising and recovering from the shared-wallet DUST-contention paths.
- Pre-dispatch validation (ContractNotPresent / replay / malformed) is covered by `pallet-midnight` unit tests.
- Genesis verification on a full local-env run: `toolkit show-night-pools` reserve / unlocked / locked mirror the seeded Cardano pools (total 24B NIGHT); `M.* == C.*` at genesis and `M.U ≤ C.L` holds with equality; no startup `UnapprovedTransfer` / `UnlockToTreasury` (seeded ICS treated as pre-existing); the dev wallet `0x..01` ends funded + DUST-registered.

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other:
- [x] N/A 

## Links

Closes #1778
Closes #1792
Refs #1809


res/local/ics-addresses.json
```diff
@@ -0,0 +1,7 @@
+{
```
LGLO: For local environment we should have either:
* only `ics-config.json`
* only `ics-addresses.json` and in `midnight-setup` run a command that generates `ics-config.json` from `ics-addresses.json` using the local environment Cardano state

local-environment/src/networks/local-env/configurations/mint-cnight-supply/mint-cnight-supply.sh
```diff
@@ -0,0 +1,199 @@
...
+
+ICS_ADDR=$(cat "$ICS_ADDR_FILE")
+RESERVE_ADDR=$(cat "$RESERVE_ADDR_FILE")
+echo "ICS Forever address:     $ICS_ADDR"
+echo "Reserve Forever address: $RESERVE_ADDR"
+
+# Verify the compiled cNIGHT policy is the one the bridge observes.
+POLICY_ID=$(cardano-cli latest transaction policyid --script-file "$CNIGHT_PLUTUS")
+echo "cNIGHT policy id: $POLICY_ID"
+if [ "$POLICY_ID" != "$EXPECTED_POLICY_ID" ]; then
```
LGLO: We have bridge tests that would fail if something is misconfigured.
This check doesn't look complete, because policy id is not enough. Asset id is what is required.

I mean: we can check it, but this occupies lines and doesn't prove that bridge will work. Even if we added check for token name (I think this part is misconfigured) then bridge can have wrong address in the config. Then it can have wrong observability checkpoint... a lot of code testing that setup is correct but not testing our system under test.

res/local/ics-addresses.json
LGLO: If we go with `ics-addresses.json`, then we don't need the file anyway. All the content is known after seeder has run: address and token

local-environment/src/networks/local-env/configurations/mint-cnight-supply/mint-cnight-supply.sh
```diff
@@ -0,0 +1,199 @@
...
+ICS_STARS=2200000000000000          # C.L = M.U (unlocked = S - M.R - M.L)
+FAUCET_STARS=16799999999126012      # C.U = M.L (locked_pool)
+TOTAL_MINT_STARS=24000000000000000  # S
+# min-UTxO lovelace bundled with each cNIGHT output (matches the e2e/manual scripts).
+MIN_UTXO_LOVELACE=1500000
+
+# The bridge is configured to observe this cNIGHT minting policy on every non-mainnet
+# environment. We assert the compiled policy matches so a contract change can't
+# silently mint under a policy the bridge ignores.
+EXPECTED_POLICY_ID=d2dbff622e509dda256fedbd31ef6e9fd98ed49ad91d5c0e07f68af1
```
rsporny: The moment cnight policy id changes this will result in false positive. We should take this from runtime values (eg. contracts-info.json)

local-environment/src/networks/local-env/configurations/mint-cnight-supply/mint-cnight-supply.sh
rsporny: instead of comparing with hardcoded value, we should compare with what was set in chain-spec (which doesn't exist yet, so I don't see a point in this here)

Earthfile
```diff
@@ -245,13 +245,20 @@ rebuild-genesis-state:
     # or if FUND_FAUCET_WALLETS=false (e.g., for mainnet)
     COPY --if-exists secrets/${NETWORK}-genesis-seeds.json /secrets/genesis-seeds.json
 
-    # Copy genesis config files (undeployed uses res/dev/)
+    # Copy genesis config files (undeployed uses res/dev/; local uses res/local/).
+    # undeployed and local are mock-main-chain dev networks: they have no cardano-tip.json
```
rsporny: local is not mock-main-chain

local-environment/src/commands/run.ts
```diff
@@ -241,6 +247,77 @@ async function runLocalEnvironment(runOptions: RunOptions) {
...
+ *
+ * - If it's already set (CI passes the just-built tag; or a power-user override), use it
+ *   as-is — the compose layer pulls/uses it.
+ * - Otherwise derive the content-hash tag for this checkout
+ *   (`ghcr.io/midnight-ntwrk/local-env-init-mnight-faucet:<git HEAD^{tree}>-<arch>`), which CI
+ *   publishes on merge, set it, and `docker pull` it. If the pull fails (unmerged branch, or
+ *   uncommitted changes to the faucet job) fall back to `earthly +init-mnight-faucet-image`,
+ *   which builds the SAME tag locally (no push).
+ */
+function resolveMnightFaucetImage(env: Record<string, string>) {
```
rsporny: that's a feature, shouldn't be added in this pr. pull should be available, if not devs have to built midnight-node image themselves, I don't think such auto resolution exists for midnight-node. drop it

local-environment/src/networks/local-env/configurations/contract-compiler/compile-contracts.sh
```diff
@@ -158,4 +158,20 @@ echo "✓ Contracts data exported successfully"
...
+# locked / reserved cNIGHT) and the compiled infinite-mint cNIGHT policy as a .plutus
+# envelope. The seeder runs in a cardano-cli-only container (no jq), so we extract
+# these here where jq + the compiled artifacts are available.
+echo "=== cNIGHT Seeding Inputs Export ==="
+jq -r '.[] | select(.name == "ICS Forever") | .address' "$OUTPUT_DIR/contracts-info.json" > "$OUTPUT_DIR/ics_forever.addr"
+jq -r '.[] | select(.name == "Reserve Forever") | .address' "$OUTPUT_DIR/contracts-info.json" > "$OUTPUT_DIR/reserve_forever.addr"
+cnight_code=$(jq -r '.validators[] | select(.title == "test_cnight_no_audit.tcnight_mint_infinite.else") | .compiledCode' "$PLUTUS_JSON")
+jq -n --arg code "$cnight_code" '{type: "PlutusScriptV3", description: "", cborHex: $code}' > "$OUTPUT_DIR/cnight_policy.plutus"
+echo "ICS Forever address:     $(cat "$OUTPUT_DIR/ics_forever.addr")"
+echo "Reserve Forever address: $(cat "$OUTPUT_DIR/reserve_forever.addr")"
```
rsporny: seems redundant, we should have what we need in runtime values, look at midnight-setup, it uses the same addresses from contracts-info.json

LGLO (dismissed):

local-environment/src/networks/local-env/configurations/init-mnight-faucet/init-mnight-faucet.sh
```diff
@@ -0,0 +1,115 @@
...
+  sleep 2
+done
+
+show_wallet() {
+  "$TOOLKIT" show-wallet --src-url "$NODE_URL" --seed "$FAUCET_SEED" 2>/dev/null
+}
+
+# 2. Wait for the bridge observation to make the transfer claimable. The mainchain
+#    follower needs the Cardano tx to reach stable depth + db-sync to index it, so give
+#    it a generous budget (~15 min; typically well under 2).
```
rsporny: this container is local-env only, 2 minutes should be max

local-environment/src/networks/local-env/configurations/init-mnight-faucet/init-mnight-faucet.sh
```diff
@@ -0,0 +1,104 @@
...
+  echo "Faucet already initialized ($FAUCET_MARKER_FILE present); skipping."
+  exit 0
+fi
+
+echo "=== init mNIGHT faucet (wallet seed ${FAUCET_SEED:0:6}..${FAUCET_SEED: -2}) ==="
+
+# 1. Wait for the node RPC to accept connections (depends_on service_started is not
+#    RPC-ready) — /dev/tcp probe, no extra tooling needed.
+echo "Waiting for $NODE_HOST:$NODE_PORT ..."
+for i in {1..180}; do
```
rsporny: this should be 1..30 (1 minute max)

local-environment/src/networks/local-env/configurations/init-mnight-faucet/init-mnight-faucet.sh
```diff
@@ -0,0 +1,104 @@
...
+show_wallet() {
+  "$TOOLKIT" show-wallet --src-url "$NODE_URL" --seed "$FAUCET_SEED" 2>/dev/null
+}
+
+# 2. Wait for the bridge observation to make the transfer claimable (the mainchain
+#    follower needs the Cardano tx at stable depth + db-sync to have indexed it).
+#    Local-env only, so 2 minutes is the cap.
+echo "Waiting for the faucet bridge transfer to become claimable..."
+CLAIMABLE=0
+for i in {1..24}; do
```
rsporny: 1 minute max, sleep 3s

local-environment/src/networks/local-env/configurations/init-mnight-faucet/init-mnight-faucet.sh
```diff
@@ -0,0 +1,104 @@
...
+  --funding-seed "$FAUCET_SEED" \
+  --amount "$CLAIMABLE" \
+  --claim-kind cardano-bridge
+
+# 4. Register for DUST production. The registration self-funds from the retroactive
+#    DUST the claimed NIGHT generates as it ages (~1 block/6 s on local-env), so retry
+#    a few times while it accrues instead of hardcoding an aging sleep.
+echo "Registering wallet for DUST production..."
+registered=false
+for i in {1..20}; do
```
rsporny: 1..10 max

local-environment/src/networks/local-env/configurations/midnight-setup/entrypoint.sh
```diff
@@ -129,35 +131,74 @@ jq --arg mapping_addr "$CNIGHT_MAPPING_VALIDATOR_ADDRESS" \
   | .mappings = {}
   | .utxo_owners = {}
   | .next_cardano_position = .observed_utxos.start
-  | .system_tx = null' res/local-environment/cnight-config.json > /tmp/cnight-config.json
+  | .system_tx = null' res/local/cnight-config.json > /tmp/cnight-config.json
 
 echo "Created cnight-config.json:"
 cat /tmp/cnight-config.json
 
+# Build the bridge ICS / reserve *observation* configs from the freshly-deployed validator
```
rsporny: we should cat the built configs

local-environment/src/networks/local-env/configurations/midnight-setup/entrypoint.sh
```diff
@@ -129,35 +131,74 @@ jq --arg mapping_addr "$CNIGHT_MAPPING_VALIDATOR_ADDRESS" \
...
+# addresses (ICS Forever / Reserve Forever) and the cNIGHT asset. The deployed addresses
+# (not the static res/local ones) are what the bridge actually locks to and observes, so the
+# c2m-bridge / ICS observation must watch these. The observation starts empty (utxos /
+# total_amount = 0) and tracks the local Cardano forward; genesis pools come from the
+# committed genesis state.
+CNIGHT_POLICY_ID=$(jq -r '.addresses.cnight_policy_id' res/local/cnight-config.json)
+CNIGHT_ASSET_NAME=$(jq -r '.addresses.cnight_asset_name' res/local/cnight-config.json)
+echo "Building ICS/reserve observation configs (ICS=$ICS_FOREVER_ADDRESS, Reserve=$RESERVE_FOREVER_ADDRESS)"
+jq -n --arg addr "$ICS_FOREVER_ADDRESS" --arg pid "$CNIGHT_POLICY_ID" --arg name "$CNIGHT_ASSET_NAME" \
+   '{illiquid_circulation_supply_validator_address: $addr, asset: {policy_id: $pid, asset_name: $name}, utxos: [], total_amount: 0}' \
```
rsporny: utxos and total amount for both isc and reserve config should use the amount and tx hash from mint-cnight-supply that seeded it

local-environment/src/networks/local-env/configurations/midnight-setup/entrypoint.sh
```diff
@@ -27,142 +28,225 @@ check_json_validity() {
...
-export CHAINSPEC_C2M_BRIDGE_CONFIG=/tmp/c2m-bridge-config.json
+export CHAINSPEC_PC_CHAIN_CONFIG=/res/local/pc-chain-config.json
+export CHAINSPEC_CNIGHT_GENESIS=/res/local/cnight-config.json
+export CHAINSPEC_ICS_CONFIG=/res/local/ics-config.json
+export CHAINSPEC_RESERVE_CONFIG=/res/local/reserve-config.json
+export CHAINSPEC_FEDERATED_AUTHORITY_CONFIG=/res/local/federated-authority-config.json
+export CHAINSPEC_SYSTEM_PARAMETERS_CONFIG=/res/local/system-parameters-config.json
+export CHAINSPEC_PERMISSIONED_CANDIDATES_CONFIG=/res/local/permissioned-candidates-config.json
+export CHAINSPEC_REGISTERED_CANDIDATES_ADDRESSES=/res/local/registered-candidates-addresses.json
+export CHAINSPEC_C2M_BRIDGE_CONFIG=/res/local/c2m-bridge-config.json
```
rsporny: would it be possible to omit leading `/` and use the values from cfg/local.toml?

res/cfg/local.toml
```diff
@@ -0,0 +1,35 @@
...
+args = [
+    "--dev",
+    "--chain", "local",
+    "--node-key", "0000000000000000000000000000000000000000000000000000000000000001",
+    "--rpc-external",
+    "--rpc-cors=all",
+    "--prometheus-external",
+    "--state-pruning", "archive",
+    "--blocks-pruning", "archive"
+]
```
rsporny: is this necessary? local cfg preset is only meant to be run as local-env

res/src/networks/definitions.rs
```diff
@@ -105,6 +105,96 @@ impl MidnightNetwork for UndeployedNetwork {
...
+/// artifacts, and sources all config from `res/local/` rather than `res/dev/`.
+pub struct LocalNetwork;
+impl MidnightNetwork for LocalNetwork {
+	fn name(&self) -> &str {
+		"local1"
+	}
+
+	fn id(&self) -> &str {
+		"local"
+	}
```
rsporny: would be good to align with cfg.toml

res/local/ics-addresses.json
rsporny: removed addresses

local-environment/src/networks/local-env/configurations/mint-cnight-supply/mint-cnight-supply.sh
```diff
@@ -0,0 +1,286 @@
...
+
+if [ -f "$SEEDED_MARKER" ]; then
+  echo "cNIGHT already seeded ($SEEDED_MARKER present); skipping."
+  exit 0
+fi
+
+echo "=== cNIGHT genesis seeding ==="
+
+# Wait for the contract-compiler artifacts (deployed-contract info + compiled policy).
```
rsporny: no need to wait, it should be ready per docker compose condition

local-environment/src/networks/local-env/configurations/mint-cnight-supply/mint-cnight-supply.sh
```diff
@@ -0,0 +1,199 @@
...
+MIN_UTXO_LOVELACE=1500000
+
+# The bridge is configured to observe this cNIGHT minting policy on every non-mainnet
+# environment. We assert the compiled policy matches so a contract change can't
+# silently mint under a policy the bridge ignores.
+EXPECTED_POLICY_ID=d2dbff622e509dda256fedbd31ef6e9fd98ed49ad91d5c0e07f68af1
+
+if [ -f "$SEEDED_MARKER" ]; then
+  echo "cNIGHT already seeded ($SEEDED_MARKER present); skipping."
+  exit 0
```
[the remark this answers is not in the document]
rsporny: Already handled — `runtime-values` **is** cleared whenever the Cardano ledger is fresh, so the stale-marker skip can't occur in practice.

`cardano/entrypoint.sh` gates on `/shared/cardano.ready` (touched only at the end of a successful fresh init). `/shared` is the named `shared-volume` and the ledger is the named `cardano-node-1-data`; teardown uses `docker compose down --volumes` (`src/lib/docker.ts`), which removes them together. So:

**fresh Cardano ledger ⟺ fresh `shared-volume` ⟺ no `cardano.ready` ⟹ entrypoint runs `rm -rf /runtime-values/*`**

…and that wipe happens before the node reports healthy, while `mint-cnight-supply` is gated on `cardano-node-1: service_healthy` — so the `cnight-supply-minted` marker is gone before the mint script ever checks it. The graceful `stop` command also `rm`s `./runtime-values` alongside the volume teardown.

The failure mode described requires wiping `cardano-node-1-data` while preserving both `shared-volume` and `runtime-values` — i.e. deleting one named volume but not the other, which `down --volumes` never does. Not worth the on-chain re-check for that pathological manual-only case.

local-environment/src/networks/local-env/docker-compose.yml
```diff
@@ -173,10 +173,10 @@ services:
     command: >
       postgres
       -c maintenance_work_mem=256MB
+      -c max_connections=300
```
LGLO: This is worrisome.

node/src/chain_spec/mod.rs
```diff
@@ -393,7 +393,13 @@ fn genesis_config<T: MidnightNetwork>(genesis: T) -> Result<serde_json::Value, C
...
 						.subminimal_transfers_flush_threshold,
 				},
-				approved_txs: vec![],
+				approved_txs: bridge_config
+					.approved_txs
+					.iter()
+					.map(|s| {
+						McTxHash::decode_hex(s).expect("Failed to decode c2m approved_txs entry")
+					})
+					.collect(),
```
LGLO, line 402: TBH, I expected to have it empty here and `jq` in the chain spec, but this is okay as well.

res/local/c2m-bridge-config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "subminimal_transfers_flush_threshold": 500000,
+  "initial_data_checkpoint": "73dca41e6839a54fd0e3c08499bfb866367da621774ee4e5f3afdaa763b8006a",
+  "approved_txs": [
+    "5c24675a6a04480399f4cdd11f2ac100ebf6012cb3989aa63cf1e03832824b24"
```
LGLO, line 5: initial data checkpoint and this tx hash could be filled by midnight-setup. It seems brittle here.

local-environment/src/networks/local-env/docker-compose.yml
rsporny: reverting

node/src/cfg/mod.rs
```diff
@@ -302,7 +302,8 @@ impl SubstrateCli for Cfg {
 				};
 				chain_config(network)
 			},
-			"local" | "dev" => chain_config(UndeployedNetwork),
+			"local" => chain_config(LocalNetwork),
```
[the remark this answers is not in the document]
rsporny: Good catch — fixed by repointing the standalone dev instructions to `dev`.

`CHAIN=local` now selects `LocalNetwork` (unfunded genesis, bootstrapped by the dockerized local-env), so the README's single-/multi-node dev commands would no longer boot a funded chain. Changed those `CHAIN=local` / `--chain local` entries to `dev` (→ `UndeployedNetwork`, funded) — the config-parameter table, all six node commands, and the prose. Also corrected the "Available Networks" list: `dev` is the funded standalone dev network, `local` is the local-env network (unfunded, not for standalone use).

util/toolkit/src/tx_generator/source.rs
```diff
@@ -250,6 +256,26 @@ impl GetTxs for GetTxsFromFile {
...
+}
+
+#[async_trait]
+impl GetTxs for GetTxsChained {
+	async fn get_txs(
+		&self,
+	) -> Result<SourceTransactions, Box<dyn std::error::Error + Send + Sync>> {
+		let mut base = self.base.get_txs().await?;
+		let overlay = self.overlay.get_txs().await?;
+		base.blocks.extend(overlay.blocks);
```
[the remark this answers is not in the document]
rsporny: Verified and fixed in `31b303aa`.

Confirmed the mechanism: file-loaded blocks are stamped `number = 0` (`RawBlockData::new_from_timestamp`), so `GetTxsChained` appended 0-numbered overlay blocks after the base's ascending live blocks. That (a) violated the ascending-order precondition of the warm-cache replay filter `partition_point(|b| b.number <= start_height)`, so overlay blocks got skipped on a warm wallet cache, and (b) made the last overlay block look like the trailing dust-warp synthetic (`number == 0`) and get dropped from the replay set.

Fix (your second suggestion — assign replay-only heights before returning the chained source): new `SourceTransactions::chain_overlay` renumbers overlay blocks to continue ascending from the base head, splicing them *before* a trailing dust-warp synthetic when the base has one so it stays last. Covered by two unit tests.

Worth noting the current e2e suite never hit this — it uses an empty `ledger_state_db` + in-memory fetch cache, i.e. the non-cached `build_fork_aware_context_raw` path — so the bug was latent, exposed only when `--src-overlay-file` combines with `--src-url` under the default (on) wallet cache. Fixed regardless.

res/local/c2m-bridge-config.json
rsporny: midnight-setup is indeed patching it, but the difference now is that it patches them in place, so whenever the values change, you'll see that in your working tree. This way we can always keep them up-to-date and I hope it will help with genesis creation.

local-environment/src/networks/local-env/configurations/mint-cnight-supply/mint-cnight-supply.sh
```diff
@@ -0,0 +1,275 @@
...
+  echo "ERROR: faucet bridge tx $BRIDGE_TX_ID submitted but not confirmed within the budget"
+  exit 1
+fi
+# midnight-setup pre-approves this hash in the c2m-bridge genesis config.
+echo "$BRIDGE_TX_ID" > "$BRIDGE_TX_HASH_FILE"
+echo "=== faucet bridge transfer submitted (tx $BRIDGE_TX_ID) ==="
+
+# The marker doubles as the seeding tx hash midnight-setup anchors the bridge checkpoint to.
+# Written LAST so a partially-completed run (e.g. bridge transfer failed) re-runs whole.
+echo "$SEED_TX_ID" > "$SEEDED_MARKER"
```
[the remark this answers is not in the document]
rsporny: let's not add any guarantees that it will work after partially completed run.

res/genesis/genesis_block_local.mn
```diff
@@ -0,0 +1 @@
+{"batches":[[]]}
```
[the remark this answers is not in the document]
rsporny: Accepted — leaving this as-is.

For local's (intentionally empty) genesis block the timestamp does fall back to `res/mainnet/cardano-tip.json`, but it's only block 0's baseline and isn't load-bearing: blocks advance from wall-clock after genesis, the mainchain slot→epoch anchor (`mc__first_epoch_timestamp_millis`) is set per-network, and DUST/bridge timing is anchored to Cardano observation. Not worth a local-only timestamp source here.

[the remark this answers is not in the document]
rsporny: @LGLO @ozgb what do you think? Maybe we should add `res/local/cardano-tip.json`?

tests/e2e/tests/rpc_abuse.rs
```diff
@@ -1,197 +1,130 @@
...
-    tracing::info!("Submitting DEPLOY_TX (second attempt - should be rejected)...");
-    let second_result = client.submit_expecting_rejection(DEPLOY_TX.to_vec()).await;
+    // First submission must land on-chain (retries past transient shared-wallet
+    // DUST contention); returns the exact bytes that landed, to replay verbatim.
+    tracing::info!("Submitting DEPLOY_TX (first attempt, expect success)...");
+    let (deploy_tx, _addr) = crate::deploy_and_confirm(&client, &url).await;
 
+    // Replaying the identical deploy must be rejected — the contract now exists.
+    tracing::info!("Replaying the same DEPLOY_TX (expect rejection)...");
+    let result = client.submit_expecting_rejection(deploy_tx.clone()).await;
```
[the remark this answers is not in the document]
rsporny: Fixed in `7a6498fc` (option a). You're right — `submit_expecting_rejection` propagated a submission-level `submit_and_watch` error (`already imported` / `temporarily banned`) as `Err` via `?`, so a replay rejected at submission failed the `assert!(result.is_ok())`. And since `deploy_and_confirm` lands the original at `InBestBlock` (not finalized), the replay commonly hits that submission-level path rather than a watched `ContractAlreadyDeployed` — it passed CI only by timing luck.

Now it catches the `submit_and_watch` error and returns it as a valid rejection; the callers' `assert_invalid_transaction` still validates the message shape (the pool messages contain "Transaction", so they pass, while a genuine infra error wouldn't).

tests/e2e/tests/lib.rs
```diff
@@ -282,6 +283,337 @@ async fn run_warmup() {
...
+        untagged: false,
+        src_file: deploy_str.clone(),
+    })
+    .expect("derive contract address from deploy tx");
+    let contract_address =
+        contract_address_decode(&addr_str).expect("decode derived contract address");
+
+    let args = GenerateTxsArgs {
+        builder: Builder::ContractSimple(ContractCall::Call(ContractCallArgs {
+            funding_seed: DEV_WALLET_SEED.to_string(),
```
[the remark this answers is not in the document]
rsporny: Good catch — addressed in `0fee548`.

You're right about the mechanism: `0x..01` is initialized with a single NIGHT + single DUST utxo, and since the store overlays the (never-submitted) deploy, the store's coin selection can only pick the deploy's overlay-only change — so submitting it alone could be rejected for missing inputs rather than `ContractNotPresent`, and the lenient `assert_invalid_transaction` would pass either way.

Fix: the two ddos tests now assert the rejection is specifically `ContractNotPresent` (ledger `custom error: 114`) via a new `assert_contract_not_present`, so a wrong-path (missing-input, 195/196) rejection fails loudly instead of masquerading as success. In practice the tests already hit 114 (the last CI run logged `✓ rejected with: … custom error: 114`), so this locks that in rather than changing behavior.

I went with tightening the assertion over "independent funding" because local-env funds **only** `0x..01` — there's no second funded wallet to draw from, and the overlay is required for the call builder to see the contract at all. If you'd prefer to actually decouple the funding (fund a second wallet at bring-up), that's a larger change I can scope separately. (`replay_attack_rejected_via_rpc` keeps the lenient assert — its rejection is `ContractAlreadyDeployed`/pool-ban, a different code.)

README.md
```diff
@@ -269,7 +269,9 @@ These are built in CI. See the workflow files for the latest `earthly` commands:
 ### Start local network
 
 **Available Networks:**
-- `local` - Development network (default)
+- `dev` - Development network with a funded genesis; use this for a standalone node
```
LGLO: I don't understand what "(un)funded genesis" means. There are 24e15 stars in every genesis.

indexer
LGLO, line 1: Should this happen in this PR?

Earthfile
```diff
@@ -1965,7 +2028,7 @@ stop-local-env:
     ARG USERARCH
     WORKDIR local-environment
     RUN npm ci
-    RUN ARCHITECTURE=$USERARCH MIDNIGHT_NODE_IMAGE=any/any npm run stop:local-env
+    RUN ARCHITECTURE=linux/$USERARCH MIDNIGHT_RESERVE_CONTRACTS_PATH="$(cd .. && pwd)/midnight-reserve-contracts" MIDNIGHT_NODE_IMAGE=any/any npm run stop:local-env
```
LGLO, line 2013: Is it required for docker-compose to remove volume?

Earthfile
```diff
@@ -463,6 +486,14 @@ rebuild-genesis-state-undeployed:
         --NETWORK=undeployed \
         --GENERATE_TEST_TXS=true
 
+# rebuild-genesis-state-local rebuilds the genesis ledger state for the local network (local-environment).
+# The local network does not fund any faucet wallets at genesis - wallets are funded at runtime via the
+# cNIGHT->DUST bridge. This MUST be followed by updating the chainspecs for CI to pass!
```
LGLO: A) cNIGHT->mNIGHT protocol bridge.
B) We should not need updating chain-spec, because there is no chain spec committed for local env. Chain spec is built by midnight-setup

Earthfile
```diff
@@ -288,13 +298,26 @@ rebuild-genesis-state:
...
     ELSE IF [ "${FUND_FAUCET_WALLETS}" = "false" ]
         RUN echo "Generating genesis without faucet wallet funding (FUND_FAUCET_WALLETS=false)"
-        RUN /midnight-node-toolkit generate-genesis \
-            --network ${NETWORK} \
-            --ledger-parameters-config /genesis-config/ledger-parameters-config.json \
-            --cnight-generates-dust-config /genesis-config/cnight-config.json \
-            --ics-config /genesis-config/ics-config.json \
-            --reserve-config /genesis-config/reserve-config.json \
-            --cardano-tip-config /genesis-config/cardano-tip.json
+        # cardano-tip.json is only present for deployed networks (mainnet etc.) that anchor
```
LGLO: Effectively we need it only for mainnet, but we don't generate genesis for it

Earthfile
```diff
@@ -245,13 +245,23 @@ rebuild-genesis-state:
...
+    # Cardano from genesis (block 0), so there is no pre-existing tip. undeployed funds
+    # faucet wallets from the hardcoded seeds injected below; local ships an unfunded genesis
+    # and is funded over the cNIGHT bridge at runtime instead.
     RUN mkdir -p /genesis-config
     IF [ "${NETWORK}" = "undeployed" ]
         COPY res/dev/ledger-parameters-config.json /genesis-config/ledger-parameters-config.json
         COPY res/dev/cnight-config.json /genesis-config/cnight-config.json
         COPY res/dev/ics-config.json /genesis-config/ics-config.json
         COPY res/dev/reserve-config.json /genesis-config/reserve-config.json
+    ELSE IF [ "${NETWORK}" = "local" ]
```
LGLO: Cosmetic: can this be alternative condition for check in line 255?

util/toolkit/src/commands/root_call.rs
```diff
@@ -330,9 +331,15 @@ async fn execute_governance_call(
 	Ok(())
 }
 
-async fn vote_on_proposal(
+/// Cast the first two committee votes on `proposal`. The votes come from
```
LGLO: Is it designed for two votes only?

util/toolkit/src/commands/root_call.rs
```diff
@@ -348,12 +355,19 @@ async fn vote_on_proposal(
...
-	api.tx()
-		.await?
-		.sign_and_submit_then_watch_default(&vote_call, signer)
-		.await?
-		.wait_for_finalized_success()
-		.await?;
+	// Submit all votes first (each returns once it's in the tx pool), so they
+	// land together in the next block(s)...
+	let mut submitted = Vec::new();
+	for (i, voter) in voters.iter().take(2).enumerate() {
```
LGLO: I can't understand why hardcoding `2`

local-environment/src/networks/local-env/configurations/mint-cnight-supply/mint-cnight-supply.sh
```diff
@@ -0,0 +1,257 @@
...
+echo "ICS Forever address:     $ICS_ADDR"
+echo "Reserve Forever address: $RESERVE_ADDR"
+
+# Wrap the compiled infinite-mint cNIGHT policy into a .plutus text envelope for
+# cardano-cli's --mint-script-file / policyid.
+jq '{type: "PlutusScriptV3", description: "", cborHex: (.validators[] | select(.title == "test_cnight_no_audit.tcnight_mint_infinite.else") | .compiledCode)}' \
+  "$PLUTUS_JSON" > "$CNIGHT_PLUTUS"
+
+POLICY_ID=$(cardano-cli latest transaction policyid --script-file "$CNIGHT_PLUTUS")
+echo "cNIGHT policy id: $POLICY_ID"
```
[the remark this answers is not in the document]
rsporny: Fixed in `13bfe7b9` + `98fd0545`. Setup was reading the policy id from the committed config while the mint used the freshly-compiled one, so a contract change would drift them. Both now derive it from the compiled blueprint, and setup patches the config in place before build-spec — bridge always watches what was minted.

util/toolkit/src/tx_generator/source.rs
```diff
@@ -250,6 +256,29 @@ impl GetTxs for GetTxsFromFile {
...
+impl GetTxs for GetTxsChained {
+	async fn get_txs(
+		&self,
+	) -> Result<SourceTransactions, Box<dyn std::error::Error + Send + Sync>> {
+		let mut base = self.base.get_txs().await?;
+		let overlay = self.overlay.get_txs().await?;
+		// Renumber overlay blocks to continue ascending from the base head — file-loaded
+		// blocks are all `number = 0`, which otherwise breaks the warm-cache replay height
+		// filter and looks like the trailing dust-warp synthetic. See `chain_overlay`.
+		base.chain_overlay(overlay.blocks);
```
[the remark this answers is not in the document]
rsporny: Fixed in `f662747`. An overlaid source now reports no chain id, so it takes the cache-free path — off-chain overlay state can't be written to or read from the live-chain cache. Complements the earlier renumbering.

res/genesis/genesis_block_local.mn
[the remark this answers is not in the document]
ozgb: This should be fine - as long as we have a timestamp in genesis, indexer/toolkit is happy. And it's usual for genesis timestamp to be some point in the past, first block to be more recent.

res/local/c2m-bridge-config.json
ozgb: Do these values change each time a new local-env starts? If so, it would be better to have these values empty

Perhaps patched configs could output to some `.patched` dir that's git-ignored, and `local.cfg` can read the configs from the `.patched` dir?

I understand the convenience of running `git diff` to see the changes, but if they're re-generated each time local-env runs, they shouldn't be checked-in to the repo

res/src/networks/definitions.rs
```diff
@@ -105,6 +105,97 @@ impl MidnightNetwork for UndeployedNetwork {
 		super::MainChainScripts::load_from_configs(&registered_candidates, &permissioned_candidates)
 	}
 }
+
+/// The `local` network backs the dockerized local-environment. Unlike
+/// `UndeployedNetwork` it runs against a real (local, dockerized) Cardano main
+/// chain — no follower mock. It has its own ledger identity (`local`, addresses
+/// `mn_addr_local1…`), its own genesis artifacts, and sources all config from
+/// `res/local/` rather than `res/dev/`.
+pub struct LocalNetwork;
```
ozgb: I think we should remove this - local environment is a dynamic config, the correct genesis config values are not known at compile-time

The `CustomNetwork` config is designed for that use-case:
```rust
/// Used when `--chain` is not specified when running `build-spec` - it will source chain values from
/// environment variables at runtime rather than hard-coded values at compile-time
pub struct CustomNetwork {
```

tests/e2e/src/api/midnight.rs
```diff
@@ -1088,12 +1088,21 @@ impl MidnightClient {
         tx_bytes: Vec<u8>,
     ) -> Result<String, Box<dyn std::error::Error>> {
```
ozgb, line 1089: With a custom error type, you could fold these two methods into one (`submit_midnight_tx` and `submit_expecting_rejection`)

Error type could be:

```rust
#[derive(thiserror::Error)]
enum SubmitError {
    #[error("error at pre-dispatch")]
    PreDispatchError,
    #[error("subxt error: {0}")]
    SubxtError(#[from] subxt::Error)
    ...
}
```

Then you could match on the Result type at the call site

not necessary for this PR

LGLO (dismissed):

Earthfile
```diff
@@ -288,13 +298,26 @@ rebuild-genesis-state:
...
-            --ledger-parameters-config /genesis-config/ledger-parameters-config.json \
-            --cnight-generates-dust-config /genesis-config/cnight-config.json \
-            --ics-config /genesis-config/ics-config.json \
-            --reserve-config /genesis-config/reserve-config.json \
-            --cardano-tip-config /genesis-config/cardano-tip.json
+        # cardano-tip.json is only present for deployed networks (mainnet etc.) that anchor
+        # genesis to a live Cardano tip. Networks spawned from their own genesis omit it —
+        # undeployed (mocked main chain) and local (its own dockerized Cardano) — so include
+        # the flag only when the file exists.
+        RUN if [ -f /genesis-config/cardano-tip.json ]; then \
```
ozgb: Rather than a big `if` branch, would be neater to construct a `$CARDANO_TIP_ARG` dynamically:

```dockerfile
RUN CARDANO_TIP_ARG=""; \
    [ -f /genesis-config/cardano-tip.json ] && CARDANO_TIP_ARG="--cardano-tip-config /genesis-config/cardano-tip.json"; \
    /midnight-node-toolkit generate-genesis \
        --network ${NETWORK} \
        --ledger-parameters-config /genesis-config/ledger-parameters-config.json \
        --cnight-generates-dust-config /genesis-config/cnight-config.json \
        --ics-config /genesis-config/ics-config.json \
        --reserve-config /genesis-config/reserve-config.json \
        ${CARDANO_TIP_ARG}
```

util/toolkit/src/tx_generator/source.rs
```diff
@@ -111,6 +111,12 @@ pub struct Source {
 	/// Load input transactions/blocks from file(s). Used as initial state for transaction generator.
 	#[arg(long = "src-file", value_delimiter = ' ', conflicts_with = "src_url", global = true)]
 	pub src_files: Option<Vec<String>>,
+	/// Overlay transactions/blocks from file(s) ON TOP of the main source. Unlike
+	/// `--src-file` this composes with `--src-url`: the generator state becomes the live
+	/// chain plus these (typically not-yet-submitted) transactions — e.g. building a
+	/// contract call against a deploy that intentionally never goes on-chain.
+	#[arg(long = "src-overlay-file", value_delimiter = ' ', global = true)]
+	pub overlay_files: Option<Vec<String>>,
```
ozgb: I'm a little wary of the `overlay_files` option

The way it works - transactions included in any overlay files are appended to the real chain in a new block by the toolkit. The block context is set to the current timestamp

I think we'll run into "invalid dust proof" errors here - the Dust state will not match the state on-chain, resulting in failed txs if they're constructed from this hypothetical chain state

@tkerber is out on PTO at the moment - some investigation is needed here.

Is there some way you can get your desired test-case without this feature? in general, appending hypothetical new blocks likely won't work due to the dust system as far as I understand

indexer
rsporny: doesn't hurt, current one is not compatible with rc.3

res/local/c2m-bridge-config.json
rsporny: no, they don't. That's why I want to keep them patched in place. The only way they can change is if we change the cardano genesis or modify something in entrypoints transactions. Current state is deterministic so these file doesn't change.

README.md
rsporny: Good point — reworded in `4b5497ec`. It's about spendable wallets, not the supply: `dev` funds dev wallets at genesis, `local` funds none (bridge-funded at runtime).

Earthfile
rsporny: Done in `f93a14a` — dropped the `local` branch. It reads `res/local` like any non-undeployed network, so it now falls through to `ELSE`, with the cardano-tip copied via `--if-exists` (genesis-spawned nets have none).

Earthfile
rsporny: Done in `f93a14a` — no more duplicated command. Added the tip flag via command substitution (matching the existing `$(if …; then echo …; fi)` pattern already used elsewhere in the file), and dropped the separate `local` config-copy branch too. Pure refactor — genesis unchanged.

Earthfile
rsporny: Not required to remove volumes — `down` drops them regardless. It's set only for parity with `up`, so compose doesn't warn about the unset var when it parses the file. Keeping it.

Earthfile
rsporny: Fixed in `731c93a3` — corrected to cNIGHT→mNIGHT, and dropped the chainspec line since local has no committed chain spec (midnight-setup builds it at runtime).

Earthfile
rsporny: true

util/toolkit/src/commands/root_call.rs
rsporny: Backed this out of the PR (`2b5d222`) — the batching had no consumer here (the bring-up funds via a pre-approved genesis tx, no governance round), so `root_call` is unchanged. You're right that it assumes a 2/3 threshold hardcoded to 2 (only correct for 3 members) — that's pre-existing, and I'll fix it properly (derive from member count + min-keys guard + test) in a dedicated toolkit PR.

util/toolkit/src/commands/root_call.rs
rsporny: reverting

res/src/networks/definitions.rs
rsporny: removed

tests/e2e/src/api/midnight.rs
rsporny: Good idea — a typed `SubmitError` would let these collapse into one. Agreed it's out of scope here; filed #1840 as a follow-up.

util/toolkit/src/tx_generator/source.rs
rsporny: Removed the feature entirely (`7c0a3c3f`). You're right that hypothetical blocks + the dust system are a bad mix — and pre_dispatch validation is already covered cleanly at the `pallet-midnight` unit level, so the e2e ddos/replay tests and `--src-overlay-file` are gone. Kept only genuine e2e (real deploys).

LGLO (dismissed):

ozgb (dismissed):

tests/e2e/tests/operational.rs
```diff
@@ -1,37 +1,123 @@
...
+/// funded by the bridge-funded dev wallet (0x..01), is accepted at the RPC and
+/// included in a block.
 #[e2e_test]
-#[ignore = "Requires fresh node state - run manually with cargo test-e2e-local"]
 async fn valid_deploy_transaction_succeeds_via_rpc() {
-    use midnight_node_res::undeployed::transactions::DEPLOY_TX;
+    use midnight_node_e2e::api::midnight::MidnightClient;
+
+    // Funded + DUST-registered at runtime by init-mnight-faucet; wait until ready.
+    crate::ensure_dev_wallet_funded().await;
```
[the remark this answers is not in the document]
rsporny: Non-blocking. The qanet e2e job runs filtered (`-- cnight::observation::`), so it never runs these tests — and an *unfiltered* `--features qanet` run was already local-only before this PR (`contract_state`, `governance`, non-observation `cnight` are all ungated and need the local stack). This PR just adds two more to that set. Rather than gate only these two (which leaves the same hole elsewhere), I filed #1842 to make the per-feature test sets coherent — gate all local-only tests, leaving `cnight::observation::` as the shared set.

LGLO (approved):

PR #498, opened 2026-01-20
Feat: impl `CandidatesDataSourceImpl` locally

NachoPal:
# Overview
The goal is to remove PC `d_parameter_policy_id` dependency.

Closes: https://shielded.atlassian.net/browse/PM-21395

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [ ] Changes are backward-compatible (or flagged if breaking)
- [ ] Pull request description explains why the change is needed
- [ ] Self-reviewed the diff
- [ ] I have included a change file, or skipped for this reason: 
- [ ] If the changes introduce a new feature, I have bumped the node minor version
- [ ] Update documentation (if relevant)
- [ ] Updated AGENTS.md if build commands, architecture, or workflows changed
- [ ] No new todos introduced

## 🧪 Testing Evidence

Please describe any additional testing aside from CI:

- [ ] Additional tests are provided (if possible)

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A

PR #1817, opened 2026-07-03
fix(cnight-observation): complete CIP-19 validation for Cardano reward addresses (#498)

m2ux:
## Summary

Consolidate Cardano reward-address (CIP-19) validation into a single validated constructor for `CardanoRewardAddressBytes`, and route the follower, genesis config, and the mapping-address setter through it so malformed, wrong-length, wrong-type, or wrong-network reward addresses are rejected with specific errors.

🐛 [Issue](https://github.com/shieldedtech/shielded-security-engineering/issues/498)  📐 [Engineering](https://github.com/shieldedtech/midnight-agent-eng/blob/mike/.engineering/artifacts/planning/2026-07-03-midnight-node-issue-498-cip19-cardano-reward-address-validation/README.md)

_Jira: [PM-20267](https://input-output.atlassian.net/browse/PM-20267)_

---

## Motivation

Cardano reward addresses supplied to the cNIGHT observation path are currently validated only partially, and in several independent places. Malformed, wrong-length, wrong-address-type, or wrong-network reward addresses can pass through the follower, the genesis config, and `set_mapping_validator_contract_address`, and the follower carries roughly three ad-hoc parsing spots each with its own error enum. An address that is not a valid CIP-19 reward address for the expected network can therefore be accepted and stored, misdirecting or silently dropping rewards and making failures hard to diagnose.

This change gives every entry point one strict, shared check that rejects each bad case with a specific, named error, replacing the scattered ad-hoc parsing with a single validated constructor and error type.

---

## Changes

- **Primitive** — adds a validated constructor `CardanoRewardAddressBytes::try_new(bytes, expected_network)` and a dedicated error type with a distinct variant per rejection case (wrong length, wrong address type, wrong network), checking length first then the CIP-19 header nibbles via `no_std` bit arithmetic; the length-only conversion is retained for test and benchmark use.
- **Follower** — the four reward-address sites (registration, deregistration, asset-create, asset-spend) share one infallible constructor that builds the reward address from the stake credential and the configured network. There is no follower-side validation or rejection, because a reward address the node constructs from an already-valid credential cannot be malformed; the now-dead reward-address error variant is removed.
- **Genesis** — validates deserialized reward-address keys against address type and expected network during the genesis build, where the network is available, rather than the length-only check used previously.
- **Mapping validator** — gives `set_mapping_validator_contract_address` and the genesis validator-address build a sibling validated path returning a specific error, kept separate from the reward-address constructor because the validator address is a different Cardano address category.
- **Tests** — cover each rejection case (wrong length, wrong type, wrong network, malformed header) plus valid type-14 and type-15 round-trips, and repair the existing genesis/extrinsic tests for the new validation.
- **On-chain rejection event** — not added. The follower filters invalid reward addresses before the inherent, so the pallet never observes a rejected registration; the per-rejection follower log provides the intended observability without a non-trivial inherent-payload change.

---

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] I have included a change file, or skipped for this reason: [reason]
- [x] No new todos introduced

---

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [x] Node Client Update
- [ ] Other
- [ ] N/A

---

## 🗹 TODO before merging

- [ ] Run and pass `cargo check` / `cargo clippy` / `cargo nextest` (deferred — see Validation status)
- [ ] Ready for review

[PM-20267]: https://shielded.atlassian.net/browse/PM-20267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


primitives/mainchain-follower/src/data_source/cnight_observation.rs
```diff
@@ -272,13 +291,21 @@ impl MidnightCNightObservationDataSourceImpl {
 			};
 
 			let reward_address = RewardAddress::new(cardano_network, &credential);
-			// Unwrap here is OK - we know the reward_address is always 29 bytes
-			let cardano_address = reward_address.to_address().to_bytes().try_into().unwrap();
+			let cardano_reward_address = match decode_reward_address(
```
LGLO: Minor: creating `RewardAddress` can be hidden in `decode_reward_address`. Make `decode_reward_address` accept `Credential`

primitives/mainchain-follower/src/data_source/cnight_observation.rs
```diff
@@ -272,13 +291,21 @@ impl MidnightCNightObservationDataSourceImpl {
...
+				reward_address.to_address().to_bytes(),
+				cardano_network,
+			) {
+				Ok(addr) => addr,
+				Err(e) => {
+					log::error!("Rejected deregistration reward address: {e} ({header:?})");
```
LGLO: I think that unwrap was correct. Perhaps hidden in renamed `decode_reward_address`.
We have two correct inputs at line 293. It is insane to reject at this point, because on chain data was correct and our algorithm is invalid.

primitives/mainchain-follower/src/data_source/cnight_observation.rs
m2ux: The reward-address helper now takes the stake credential and the network and builds the reward address internally, so the construction is no longer duplicated across the four follower sites. Done in commit 77b5895.

*Posted by Claude (Opus 4.8) on behalf of Mike Clay.*

primitives/mainchain-follower/src/data_source/cnight_observation.rs
m2ux: The reward address is built from an already-valid credential and our own configured network, so it is always a valid 29-byte address; validating and rejecting it here was wrong — a failure could only mean our own construction is broken, and skipping it would silently drop a legitimate registration. The helper is now infallible and keeps the original expect/unwrap semantics, hidden behind the renamed constructor. The genuine CIP-19 validation stays at the genesis entry point, where addresses actually arrive from outside. Done in commit 77b5895.

*Posted by Claude (Opus 4.8) on behalf of Mike Clay.*

PR #1849, opened 2026-07-08
feat(1474): expose ledger-emitted events from the node

### LINKED ISSUE #1474: Node should expose events emitted by the Ledger
### Context & versions

The node discards all events emitted by the ledger when processing a transaction:
https://github.com/midnightntwrk/midnight-node/blob/c06fbbda048744f065b79f0e3187d90cb0c2a088/ledger/src/versions/common/api/ledger.rs#L149-L150

This means the indexer has to re-apply transactions locally to re-create the discarded events:
https://github.com/midnightntwrk/midnight-indexer/blob/6856f194ef6622412c134d614e63355d76e2cc9d/indexer-common/src/domain/ledger/ledger_state.rs#L253-L257

This means all indexer implementations must re-compute results already computed by the node, creating duplicated work across the network.

**Architecture:** Worth noting - if we did publish all events, this would increase network traffic. We should assess whether we want to complete this ticket, or if we're happy with the current architecture.

### Steps to reproduce

Normal operation, no ledger events are emitted by the node

### Expected behavior

All ledger events should be emitted by the node

### Actual behavior

No ledger events are emitted by the node

m2ux:
## Summary

Expose the ledger's per-transaction events as additive Substrate runtime events so the indexer and other consumers read them from the node instead of re-executing every transaction to reconstruct them.

🐛 [Issue](https://github.com/midnightntwrk/midnight-node/issues/1474)  📐 [Engineering](https://github.com/shieldedtech/midnight-agent-eng/blob/mike/.engineering/artifacts/planning/2026-07-08-node-expose-ledger-events/README.md)

---

## Motivation

When the node applies a transaction the ledger emits a stream of events describing what happened, and the node immediately discards it. Every consumer that needs those events — the indexer above all — must re-run each transaction locally to reconstruct them, duplicating compute on every operator's machine and slowing time-to-visibility for wallets and explorers.

---

## Changes

This change makes the node expose the events that the ledger already produces when it applies a transaction. The node now deposits those events as ordinary runtime events, carrying each one as the ledger's own self-describing tagged bytes, so consumers can read them straight from chain state. Both ordinary user transactions and system transactions emit their events, and the system-transaction event now carries the transaction hash so a consumer can line it up with the matching ledger events. Because deposited events live in per-block event storage that is cleared every block rather than in the gossiped block body, publishing them adds no ongoing traffic between peers.

Exposing a new event surface changes the runtime's public shape, so the work is built to stay safe across the window where a network is part-way through an upgrade. The ledger's host functions are versioned: the previous return shapes are preserved unchanged and the events travel on a new version alongside them, so an older runtime on a newer node still reads the old shape cleanly, while a newer runtime on an older node fails immediately and visibly rather than silently misreading data. The upgrade tooling now refuses to activate a new runtime until every validator is confirmed to be on a capable binary, enforcing a binaries-first rollout, and the release pipeline refuses to publish a changed runtime under an unchanged version number. The runtime version is raised to 2.2.0 to give the new surface a distinct identity, and the published metadata is regenerated so off-chain decoders can see and decode the new events.

The work also hardens two areas the new event stream interacts with. The cross-chain bridge previously cleared its pending accounting records as soon as it attempted a system transaction, even on failure, which could permanently lose a transfer's bookkeeping; those records are now cleared only when the transaction actually succeeds, so a failure leaves the transfer recoverable while still guaranteeing each one is processed at most once. And because emitting events is real work, both paths that deposit them now reserve block weight for each event so a large burst cannot quietly exceed a block's budget, with the accompanying benchmark anchored to the true per-block data ceiling rather than a small fraction of it.

---

## 🤖 AI Assistance

- **Assistant / Model:** Claude Code / claude-opus-4-8 (1M context)
- **Context scope:** mixed
- **Prompt classes:** code-generation, test-writing, refactoring, docs
- **Provenance log:** [08-provenance-log.md](https://github.com/shieldedtech/midnight-agent-eng/blob/mike/.engineering/artifacts/planning/2026-07-08-node-expose-ledger-events/08-provenance-log.md)

---

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] I have included a change file, or skipped for this reason: `changes/runtime/added/expose-ledger-events.md`
- [x] No new todos introduced

---

## 🔱 Fork Strategy

- [x] Node Runtime Update
- [ ] Node Client Update
- [ ] Other
- [ ] N/A

---

## 🗹 TODO before merging

- [x] Rebuild the runtime metadata for the new event variants
- [x] Sync the branch with `main`
- [x] Bump the node minor version for this new feature
- [x] Run clippy (workspace clippy with `-D warnings` is green; fmt is clean)
- [x] Run the full `cargo test` suite (green in CI)
- [ ] Run the re-anchored 50 MB `bench_block_full_of_events` benchmark and fold the measured per-event cost into the placeholder weights
- [ ] Add the bridge failure-branch test for the consume-on-success fix (the mock executor currently always succeeds); remaining partial-success / integration cases are in the [test plan](https://github.com/shieldedtech/midnight-agent-eng/blob/mike/.engineering/artifacts/planning/2026-07-08-node-expose-ledger-events/06-test-plan.md)
- [ ] Confirm with stakeholders that the indexer is the only service that needs to read these events directly from the node ([asked on the issue thread](https://github.com/midnightntwrk/midnight-node/issues/1474#issuecomment-4914140099))
- [ ] Ready for review


pallets/midnight-system/src/lib.rs
```diff
@@ -130,16 +131,20 @@ pub mod pallet {
...
+				<T as Config>::LedgerStateProviderMut::mut_ledger_state(|state_key| {
+					let result = LedgerApi::apply_system_transaction(
+						&state_key,
+						&midnight_system_tx.clone(),
+						block_context,
+						runtime_version,
+					)
+					.map_err(Error::<T>::from)?;
+					Ok::<(Vec<u8>, (Hash, Vec<LedgerEvent>)), Error<T>>((
+						result.state_root,
```
m2ux, line 164: Fixed in commit 0ea9a909: the deposited system-transaction event now carries the ledger transaction hash rather than the state root, on both the governance path and the internal executor, so it correlates with its ledger events by hash. Fixing this also surfaced and corrected a latent state-key writeback issue on the same path.

*Posted by Claude Code (Opus 4.8) on behalf of Mike Clay.*

ledger/src/common/types.rs
```diff
@@ -53,13 +68,15 @@ pub struct TransactionAppliedStateRoot {
 	pub claim_rewards: Vec<u128>,
 	pub unshielded_utxos_created: Vec<UtxoInfo>,
 	pub unshielded_utxos_spent: Vec<UtxoInfo>,
+	pub events: Vec<LedgerEvent>,
```
m2ux: Fixed in commit 0ea9a909: the apply-return structs are reverted to their events-free shape, byte-identical to main, and the events now ride new versioned WithEvents host functions. There is no negotiate-down, so a new runtime on an old binary fails at WASM instantiation on the missing import rather than decoding silently; the pre-activation validator-binary gate keeps the mixed-version window safe by enforcing a binaries-first rollout. An old-version events-free encoding test was added.

*Posted by Claude Code (Opus 4.8) on behalf of Mike Clay.*

pallets/midnight/src/benchmarking.rs
```diff
@@ -17,14 +17,67 @@ use super::*;
...
+use midnight_node_ledger::types::{LedgerEvent, LedgerEventSource};
+
+/// Per-event payload size for the event-emission benchmark, in bytes.
+///
+/// Sized to a deploy-heavy `ContractDeploy`/`ContractLog` event (~4 KiB of
+/// tagged `content_tagged_bytes`).
+const BENCH_EVENT_PAYLOAD_BYTES: usize = 4 * 1024;
+
+/// Upper bound on the number of ledger events deposited in a single block for
+/// the worst-case benchmark. `MAX_BENCH_EVENTS * BENCH_EVENT_PAYLOAD_BYTES`
```
m2ux: Fixed in commit 0ea9a909: the benchmark is re-anchored to the real 50 MB churn ceiling instead of the earlier roughly 1 MiB bound. The benchmark run that folds the measured per-event cost into the placeholder weights is still pending and is user-run, so no result is claimed yet.

*Posted by Claude Code (Opus 4.8) on behalf of Mike Clay.*

changes/runtime/added/expose-ledger-events.md
m2ux, line 1: Fixed in commit 0ea9a909: consumption of the pending accounting state is now conditional on execution success, so a failed system transaction preserves it for a genuine retry while single-use protection still holds on success. A dedicated failure-branch test is still pending because the current mock executor always succeeds and cannot exercise that path yet.

*Posted by Claude Code (Opus 4.8) on behalf of Mike Clay.*

m2ux, line 1: Fixed in commit 0ea9a909: a pre-activation gate now refuses to authorize the runtime upgrade when a connected node's active spec version is below the required minimum, with an explicit opt-out for deliberate lagging-binary rollouts.

*Posted by Claude Code (Opus 4.8) on behalf of Mike Clay.*

m2ux, line 1: Fixed in commit 0ea9a909: the advertised include and exclude filters are now applied to the discovered service list before services are recreated, so the flags take effect.

*Posted by Claude Code (Opus 4.8) on behalf of Mike Clay.*

m2ux, line 1: Fixed in commit 0ea9a909: the runtime spec version is bumped, now 2.2.0 after the sync with main, and a release-workflow guard now fails when the runtime source changed since the last runtime tag but the spec version did not increase.

*Posted by Claude Code (Opus 4.8) on behalf of Mike Clay.*

m2ux, line 1: Fixed in commit 0ea9a909: both deposit paths now carry a per-event weight term, declared as a pre-dispatch worst-case allowance and refined to the actual event count post-dispatch. The per-event figures are conservative placeholders pending the benchmark run, which is user-run.

*Posted by Claude Code (Opus 4.8) on behalf of Mike Clay.*

ledger/src/host_api/ledger_7.rs
```diff
@@ -179,6 +190,36 @@ pub trait LedgerBridge {
 				runtime_version,
 			)
 		}
+		.map(Into::into)
+	}
+
+	#[version(3)]
+	fn apply_transaction(
```
justinfrevert: Do we have confirmation that old ledger versions have events to expose?

m2ux: Yes — confirmed against the ledger crate sources, not just the interface. Versions 7 (midnight-ledger 7.0.3) and 8 (8.1.0) both populate the event stream during transaction apply: Zswap inputs and outputs, contract logs, parameter changes, and dust operations are all pushed and returned on the success and partial-success paths, the same as version 9 (which builds the same events via deferred closures materialized at apply time). So exposing events from the older versions is not exposing empty streams. The only cross-version difference is the internal event-tag version inside the ledger crate, which the node carries as opaque bytes and does not branch on.

*Posted by Claude Code (Opus 4.8) on behalf of Mike Clay.*

gilescope:
Prototyped on top of this branch locally (nothing pushed yet). Three things worth passing on.

**1. `EventDetails` derives `Storable`, not `Serializable` — so `content_tagged_bytes` isn't a flat record.**

The `Storable` derive generates `Serializable` as `Sp::new(self).serialize()` (`storage-macros/src/lib.rs:144`), and `Sp`'s impl is `serialize_to_node_list()` (`storage-core/src/arena.rs:1954`) — a topologically-sorted arena DAG. It's self-contained (deserialising allocates into the local `default_storage()`), but graph-shaped rather than struct-shaped.

Consequence for `docs/ledger-events.md`: telling consumers to decode with `tagged_deserialize::<EventDetails>` is correct, but it requires linking the ledger's *storage* crate, not just a SCALE codec. Worth stating outright — it also rules out mirroring the payload as a flat runtime type and re-decoding it there, which was the first thing I tried.

**2. Typed events instead of opaque bytes.**

Mapped `EventDetails` host-side in `versions/common/` — one impl covers ledger 8 and 9, the variants and field names are identical — into a flat `TypeInfo` enum. PJS then decodes straight from metadata: no type bundle, no decoder RPC.

Big payloads are referenced rather than embedded: `ContractDeploy` carries `address` + a 32-byte content key (`ArenaKey::hash()` — note `as_typed_key()` returns `Direct(node)` for a freshly-allocated node and inlines the whole subtree) instead of the full `ContractState`; likewise `ParamChange`. `ContractLog.logged_item` stays bytes, since only the emitting contract's ABI can interpret it.

Also fanned out to one pallet event variant per kind, so an explorer's list reads `midnight.DustSpendProcessed` rather than a run of identical `midnight.LedgerEvent` rows you have to expand one at a time.

Glad to raise this as a follow-up stacked on yours, or just hand you the diff — your PR, your call.

**3. Merging `main` has two breaks that git resolves cleanly and wrongly.**

- `ledger_8.rs` ends up with two `#[version(2)] fn apply_transaction`: `main` bumped it for tblock skewing (#1924 / #2031), this branch bumped it for the events return. Needs sequencing, or a separate fn name.
- `main` added a 7th `skew_tblock` parameter to `Bridge::apply_transaction`. Git applies it to the v1 call sites it already knew about and silently leaves this branch's events variants calling with six args.

Both compile-fail rather than misbehave, but they're easy to "fix" in the wrong direction while resolving.

PR #1865, opened 2026-07-13
feat(runtime): add pallet-babe

LGLO:
# Overview

Add pallet babe to the runtime. Consensus is still AURA. It is adding pallet and API only, to enable work on pallet consensus engine.

Epoch length is taken from pallet-sidechain.

There is no migration to set data of the babe pallet, because it will be done pallet-consensus-engine in the future PRs.
There is no runtime upgrade spec version increase because we update it rarely and closer to releases.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A


local-environment/src/networks/local-env/configurations/contract-compiler/Dockerfile
```diff
@@ -29,11 +29,10 @@ FROM node:22-slim@sha256:7378f5a4830ef48eb36d1abf4ef398391db562b5c41a0bded83192f
 # renovate: datasource=debian depName=unzip versioning=deb
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-        jq=1.6-2.1+deb12u2 \
-        curl=7.88.1-10+deb12u14 \
-        ca-certificates=20230311+deb12u1 \
-        vim=2:9.0.1378-2+deb12u2 \
-        unzip=6.0-28 && \
+        jq \
```
LGLO, line 32: These versions tend to disappear. I prefer avoid them in local image to have less failed builds

Klapeyron (dismissed):
we keep it inactive in a sense, that OneSessionHandler from pallet_babe will be not active, due Babe key is not a part of the session keys. Having that, pallet will be not providing consensus digest on session rotation

Klapeyron (approved):

runtime/src/lib.rs
LGLO, line 1: It will be increased in a PR closer to completing the epic. We don't need this one being updated separately.

LGLO:
Tested runtime upgrade with increased `spec_version` (I do not update it in PR because it will be later, together with other changes) and runtime upgrade succeeded. The chain was still producing blocks. It is true that `currentEpoch()` and `nextEpoch()` runtime calls panic, but these are not used. These need to be set later when enabling BABE consensus.

PR #1872, opened 2026-07-14
experiment: batch verification

ozgb:
# Overview

**Latest test images (31/07/29):** https://github.com/midnightntwrk/midnight-node/actions/runs/30634626825

**Experimental / draft.** Batch ZK-proof verification for Midnight transactions, so the expensive aggregate crypto runs once per block/batch instead of once per transaction. Everything is gated behind node config and **defaults OFF**, so with the flags unset the node behaves as today.

Two ingress points feed a process-global proof cache that downstream `get_verified_transaction` reads to skip (defer) the ZK crypto:

- **Block import** (`batch_verify_block_import`): received blocks are batch-verified up front (fail-fast; a bad proof rejects the block). *(landed earlier on this branch)*
- **Mempool** (`batch_verify_mempool`): external submissions are batched natively via a custom `MidnightChainApi` + bounded queue + blocking worker pool.

This PR adds the two pieces that make the mempool path real and safe to enable:

- **Isolation fallback (ledger):** when an aggregate batch fails, each transaction is re-verified individually to isolate the offender(s) — caching `false` for bad proofs and warming caches for good ones — instead of panicking. Extracts a shared `warm_verified_tx` helper.
- **Mempool ingress (node):** `MidnightChainApi` wraps the stock `FullChainApi`, delegating everything except `validate_transaction`; external Midnight `send_mn_transaction` extrinsics are queued and drained in batches by N blocking workers, which warm the caches and build the same validity tag the runtime would (`with_tag_prefix("Midnight").longevity(600).and_provides(tx_hash)`). Invalid/unavailable cases delegate to the runtime for the authoritative result.

The transaction pool is now always the SingleState `BasicPool<MidnightChainApi>` (batch verification requires the single-state ChainApi seam). Pool limits are reconstructed from the CLI (`--pool-limit`/`--pool-kbytes`/`--tx-ban-seconds`); `--pool-type` is intentionally not applied.

## ⚙️ Configuration (new `MidnightCfg` fields, for perf testing)

All new fields live in `node/src/cfg/midnight_cfg/mod.rs`, default OFF/unset, and are optional in `res/cfg` presets — existing presets need no changes to keep today's behavior.

| Field | Type | Default | Meaning |
|---|---|---|---|
| `batch_verify_block_import` | bool | `false` | Enable batch ZK-proof verification for received blocks at block import. Fail-fast: a bad proof anywhere in the batch rejects the block (no fallback). |
| `batch_verify_mempool` | bool | `false` | Enable batch verification on the mempool ingress path (external `send_mn_transaction` submissions). A bad proof is isolated per-tx, not panicked. |
| `batch_verify_max_batch_size` | usize | `64` | **M** — max transactions verified together in one aggregate crypto call. Upper bound on amortization; also an upper bound on per-batch latency, since every tx in a batch waits for the whole aggregate call to finish. |
| `batch_verify_target_batch_size` | usize | `16` | **k_target** — queue depth (mempool path) that triggers an immediate dispatch, without waiting for the age timeout. Lower = lower latency but smaller/less-amortized batches; higher = better amortization but more latency under light load. |
| `batch_verify_max_age_ms` | u64 | `50` | **tau** — max time (ms) the oldest queued tx waits before a partial batch dispatches anyway (mempool path). Caps worst-case added latency when traffic is too light to hit `k_target`. |
| `batch_verify_workers` | usize | `4` | **N** — number of blocking worker tasks draining the mempool batch queue concurrently. Each runs one aggregate crypto call at a time, so more workers only help up to available CPU cores. Mempool-only; block import is single-threaded per block. |
| `batch_verify_queue_capacity` | usize | `4096` | Bounded mempool queue capacity. Submissions beyond this are shed immediately with a pool `ImmediatelyDropped` error instead of being queued. |

**Tuning notes** (see `scripts/tests/batch-verify-perf/FINDINGS.md` for the block-import harness, its methodology, and open caveats):

- Batching only pays off once a batch carries *many* proof-txs — at batch sizes ≤5 the aggregate call's fixed cost dominates and there's ~nothing to amortize (measured on the block-import path). `batch_verify_max_batch_size` / `batch_verify_target_batch_size` are the levers to explore the favourable regime; workload needs enough proof-txs per block/window to actually fill a batch (see the FINDINGS.md note on the DUST-output ceiling that limited the current harness).
- `batch_verify_target_batch_size` and `batch_verify_max_age_ms` jointly set the mempool latency/throughput trade-off: raising either trades added tail latency for larger, more efficient batches.
- `batch_verify_workers` is CPU-bound — sizing it above available cores just adds contention, not throughput.
- `batch_verify_queue_capacity` is a burst-absorption/shedding valve, not a throughput knob: too small drops legitimate traffic under burst (watch `queue_rejected_total`), too large just grows tail latency for queued txs.
- Useful metrics while tuning: `midnight_batch_verify_batch_size` (histogram), `midnight_batch_verify_duration_seconds` (per-batch aggregate crypto time, both ingress paths), `midnight_batch_verify_batches_total{outcome=...}`, `midnight_batch_verify_txs_total`, `midnight_batch_verify_queue_depth`, `midnight_batch_verify_queue_rejected_total`, `midnight_batch_verify_dispatch_reason_total{trigger=...}` (`k_target` vs `tau`), `midnight_batch_verify_fallback_total` (mempool-only).
- New `just` targets for the block-import A/B harness: `just batch-verify-perf-prime <NODE_IMAGE> <TOOLKIT_IMAGE>` (prime + archive a proof-heavy chain, once) and `just batch-verify-perf-bench <NODE_IMAGE>` (A/B benchmark off vs on against the primed archive) — see `scripts/tests/batch-verify-perf/README.md`.

## 🔑 Running the toolkit outside the built image

The ledger pin on this branch reports `static/version = 10-dust-zswap-v3`, and neither `srs.midnight.network/dust/10-dust-zswap-v3/*` nor `.../zswap/10-dust-zswap-v3/*` is published — so a local prover can't fetch its Dust/Zswap keys, it has to compile them:

```
just seed-zk-keys
```

It writes to `$MIDNIGHT_PP` / `$XDG_CACHE_HOME/midnight/zk-params` / `$HOME/.cache/midnight/zk-params` — the same lookup order `MidnightDataProvider` uses — builds `zkir` from `midnight-zkir-v3` on first run (halo2 crate, a few minutes), then hash-checks the generated keys against the `.sha256` files in the pinned ledger checkout. Idempotent, and it reads `static/version` from whatever rev is currently pinned, so re-run it after any pin bump that moves that version. Nothing else is needed; `.envrc` already exports `MIDNIGHT_LEDGER_TEST_STATIC_DIR` for the contract fixtures.

Inside the images this is already handled: `+toolkit-image` bakes the same `+zk-keys` artifacts in and `images/toolkit/Dockerfile` sets `MIDNIGHT_PP=/.cache/midnight/zk-params`. `+build-test-toolkit` and `+local-env-ci` now copy them in too — without that they 404'd on `zswap/10-dust-zswap-v3/spend.prover` and took out `Test Toolkit` and `Local Environment Tests`.

## 🗹 TODO before merging

- [ ] Ready
- [x] Integration validation: with both flags on in a dev preset, sync a chain and author blocks; confirm received blocks import via the wrapper, mempool/authored txs warm the proof cache (no downstream inline-verify error logs), and a bad-proof submission is isolated (not a panic)
- [x] Decide whether to honor/warn on `--pool-type fork-aware` (currently ignored; SingleState is forced)
  - Fine as-is - this is just an experiment, not to be merged 
- [ ] Bump node minor version if this graduates from experiment to a shipped feature
  - It will not be shipped by this PR 

## 📌 Submission Checklist

- [x] Changes are backward-compatible (both flags default off; note: the pool is now always SingleState, so a `--pool-type fork-aware` run now builds SingleState)
- [ ] Self-reviewed the diff
- [x] I have included a change file, or skipped for this reason: experimental/draft branch
- [x] No new todos introduced (the previous `todo!()` fallback is now implemented)

## 🧪 Testing Evidence

- `cargo check -p midnight-node-ledger -p midnight-node` — clean
- `cargo fmt --check` — clean
- `cargo clippy` — no new findings (only pre-existing, workspace-allowed lints on touched files)
- Unit tests:
  - ledger: `fallback_isolates_bad_proof_and_caches_outcomes` (mixed good/bad batch isolates the offender, caches `true`/`false` correctly, correct per-tx results) — passes on all 3 ledger versions
  - node: 7 tests for the worker pool — dispatch at `k_target` and `tau`, bounded-queue shedding (`ImmediatelyDropped`), oneshot resolution, invalid/unavailable → delegate, and native validity-tag shape

## 🔱 Fork Strategy

- [x] Node Client Update

Runtime/pallets are unchanged; the ledger crate is invoked natively by the node. No metadata rebuild required.

The **`undeployed` genesis was rebuilt**, though: the ledger pin's regenerated zswap circuits (`spend.verifier` `544554ef…` → `2098f931…`) invalidated the committed genesis proofs, so the toolkit's replay-from-genesis failed `well_formed` → `batch_proof_verify` with *"Transcript error: Invalid BLS12-381 scalar encoding in proof (batch size: 28)"*. `local` was regenerated too and came out byte-identical — it funds no faucet wallets at genesis, so it has no zswap proofs to invalidate. No chainspec rebuild is needed for either: `res/cfg/dev.toml` points at the `.mn` files by path and `midnight-node-res` `include_bytes!`s them.

## Links

N/A (experimental)


chrispalaskas:
**Field report from the perfnet A/B (both flags on treatment nodes, same binary `c9d9e552` fleet-wide): `batch_verify_mempool` skips state-dependent validation, which admits never-includable transactions and breaks dependent-tx flows.**

In `batch_chain_api.rs`, a batch-verified proof short-circuits straight to a fabricated validity:

```rust
let out = match (results.get(idx), runtime_version) {
    (Some(Ok(())), Some(v)) => {
        WorkerOutcome::Validated(success_validity(v, &item.tx_bytes))
    },
    _ => WorkerOutcome::Delegate,
};
```

The stock path (`pallet-midnight::validate_unsigned` → `LedgerApi::validate_transaction`) checks the tx against current ledger state — dust double spends, `InputNotInUtxos`, `ReplayProtectionViolation`, ctime windows — and rejects conflicts **at submission** with a visible `invalid`. The batch path verifies proofs only, so any well-proven but state-conflicting tx is admitted as Ready.

Observed consequences on perfnet (register-dust + funding-cascade workload, submissions to a treatment node):

- Admitted conflicted txs become pool zombies: `🚫 Rejecting … at pre-dispatch: guaranteed execution would fail: DustDoubleSpend/…IntentAlreadyExists/…InputNotInUtxos` at every authoring slot, then `🚫 Rejected … from mempool` at revalidation — and the eviction emits no watch event (midnightntwrk/midnight-node#2056), so clients time out with zero status.
- Funding cascades regress specifically: a child spending a not-yet-included parent output used to be rejected at submission and retried later; now it is admitted, then evicted + temp-banned by revalidation *before* its parent lands — turning a transient ordering race into a permanent loss.
- Non-conflicting txs on the same node at the same time are included and finalized normally, so this only shows up under dependent/racy workloads — which is why sync/author smoke tests pass.

Same workload against a control node (flags off) behaves as on `main`.

Suggested fix, matching the PR's stated design ("warm the proof cache; downstream `get_verified_transaction` skips the ZK crypto"): after a successful batch verification, **always return `WorkerOutcome::Delegate`** instead of fabricating the validity. The runtime validation then runs the cheap state checks and skips the expensive proof verification via the warmed cache — preserving the amortization win without changing admission semantics. (The fabricated-tag path also bypasses the runtime's `tblock`/`MaxSkippedSlots` ctime simulation, which is a second, subtler semantic drift.)

PR #1892, opened 2026-07-16
feat: add pallet-consensus-engine to support AURA to BABE migration

### LINKED ISSUE #1740: [Aura→BABE] RT-01 — pallet-consensus-engine: consensus state machine + flip logic
**Workstream:** Runtime / Pallets (midnight-node-runtime)  
**Part of:** Aura→BABE consensus migration — [MIP](https://github.com/midnightntwrk/midnight-improvement-proposals/blob/main/mips/mip-0010-aura-to-babe-migration.md)  
**First-pass sizing (to confirm with team):** 8–12 person-days

## Summary
New `pallet-consensus-engine` that owns the dual-engine state machine and the automatic flip described in the Migration State Machine section of the MIP.

## Scope
- `ActiveEngine` flag (Aura | BABE) with transition guards
- `BootstrapArmed` and `FlipAtSlot` storage + governance-gated setters
- Automatic flip at the recorded boundary slot: switch `ActiveEngine` → BABE, derive epoch-0 authorities from the live committee, emit `NextEpochData` digest announcing epoch-1 authorities
- Armed-window inertness (no flip on arm); scheduled-flip enforcement (defer until GRANDPA finalizes ≥1 injected block)
- Epoch orchestration / boundary alignment hooks

## Acceptance criteria
- Block-import requirements (1.5), (2.1), (2.4), (2.7), (2.8) satisfied
- Flip occurs only at a BABE epoch boundary slot ≥ parent slot
- State transitions covered by unit tests

## Dependencies
- Blocked by SPK-01 (epoch-boundary binding) and RT-04 (committee staging)

---
_First-pass breakdown derived from the MIP Implementation section. Sizing and implementation plan to be reviewed and confirmed by the team._

LGLO:
# Overview

State machine for AURA to BABE migration.
Issue: #1740 

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A


pallets/consensus-engine/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+		/// [`Hooks::on_initialize`].
+		#[pallet::call_index(1)]
+		#[pallet::weight(T::WeightInfo::schedule_flip())]
+		pub fn schedule_flip(origin: OriginFor<T>) -> DispatchResult {
+			T::GovernanceOrigin::ensure_origin(origin)?;
+			if EngineState::<T>::get() == State::ArmedBabe {
+				EngineState::<T>::put(State::ScheduledFlip);
+			}
+			Ok(())
+		}
```
Klapeyron, line 190: isn't this step automatic once `arm_babe` succeeds? I mean if we need additional governance action, due we can after `arm_babe` expect that flip will occur on the next session rotation

runtime/src/lib.rs
```diff
@@ -939,6 +947,16 @@ impl pallet_throttle::Config for Runtime {
 	type WindowSize = WindowSize;
 }
 
+impl pallet_consensus_engine::Config for Runtime {
```
Klapeyron, line 947: please also add FindAuthor for pallet_authorship

pallets/consensus-engine/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+			);
+		}
+
+		fn current_slot_from_aura_digest() -> Option<Slot> {
+			frame_system::Pallet::<T>::digest().logs.iter().find_map(|log| {
+				log.as_pre_runtime().and_then(|(id, mut data)| {
+					(id == AURA_ENGINE_ID).then(|| Slot::decode(&mut data).ok()).flatten()
+				})
+			})
+		}
```
Klapeyron: there exists https://paritytech.github.io/polkadot-sdk/master/pallet_aura/pallet/type.CurrentSlot.html which we can use

pallets/consensus-engine/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+
+		fn has_babe_pre_digest() -> bool {
```
Klapeyron: I suggest to just use `pallet_babe::FindAuthor` implementation

pallets/consensus-engine/src/mock.rs
```diff
@@ -0,0 +1,130 @@
...
+		DigestItem::PreRuntime(AURA_ENGINE_ID, Slot::from(slot).encode()),
+		DigestItem::PreRuntime(
+			BABE_ENGINE_ID,
+			BabePreDigest::SecondaryPlain(SecondaryPlainPreDigest {
+				authority_index: 0,
+				slot: Slot::from(slot),
+			})
+			.encode(),
+		),
+	]);
```
Klapeyron: comment maybe not about this particular test code, but rather a request to document somewhere that ordering of this digests matters, due tools that are out of our control may relay on that https://github.com/polkadot-js/api/blob/db81417393858fa40b7b8168d80d52280db407d2/packages/api-derive/src/type/util.ts#L13

pallets/consensus-engine/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+				// Before arming, the node must not emit BABE pre-digests. A block
+				// carrying one would let `pallet-babe` self-initialize its genesis
+				// epoch prematurely (see `BabeMigration::on_arm`), so reject it. This
+				// is deterministic — every node reads the same header — so a
+				// misbehaving author's block is rejected on import, not just locally.
+				State::Aura => {
+					assert!(
+						!Self::has_babe_pre_digest(),
+						"BABE pre-runtime digest present while in state 'Aura'",
+					);
```
Klapeyron, line 143: maybe this match for https://github.com/midnightntwrk/midnight-node/pull/1892/changes#r3613066028 will be appropriate? Also I think we should keep in hold this semantics here, so for the flip ensure right order of the digests, so if we will have both ensure aura is first and after the flip don't allow for aura. Also it is worth IMO to relay on FindAuthor APIs instead of parsing of digests directly. The last guard may be worth to comment and remove once pallet_aura will be removed from the runtime

pallets/consensus-engine/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+
+	/// Binding invoked at the consensus flip to bootstrap `pallet-babe`.
+	///
+	/// The consensus-engine pallet owns *when* the flip happens; the runtime owns
+	/// *how* to bootstrap BABE (its storage and the AURA→BABE key bridge).
+	pub trait BabeMigration {
+		/// Called when BABE is armed, before the node starts emitting BABE
+		/// pre-runtime digests. Pre-seeds `pallet-babe` so its genesis self-init
+		/// (guarded on `GenesisSlot == 0`) never fires while armed, which would
+		/// otherwise deposit a bogus epoch descriptor into a header we can't retract.
```
Klapeyron: The pallet abstractions are correct as written, but after the second look they still lean on Aura and Babe being present in the runtime. With the current shape, wiring Config to both pallets and using their APIs is a reasonable choice — it simplifies a few abstractions without adding much complexity, and it’s a normal pattern in the polkadot-sdk.

pallets/consensus-engine/src/lib.rs
LGLO: For this one I prefer to read pre-digest and not depend on pallets ordering in runtime.

pallets/consensus-engine/src/mock.rs
LGLO: Importance of ordering of PreRuntime digests doesn't really belong to the Runtime. It should come when node modifications are coming in place.

pallets/consensus-engine/src/lib.rs
LGLO: I'll reimplement pallet to depend on the two other pallets.

runtime/src/lib.rs
Klapeyron, line 947: also pallet_timestamp has a hook for Aura (`OnTimestampSet`), we should adjust it

runtime/src/lib.rs
```diff
@@ -1131,6 +1149,10 @@ mod runtime {
 	// Throttling
 	#[runtime::pallet_index(51)]
 	pub type Throttle = pallet_throttle::Pallet<Runtime>;
+
+	// Consensus engine transition state machine
+	#[runtime::pallet_index(52)]
+	pub type ConsensusEngine = pallet_consensus_engine::Pallet<Runtime>;
```
Klapeyron, line 1151: should we for now exclude calls for this pallet?

runtime/src/lib.rs
LGLO: Done

pallets/consensus-engine/src/lib.rs
LGLO: This conflicts with checking that both AURA and BABE predigests are present and in correct order.
I'll keep the lower level code here.

pallets/consensus-engine/src/lib.rs
LGLO: It is documented in the pallet documentation that it is manual, because we want to observe final blocks with the pre-digest before hitting the button.

runtime/src/lib.rs
LGLO: I don't think so. Redundant work. I assume full power and responsibility of governance authority. It would impair testing but isn't really increasing safetly.

pallets/consensus-engine/src/mock.rs
Klapeyron: It does, due if it is an invariant for us (preserving external API tools to operate) then runtime has to verify if that is consistent

pallets/consensus-engine/src/lib.rs
Klapeyron: But it is not making it dependent on the ordering of pallets in runtime, actually the right source of truth in particular belongs to pallet_sidechain, which we should also adjust to be generic here

runtime/src/lib.rs
LGLO: Thank you!
Done

pallets/consensus-engine/src/lib.rs
LGLO: I've updated the function in the assert, it requires a one unique BABE pre-digest that is positioned after AURA pre-digest.

pallets/consensus-engine/src/mock.rs
LGLO: I updated pre-runtime digests checks to require unique BABE after AURA for armed aura phase.

runtime/src/lib.rs
```diff
@@ -395,7 +395,9 @@ impl pallet_aura::Config for Runtime {
 }
 
 impl pallet_authorship::Config for Runtime {
-	type FindAuthor = pallet_session::FindAccountFromAuthorIndex<Self, Aura>;
+	// `ConsensusEngine` implements `FindAuthor`, routing author lookup to AURA before
+	// the consensus flip and BABE after it.
```
Klapeyron: nit: I think not needed comment, we do not need to know here what it is doing under the hood

Klapeyron (dismissed):

Klapeyron (approved):

PR #1902, opened 2026-07-21
feat(toolkit): make the node RPC request timeout configurable

### LINKED ISSUE #1853: toolkit: node RPC request timeout is fixed at jsonrpsee's 60s default — Metadata_metadata_at_version times out on slow hardware
During Antithesis QA bring-up (shielded), `midnight-node-toolkit generate-txs register-dust-address` fails repeatedly with:

```
failed to get transactions: error creating client
Error: GetTransactions(NodeClientError(OnlineClientAtBlockError(CannotGetMetadata {
  block_hash: Hex("c0607df0284a21df42e67708631b7dc3a0c4638fb2f54150237588fd432f400c"),
  reason: "Error calling Metadata_metadata_at_version: Backend error: RPC error: RPC error: RPC error: client error: Request timeout" })))
```

## Cause

The toolkit's node client is built with `RpcClient::from_insecure_url` (`util/toolkit/src/client.rs`), which uses jsonrpsee's default **60 s request timeout** with no way to override it from the CLI (`CLIENT_CONNECT_TIMEOUT` only bounds the connect/backoff loop, and the `timeout` param of `MidnightNodeClient::new` isn't user-reachable for this path).

The wallet-state fetch behind `register-dust-address`/`single-tx` creates an `OnlineClientAtBlock` per fetched block; each one re-fetches metadata via the `Metadata_metadata_at_version` runtime call — a heavy WASM execution returning ~MBs. On slow/emulated hardware (Antithesis), and with other RPC consumers loading the node (chain-indexer initial sync in our case), that call can exceed 60 s, killing the whole command.

The failure is transient — our retry loop eventually gets through (attempt 3 succeeded) — but each timed-out attempt costs ~3.5 minutes of wall-clock inside the timeouts, on the critical bring-up path of a time-boxed test run. And since new blocks accumulate during each stalled attempt, the next attempt has *more* per-block metadata fetches to survive, so the failure mode compounds.

## Ask

Make the RPC request timeout configurable, e.g. a `--rpc-request-timeout <secs>` flag or `TOOLKIT_RPC_REQUEST_TIMEOUT` env var threaded into a `jsonrpsee` client builder in `MidnightNodeClient::new_without_timeout`. (Possibly also worth caching metadata across `at_block` clients when the runtime version hasn't changed — the per-block re-fetch is what makes this quadratic-ish on long fetches.)

## Context

- toolkit at `38b86540`, run against a 5-node docker network under Antithesis (no fault injection active yet at the time of failure — this is plain emulated-hardware slowness + RPC contention).
- Downstream mitigation: antithesishq/poc-shielded points the faucet-init container at a node the indexer isn't syncing from, and retries; that shrinks but doesn't remove the exposure — every toolkit workload command under fault injection faces the same fixed 60 s ceiling.

wbaxterh:
## Summary

Fixes #1853.

Adds a global `--rpc-request-timeout <secs>` flag (env: `MN_RPC_REQUEST_TIMEOUT`)
to the toolkit, replacing the fixed 60 s jsonrpsee default that heavy requests
like `Metadata_metadata_at_version` can exceed on slow/emulated hardware or a
loaded node — the Antithesis bring-up failure described in the issue.

## Changes

- `client.rs`: new `rpc_client_with_timeout()` mirrors subxt's own jsonrpsee
  client construction (same WS transport, buffer capacity, and error mapping)
  with `.request_timeout(...)` applied. The timeout lives in a process-wide
  `OnceLock` set by `main` before dispatch — the same pattern as
  `--replay-concurrency` configuring the global rayon pool.
- `MidnightNodeClient::new_without_timeout` now uses it (this covers all
  `Source`/fetcher/sender paths), and the three direct
  `OnlineClient::from_insecure_url` call sites (`root-call`,
  `runtime-upgrade`, `update-ledger-parameters`) are switched to
  `from_rpc_client`, so every node-facing command honours the flag.
- Env var follows the toolkit's existing `MN_*` convention rather than the
  `TOOLKIT_*` name suggested in the issue.
- `jsonrpsee` added as a direct dep pinned to the workspace version (0.24.11,
  identical to what subxt resolves — no new duplicate in the graph).

## Verification

- Against a WS server that completes the handshake then never responds
  (the slow-node failure mode): `--rpc-request-timeout 2` fails each request
  after ~2 s and the connect-retry loop advances (3 attempts in 11 s);
  the default hangs a single request for the full 60 s;
  `MN_RPC_REQUEST_TIMEOUT=1` behaves identically to the flag (5 attempts in 11 s).
- Happy path: `show-wallet` against a local dev node succeeds unchanged with
  the flag set.
- `cargo test -p midnight-node-toolkit --lib` (124 passed) and the trycmd
  CLI snapshot tests pass; clippy and fmt clean.

## Notes

- The issue also suggests caching metadata across `at_block` clients (the
  per-block re-fetch makes long fetches quadratic-ish). That's a separate,
  larger change — happy to look at it as a follow-up if the team wants.


ozgb (approved):

ozgb (changes requested):
Would prefer it if the timeout was passed via an arg rather than set in a global static var

wbaxterh:
Reworked in 588c9e0 — the `OnceLock` global is gone and the timeout is now passed as an argument end to end:

- `main` resolves `--rpc-request-timeout` / `MN_RPC_REQUEST_TIMEOUT` to a `Duration` (falling back to the exported `DEFAULT_RPC_REQUEST_TIMEOUT` = jsonrpsee's 60s) and hands it to `run_command`.
- `run_command` threads it into every command that opens a node connection — including the deep `generate-txs` paths (source/fetcher/sender), which was the original motivation in #1853.
- `MidnightNodeClient::new` now takes explicit `connect_timeout` and `request_timeout` params (the old `timeout` arg was really the connect/backoff bound, so this also disambiguates it).
- CLI surface is unchanged; test call sites pass the default explicitly.

`cargo fmt --check`, clippy (`--tests`, zero lints), and the 124 lib tests are green. Ready for another look whenever you have a moment — and happy to adjust the threading shape if you'd prefer a config struct over positional params.

wbaxterh:
Same housekeeping note as #1927: branch is updated onto current main (the arg-threading refactor merged cleanly, local fmt/clippy/tests green) and the fresh workflow runs await maintainer approval to start. Re-review whenever you have a moment @ozgb 🙏

wbaxterh:
@ozgb bumping this one. The `OnceLock` global came out in 588c9e0 and the timeout is threaded as an argument from `main` through `run_command` into every command that opens a connection, including the `generate-txs` source/fetcher/sender paths that were the original motivation in #1853. Branch is current with main as of a268b0a.

Re-review whenever you have a moment 🙏 If positional params aren't the shape you want, I'll move it to a config struct.

wbaxterh:
Merged current main into the branch (5c86dfc8) to clear the conflicts from the replay-checkpoint work that landed upstream — `replay_checkpoint_interval` is captured before the `src` move in `TxGenerator::new` and in each command, alongside the arg-threaded `rpc_request_timeout`. `cargo check`/`clippy`/`fmt` green locally on the toolkit including tests. No force-push, per repo convention. Ready for another look @ozgb.

PR #1664, opened 2026-06-09
docs: add SDET-oriented "How to test the Midnight Node" guide

rsporny:
# Overview

Adds an SDET-oriented onboarding doc at `docs/tests/how-to-test-node.md`. The
guide explains how the node is structured (node vs runtime, pallets, extrinsics,
metadata), what we ship per release (deterministic WASM + attestations), and how
each layer of automation is exercised — from the Rust e2e suite against
local-env and qanet to the script-based smoke / toolkit family, the Cardano
hard-fork rehearsal at `local-environment/src/networks/local-env/hardfork-pv11.sh`,
the CI surface, and the release-evidence process. It also collects test data,
env vars, repo scripts, known gotchas, and internal video references in one
place so new SDETs don't have to grep for them.

Audience: SDETs already familiar with Substrate-flavoured blockchains, Cardano
basics, and Rust tooling. Intended as a draft presentation / reference doc.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] All commits are signed off (`git commit -s`) for the DCO
- [x] Changes are backward-compatible (or flagged if breaking)
- [x] Pull request description explains why the change is needed
- [x] Self-reviewed the diff
- [x] I have included a change file, or skipped for this reason: docs-only change, no product impact
- [x] If the changes introduce a new feature, I have bumped the node minor version 
- [x] Update documentation (if relevant)
- [x] Updated AGENTS.md if build commands, architecture, or workflows changed 
- [x] No new todos introduced

## 🧪 Testing Evidence

Docs-only change. Verified locally:

- `docs/tests/how-to-test-node.md` renders correctly in GitHub markdown preview.
- Code paths, script paths, and config keys mentioned in the guide were
  cross-checked against the repo (`Earthfile`, `tests/e2e/`, `scripts/`,
  `local-environment/`, `.github/workflows/`,
  `local-environment/src/networks/local-env/configurations/genesis/shelley/genesis.json`).
- All referenced GitHub workflow jobs and Earthly targets exist on `main`.

- [ ] Additional tests are provided (if possible) 

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other
- [x] N/A

## Links

- Closes shieldedtech/shielded-qa#15

PR #1921, opened 2026-07-24
fix: get_block_by_hash handles large block numbers

Klapeyron:
# Overview

In midnight-node repo, partner-chains/toolkit/data-sources/mock/src/block.rs:53, get_block_by_hash computes let timestamp = block_number * 20000; where block_number is a u32 parsed from arbitrary hash bytes. This multiplication overflows u32 and panics in debug builds ('attempt to multiply with overflow'), crashing any CFG_PRESET=dev node run from a debug binary. Fixed by widening to u64 before multiplying.

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11368,6 +11368,7 @@ dependencies = [
  "sidechain-mc-hash",
  "sp-partner-chains-bridge",
  "sp-timestamp",
+ "tokio",
 ]
 
 [[package]]
--- a/partner-chains/toolkit/data-sources/mock/Cargo.toml
+++ b/partner-chains/toolkit/data-sources/mock/Cargo.toml
@@ -26,6 +26,9 @@ pallet-sidechain-rpc = { workspace = true, optional = true }
 authority-selection-inherents = { workspace = true, optional = true }
 sp-partner-chains-bridge = { workspace = true, optional = true }
 
+[dev-dependencies]
+tokio = { workspace = true }
+
 [features]
 default = ["std", "block-source", "candidate-source"]
 std = [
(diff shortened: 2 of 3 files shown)


wbaxterh (commented):
Verified this fix against the forward computation and ran the numbers — it's correct and more impactful than the description suggests.

**Verification.** `get_latest_stable_block_for` computes `block_number = (timestamp_millis / 20000) as u32` with u64 millis, so the reverse must widen before multiplying — exactly what this change does.

**Impact is bigger than "large block numbers":** the multiply overflows u32 for any `block_number ≥ 214,749`, and since the first four big-endian bytes of a real hash are effectively uniform, ~99.995% of real hashes trigger it. So in a debug build `get_block_by_hash` panics near-deterministically, and in a release build it silently wraps and returns a wrong timestamp — the silent-corruption half is arguably the worse bug and worth a line in the PR description.

**One adjacent observation** (pre-existing, non-blocking): the docstring says this is the reverse of `get_latest_stable_block_for`, but the forward path sets `slot = block_number as u64` while this reverse sets `slot = epoch.into()` — round-tripping a block through hash → `get_block_by_hash` therefore changes its slot. If intentional, a comment would help; otherwise `McSlotNumber(block_number as u64)` here would make the reversal exact. Happy to open a follow-up issue if you'd rather keep this PR minimal.

**Housekeeping** so the checks pass: your commit is DCO-signed, so the first checklist box can be ticked; the `[lib] test = false` removal is right (it arrived with #1664 when the crate had no tests, and the new test needs the harness); and the repo's changes-check will want either a change file under `changes/node/changed/` or a maintainer to apply the `skip-changes-check-*` labels since there's no linked issue.

Nice catch — dev-preset debug nodes crashing on a mock lookup is the kind of papercut that eats an afternoon.

LGLO (approved):

PR #1800, opened 2026-07-01
feat: generic session keys & session pallet migration of PR 1026

Klapeyron:
# Overview

Integrates upstream Partner Chains changes for automatic session key registration ([#1026](https://github.com/input-output-hk/partner-chains/pull/1026)) and removal of `pallet-partner-chains-session` ([#1025](https://github.com/input-output-hk/partner-chains/pull/1025)), adapted for midnight-node's `polkadot-stable2603` workspace.
Partner Chains runtimes now use stock Substrate `pallet_session` directly. `pallet_session_validator_management` (`SessionCommitteeManagement`) implements `SessionManager` and `ShouldEndSession` inline (the old `PalletSessionSupport` wrapper and `pallet-partner-chains-session` crate are removed). Committee members' session keys are registered automatically at genesis and on each committee rotation via `pallet_session::SessionInterface::set_keys`, so block producers no longer need to call `set_keys` manually.
Midnight-node-specific wiring:
- Runtime migrates off `pallet-partner-chains-session` to a direct `pallet_session::Config` impl delegated to `SessionCommitteeManagement`.
- `pallet_session::historical` is enabled and wired (`Historical` pallet + `historical::Config`) so privileged `SessionInterface::set_keys` works under `polkadot-stable2603`'s ownership-proof checks (dispatching the `set_keys` extrinsic with an empty proof silently fails).
- `pallet_session` extrinsics are disabled via `#[runtime::disable_call]`; key registration is internal only.
- Genesis chain spec no longer duplicates session keys — registration happens through committee management genesis.
- The local `session_manager` module (`ValidatorManagementSessionManager`) is removed.
- Runtime `spec_version` bumped to `002_001_000`; metadata regenerated.

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [ ] All commits are signed off (`git commit -s`) for the DCO
- [ ] Changes are backward-compatible (or flagged if breaking)
- [ ] Pull request description explains why the change is needed
- [ ] Self-reviewed the diff
- [ ] I have included a change file, or skipped for this reason: 
- [ ] If the changes introduce a new feature, I have bumped the node minor version
- [ ] Update documentation (if relevant)
- [ ] Updated AGENTS.md if build commands, architecture, or workflows changed
- [ ] No new todos introduced

## 🧪 Testing Evidence

Please describe any additional testing aside from CI:

- [ ] Additional tests are provided (if possible)

## 🔱 Fork Strategy

- [ ] Node Runtime Update
[...]

Issue #1759, opened 2026-06-23
[Aura→BABE] SPK-02 (spike) — Session-key carrier decision


DarrenOliveiro-Priestnall:
**Workstream:** Spike / open design question (midnight-node)  
**Part of:** Aura→BABE consensus migration — [MIP draft](https://github.com/midnightntwrk/midnight-improvement-proposals/blob/docs/mip-aura-to-babe-migration/mips/mip-xxx-aura-to-babe-migration.md)  
**First-pass sizing (to confirm with team):** 2–3 person-days

## Summary
Decide the session-key carrier: phantom holder over the current `pallet_partner_chains_session` vs a plain generic session key (if generic-key work lands this quarter). Requirement is unchanged (registered, proven, usable keys at flip); the choice affects only sequencing.

## Scope
- Compare phantom-holder vs generic session key
- Recommend carrier and sequencing

## Acceptance criteria
- Decision recorded; RT-03 implementation approach unblocked

## Dependencies
- Blocks RT-03

---
_First-pass breakdown derived from the MIP Implementation section. Sizing and implementation plan to be reviewed and confirmed by the team._

rsporny:
### Test evidence — pallet-session migration (#1800) + aura-fork removal (#1700)

Verified on a 5-validator `local-env` at runtime 2.1.0 (`specVersion 2001000`) over **2200 finalized blocks (#31–#2230, ~221 Cardano epochs / ~440 sessions)**, including a **live permissioned-candidate removal (5→3, D(5,0) held)** performed mid-run via `change-federated-ops`.

**Committee / seat behaviour**
- Removal auto-detected at Cardano epoch 132 (N: 5→3). Removed FNOs produced **507 blocks before, 0 after** (only the 2 expected boundary blocks at the transition epoch), confirming they stop producing.
- The 3 remaining FNOs absorbed the freed seats; with P=5 / N=3 each epoch draws 1–2 seats per FNO (guaranteed `floor(5/3)=1` + random-fill of the 2 leftovers) and it averages out over Cardano epochs — final totals **568 / 564 / 559** (kept) vs **255 / 254** (removed, frozen at removal).

**Assertions — all pass (2200/2200 blocks, through the 5→3 change):**
[...]

PR #1700, opened 2026-06-16
feat: replaced aura fork with sc-partner-chains-consensus

Klapeyron:
# Overview

Remove `sc-partner-chains-consensus-aura` and `sp-partner-chains-consensus-aura`. Add consensus-agnostic `sc-partner-chains-consensus`, which runs the full Partner Chains inherent check at whichever point the consensus gadget checks inherents: `PartnerChainsBlockImport` (with `PartnerChainsBodyRestore`) wraps a consensus block import for stacks that check inherents at import (e.g. `BabeBlockImport`), and `PartnerChainsVerifier` wraps the import-queue verifier for stacks that check inherents there instead (e.g. Aura). `PartnerChainsProposerFactory` injects the `mcsh` digest at proposal time.

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [ ] All commits are signed off (`git commit -s`) for the DCO
- [ ] Changes are backward-compatible (or flagged if breaking)
- [ ] Pull request description explains why the change is needed
- [ ] Self-reviewed the diff
- [ ] I have included a change file, or skipped for this reason: 
- [ ] If the changes introduce a new feature, I have bumped the node minor version
- [ ] Update documentation (if relevant)
- [ ] Updated AGENTS.md if build commands, architecture, or workflows changed
- [ ] No new todos introduced

## 🧪 Testing Evidence

Please describe any additional testing aside from CI:

- [ ] Additional tests are provided (if possible)

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A

PR #1926, opened 2026-07-24
test: add e2e block-production & committee regression guard

rsporny:
# Overview

Adds an e2e regression guard for block production and committee rotation,
locking into CI the deterministic invariants that were validated manually for
the aura-fork removal (#1700) and the pallet-session migration (#1800).

`tests/e2e/tests/block_production.rs` audits the newest finalized blocks and asserts:

- **Authorship legitimacy** — every block's AURA author
  (`aura.authorities.at(parent)[slot % len]`) is a member of
  `sidechain_getEpochCommittee` for the parent's sidechain epoch. Reading the
  **parent** state is deliberate: the first block of a session is produced by
  the *previous* committee (rotation runs during that block), so this is
  boundary-safe. As the block is finalized (the node's verifier already checked
  its seal), this confirms only legitimate current-committee members produce.
- **Committee source agreement** — per epoch, on-chain `aura.authorities` equals
  `sidechain_getEpochCommittee` (two independent sources must agree).
- **session/aura alignment** — `session.validators` and `aura.authorities` stay
  equal length (the pallet_session → aura handoff).
- **session index monotonic** and **finalized heights contiguous**.

The test is `#[ignore]` so the parallel suite skips it; `+local-env-ci` runs it
as the **final** step (after the suite + health check) so it analyses the most
blocks — bounded to the newest 200. Supporting `MidnightClient` helpers were
added (`get_epoch_committee`, `get_block_digest_info`, `get_aura_authorities_hex_at`,
`get_session_index_at`, `get_session_validators_len_at`, `get_sidechain_epoch_and_slot`,
`get_mainchain_epoch`). Also drops a pre-existing unused `self` import in the
cnight e2e test (separate commit).

Statistical/experimental analysis (seat-allocation convergence, per-Cardano-epoch
tables, permissioned-candidate-removal detection) intentionally stays out of CI —
it needs hundreds of epochs and a mid-run governance action the ~120-block CI run
can't provide.

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

Ran the ignored test against a local `local-env` (runtime 2.1.0):

```
cargo test --test e2e_tests --no-default-features --features local -- --ignored block_production --nocapture

INFO block_production_and_committee_invariants: slots-per-epoch = 5
INFO block_production_and_committee_invariants: auditing finalized blocks #5691..#5891 (201 blocks)
INFO block_production_and_committee_invariants: ✅ block-production audit passed: 200 block(s), 41 epoch(s), 40 session-boundary block(s)
test block_production::block_production_and_committee_invariants ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 31 filtered out; finished in 14.46s
```

- Normal suite run (without `--ignored`) reports it as `ignored` (0 run), confirming it only runs in the dedicated final `+local-env-ci` step.
- `cargo fmt` and `cargo clippy -p midnight-node-e2e --features local --tests` are clean.

## 🔱 Fork Strategy

- [x] N/A 

## Links

- Issue: https://github.com/midnightntwrk/midnight-node/issues/1759

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -1834,6 +1834,10 @@ local-env-ci:
             echo "=== post-suite liveness check ===" && \
             cd "$ROOT" && \
             ./local-environment/check-health.sh -u http://localhost:9933 -b 50 -t 360 && \
+            echo "=== block-production / committee audit (runs last, over the most blocks) ===" && \
+            ( cd "$ROOT/tests/e2e" && \
+              cargo test --test e2e_tests --no-default-features --features local -- --ignored block_production --nocapture ) && \
+            cd "$ROOT" && \
             rm -f /root/.docker/config.json
     END
 
--- a/changes/node/added/block-production-committee-e2e.md
+++ b/changes/node/added/block-production-committee-e2e.md
@@ -1,0 +1,24 @@
+#tests #ci
+# Add e2e regression guard for block authorship & committee rotation
+
+Adds an ignored e2e test (`tests/e2e/tests/block_production.rs`) that audits the
+newest finalized blocks and asserts the deterministic invariants behind the
+aura-fork removal (#1700) and pallet-session migration (#1800):
+
+- every block's AURA author (`aura.authorities.at(parent)[slot % len]`) is a
+  member of `sidechain_getEpochCommittee` for the parent's sidechain epoch
+  (parent-state read → boundary-safe: the first block of a session is produced
+  by the previous committee);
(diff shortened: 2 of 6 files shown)


LGLO:
We are going to remove AURA, work is in progress, it should happen this year. I think it's too late to bring these tests.

rsporny:
I'll adapt this once we migrate to BABE.

PR #1922, opened 2026-07-24
fix(toolkit): self-funded register-dust-address with multiple NIGHT UTXOs

jacek-kurkowski-shielded:
# Overview

Closes: #1896
Closes: #1906

Self-funded `generate-txs register-dust-address` (no `--funding-seed`) panicked with `InsufficientDustForRegistrationFee` whenever the wallet held more than one NIGHT UTXO.

The builder requested a fee allowance (`allow_fee_payment`) equal to the retroactive DUST summed over *all* of the wallet's NIGHT UTXOs, but the ledger only grants the retroactive DUST of generationless NIGHT spent in the intent's *guaranteed* unshielded offer - which holds a single UTXO, since more would exceed the ledger's `time-to-dismiss` limit (`well_formed` enforces it.

The builder now selects the generationless UTXO with the most accrued retroactive DUST for the guaranteed offer and requests exactly its availability. UTXOs already backing DUST generation (e.g. after a register/deregister round-trip) are excluded via a new `BuilderContext::backs_dust_generation` method. Balancing failures now return a `RegisterDustAddressError` with actionable guidance (consolidate NIGHT, wait for DUST to accrue, or use `--funding-seed`) instead of panicking.

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [X] All commits are signed off (`git commit -s`) for the DCO
- [X] Changes are backward-compatible (or flagged if breaking)
- [X] Pull request description explains why the change is needed
- [X] Self-reviewed the diff
- [X] I have included a change file, or skipped for this reason: 
- [X] If the changes introduce a new feature, I have bumped the node minor version
- [X] Update documentation (if relevant)
- [X] Updated AGENTS.md if build commands, architecture, or workflows changed
- [X] No new todos introduced, except `IndexerContext::backs_dust_generation` (`todo!()`, consistent with that file's existing R3/R4 indexer placeholders)

## 🧪 Testing Evidence

New offline regression suite `util/toolkit/tests/register_dust_address.rs` (no node or proof server needed - registration txs carry no ZK proofs; built against a hand-crafted `LedgerState` and validated/applied with the real ledger code):

- multi-UTXO and single-UTXO self-funded registration succeed and apply with `TransactionResult::Success`
- UTXOs already backing generation are excluded from the allowance
- zero accrued DUST and empty wallet return the actionable error instead of panicking

[...]

PR #1927, opened 2026-07-25
fix(toolkit): verify the finalized chain before reporting a send as failed

### LINKED ISSUE #1854: toolkit: sender finality watcher loses txs across fork retractions — reports FAILED_TO_FINALIZE for a tx finalized 22s after send
During Antithesis QA bring-up (shielded, node/toolkit at `38b86540`), `midnight-node-toolkit generate-txs claim-rewards` failed with:

```
SENDING url="ws://midnight-node-2:9934" midnight_tx_hash="0x0e882f028441f503e7c40438161215f58ef76d5bd6ba0a7ea5481baa8d38ea23"
SENT
BEST_BLOCK                                   <- t=284.4s
Timeout waiting for finalization after 60 seconds
FAILED_TO_FINALIZE                           <- t=344.5s
Failed to send tx 0: tx reached best block but was not finalized within timeout
Error: SendTransactions("one or more destination tasks failed")
```

**The tx was on-chain and finalized 22 seconds after the send.** The chain-indexer log for the same run shows tx `0e882f02…` in block **#4**, finalized at t=306.5s — canonical blocks #2 and #3 are empty:

```
t=285.5  block #1 finalized  txs=1   (the bridge DistributeNight)
t=292.2  block #3 finalized  txs=0
t=306.5  block #4 finalized  txs=1   <- contains 0e882f02…
```

## What happened

The `BEST_BLOCK` status fired at t=284.4s, when the best block was ~#3 — but canonical #2/#3 are empty and the tx finalized in #4. So the tx was first included in a short-lived fork block that got retracted (a normal early-chain aura fork), returned to the pool, and was re-included in #4.

The sender's finalization watcher (`util/toolkit/src/sender.rs`, `FINALIZED_TIMEOUT = 60s` after `BEST_BLOCK`) never followed the tx through the retraction — it kept waiting on the dead branch until the hard timeout, then failed the whole command, even though the tx had been finalized for 38 seconds at that point.

## Impact

Any caller treating the toolkit exit code as the outcome of the send gets a false negative and may retry a claim/transfer that already landed. For `claim-rewards` (at-most-once semantics) that's particularly awkward. On local-env-style chains this window (early forks right after chain start) is exactly when bring-up scripts run.

## Ask

- Follow the tx across retractions: on `Retracted`/`Usurped`, keep watching for the re-inclusion (subxt reports these statuses on the watch stream) instead of waiting for finality of the retracted block.
- Or, on watch timeout, check whether the tx made it into the finalized chain before reporting failure.
- Relatedly, `FINALIZED_TIMEOUT`/`BEST_BLOCK_TIMEOUT` are hardcoded (60s/30s) — same configurability ask as #1853.

Downstream we now verify wallet state when the command fails (antithesishq/poc-shielded#12), but exit codes shouldn't lie about landed txs.

wbaxterh:
# Overview

`toolkit send` could report a landed transaction as failed. When a tx's block was retracted by a fork, the watch stream could stay silent about the re-inclusion: the sender waited out the finalization timeout on the dead branch and exited `FAILED_TO_FINALIZE` while the tx had been finalized in another block tens of seconds earlier (22s in the issue's repro). Callers that treat the exit code as the outcome — most importantly claim-rewards' at-most-once semantics — then retried operations that had already landed.

This PR makes the finalized chain the source of truth, in three layers:

1. **Retractions are visible.** `NoLongerInBestBlock` was silently swallowed during the finalization wait; it now logs `tx retracted from best block; watching for re-inclusion`.
2. **The chain gets the last word.** On watch timeout the sender scans finalized blocks newest-first (bounded at 64 below the finalized head) for the extrinsic hash. If found, it logs `FINALIZED_AFTER_RETRACTION` with the actual block hash and exits success — the exit code can no longer contradict a finalized transaction.
3. **Watch phases are configurable** via `MN_SEND_BEST_BLOCK_TIMEOUT` / `MN_SEND_FINALIZED_TIMEOUT` (seconds), per the configurability ask in the issue. Environment variables rather than CLI flags deliberately, to avoid textual conflicts with #1902, which edits the same CLI files.

**Caveat, stated honestly:** the retraction path itself — a real fork retracting and re-including the tx — cannot be forced deterministically in CI. The fallback scan is e2e-tested against real finalized blocks (positive and negative cases), and the wiring between the timeout and the scan is deliberately small.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [x] I have included a change file
- [ ] If the changes introduce a new feature, I have bumped the node minor version — N/A, bug fix
- [ ] Update documentation (if relevant) — the change file documents the new env overrides
- [ ] Updated AGENTS.md if build commands, architecture, or workflows changed — N/A
- [x] No new todos introduced

## 🧪 Testing Evidence

- `cargo fmt --check` and `cargo clippy -p midnight-node-toolkit --tests` — clean.
- New e2e test `finalized_chain_scan_finds_real_extrinsic` (`util/toolkit/tests/toolkit_e2e.rs`) runs the fallback scan against a live dev node: the positive case locates the timestamp inherent from a real finalized block, the negative case (bogus extrinsic hash) correctly reports not-found. Green in ~33s.

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other:
- [x] N/A — toolkit-only change; no node runtime or client update, no network impact.

## Links

Closes #1854


wbaxterh:
Both bot findings addressed in 1f05210:

- **P1 (change-file metadata):** the `#toolkit #fix` tag row is now added. The `PR:` row was already added in a7139f4 (the bot reviewed the initial commit f1313a66, before that follow-up landed), so the changes-check workflow's `pull/1927` search passes.
- **P2 (fixed scan depth vs configurable timeout):** good catch — the depth now scales with `MN_SEND_FINALIZED_TIMEOUT` (timeout/6s blocks + 16-block margin for finality lag, floored at the previous 64), so a long timeout can no longer outrun the scan window. Covered by three new unit tests including the saturation case.

util/toolkit/src/sender.rs
```diff
@@ -363,31 +390,100 @@ impl Sender {
...
+				Ok(Some(b)) => b,
+				Ok(None) => return None,
+				Err(e) => {
+					log::warn!("finalized-chain check: failed to fetch block: {e}");
+					return None;
+				},
+			};
+
+			for ext in &block.block.extrinsics {
+				let ext_hash = format!("0x{}", hex::encode(sp_crypto_hashing::blake2_256(&ext.0)));
```
LGLO: Cosmetic: I would rather compare bytes instead formatting as string.

LGLO (dismissed):
Codex concern is valid, but the PR is a move in good direction anyway.

wbaxterh:
Thanks @LGLO! Both remaining points addressed in 2aa596a:

- **Codex P1 (inconclusive scan ≠ absence)** — valid, and now fixed properly: `find_in_finalized_chain` returns `Ok(Some(hash))` / `Ok(None)` / `Err(FinalityScanError)`, and an RPC failure during the scan now reports **`FINALITY_UNVERIFIED`** with a dedicated `SenderError::FinalityUnverified` variant (message explicitly warns not to treat the exit as proof of failure for at-most-once callers) instead of collapsing into `FAILED_TO_FINALIZE`.
- **Byte comparison** — the target hash is hex-decoded once up front and each extrinsic's blake2-256 is compared as raw bytes; no more per-extrinsic string formatting.

On the red CI: every failing job dies in an infra step — `Login to GHCR` (builds, lint) and `Install Season` (check-changes) — with no output, and the same failures hit #1922's latest commit, so it looks repo-wide rather than anything in this diff. Happy to rebase/retrigger whenever it's healthy again.

wbaxterh:
Branch updated onto current main (clean merge, clippy green locally including the new `ecdsa_signature` test target). The refreshed workflow runs are sitting in *action required* — looks like fork-PR runs now need maintainer approval, so a click on "Approve and run" whenever convenient will get the checks green. The previous red checks were the GHCR/Season infra outage.

wbaxterh:
Rebased onto current main and reconciled with #1943, which landed while this was in review. The two changes solve complementary halves of the false-`FAILED_TO_FINALIZE` problem, and the branch now composes them:

- **#1943 (merged):** the watcher *dies* (terminal pool event, subscription error, stream end) → poll `is_block_finalized` on the already-known including block.
- **This PR:** neither the watcher nor the known block ever learns the truth, because a fork retraction re-included the tx in a **different** block (#1854) — `is_block_finalized(best_block)` answers false while the tx is finalized elsewhere.

In the reconciled sender, #1943's flow runs unchanged; only when it concludes *not finalized* does the finalized-chain scan run as a last resort — found → `FINALIZED_AFTER_RETRACTION` (success), definitively absent → `FAILED_TO_FINALIZE { reason }` exactly as on main, scan couldn't complete → `FINALITY_UNVERIFIED` via the new `SenderError::FinalityUnverified` (per the earlier bot P1: an RPC failure during the scan must not read as proof of absence).

Also preserved from this branch: retraction visibility (`NoLongerInBestBlock` is logged — main still swallows it silently), env-configurable watch phases (`MN_SEND_BEST_BLOCK_TIMEOUT` / `MN_SEND_FINALIZED_TIMEOUT`, defaulting to main's constants), timeout-scaled scan depth with unit tests, and the e2e coverage.

@LGLO the shape changed enough since your approval that a fresh look is warranted — fmt/clippy/lib tests green locally, and the workflow runs will need the approve-and-run click again. 🙏

PR #1928, opened 2026-07-25
fix(ledger): include tblock in strict tx-validation cache key

### LINKED ISSUE #1924: Incorrect `tblock` used when running `well_formed` for first transaction in block
### Context & versions

Issue reported in the indexer, causing a sync stall on the Preview network: https://github.com/midnightntwrk/midnight-indexer/issues/1365

> As reported in the issue, the key is that the timestamp for the transaction is 3s ahead of the block time. The ledger will fail a transaction with `OutOfDustValidityWindow` if the `ctime` of the transaction is `> tblock` in the `block_context`.
> 
> The node should be setting the block context `tblock` == the block time.
> 
> When a transaction enters the mempool, we run a validity check on the transaction. If the check passes, we cache the validity result with the cache key == `(tx_hash, ledger_state_key)`
> 
> The problem is - when a transaction enters the mempool, we bump `tblock` to be (by default) two slots ahead of the current block time: https://github.com/midnightntwrk/midnight-node/blob/2a2fac3bd9dcc1260b3862f335e77c60dc5ecc84/pallets/midnight/src/lib.rs#L450-L454
> ```rust
> 			block_context.tblock = block_context
> 				.tblock
> 				.saturating_add(slot_duration_secs) // 1 slot
> 				.saturating_add(skipped_slots_margin); // + 1 slot = 2 slots
> ```
> 
> So when it comes to including that transaction in the block, we first check the cache - if we find a matching key (`(tx_hash, ledger_state_key)`), we skip the well-formed check, and just apply the transaction. In practice, only the first transaction in the block hits this cache, because the `ledger_state_key` moves on after a transaction is applied.
> 
> So this is a bug with the node - we can fix this in the indexer by using the two-slots-ahead tblock time to the first transaction

What is unknown at the moment: **Does this cause the node to fail to sync with the chain too?**
  - Re-syncing preview to find out: https://github.com/shieldedtech/shielded-sre/issues/432

### Steps to reproduce

- Construct a transaction with `tblock` set to >6s after the current block time
- Submit it to the node
- Attempt to index this chain with the indexer

### Expected behavior

Indexer indexes successfully

### Actual behavior

Indexer fails to import block

PrazwalR:
Fixes #1924. The strict transaction-validation cache in `midnight-node-ledger`
was keyed only by `(state_hash, tx_hash)`, but the cached `well_formed` verdict
also depends on `tblock` — a transaction is malformed with
`OutOfDustValidityWindow` when its `ctime > tblock`.

`pallet_midnight`'s mempool `validate_unsigned` path deliberately inflates
`tblock` by up to `MaxSkippedSlots + 1` slots (≈2 slots) so soon-to-be-valid
transactions aren't dropped from the pool. Because the strict cache key ignored
`tblock`, that inflated "valid" verdict was then reused at block-application
time (`apply_transaction`) and during `pre_dispatch`, both of which use the
true, lower block-time `tblock`. Only the first transaction in a block hit this
cache (the `state_hash` moves on after each applied tx), so a cache-warm
authoring node would accept a first tx whose `ctime` is up to ~2 slots ahead of
block time, while a cache-cold importer/indexer re-evaluating `well_formed` at
the true block time rejected it with `OutOfDustValidityWindow` — a
non-deterministic, cache-dependent block validity that caused the Preview sync
stall.

The fix adds `tblock` to `StrictTxValidationKey`. Block application and import
now miss the inflated mempool entry and recompute `well_formed` at the true
block time, so authoring, import and indexing agree. The mempool inflation is
preserved (forward-dated txs stay in the pool and are included in a later block
once block time catches up), and the normal case keeps its caching benefit
(`pre_dispatch` and `apply_transaction` share the same block-time key).

Closes #1924

### DIFF
--- a/changes/node/changed/strict-tx-cache-tblock-key.md
+++ b/changes/node/changed/strict-tx-cache-tblock-key.md
@@ -1,0 +1,18 @@
+#ledger
+# Include tblock in the strict transaction-validation cache key
+
+The strict transaction-validation cache in `midnight-node-ledger` was keyed only by
+`(state_hash, tx_hash)`, but the cached `well_formed` verdict also depends on `tblock` (a
+transaction is malformed with `OutOfDustValidityWindow` when its `ctime > tblock`). The mempool
+validation path in `pallet_midnight` deliberately inflates `tblock` by up to
+`MaxSkippedSlots + 1` slots to avoid prematurely dropping soon-to-be-valid transactions; because
+the cache lookup ignored `tblock`, that inflated "valid" verdict then leaked into block
+application, which uses the true, lower block-time `tblock`. As a result the first transaction in
+a block could be accepted by a cache-warm authoring node yet rejected with
+`OutOfDustValidityWindow` by a cache-cold importer or indexer — a consensus divergence that
+caused Preview sync stalls. Adding `tblock` to `StrictTxValidationKey` keeps the mempool-time and
+block-application verdicts separate, so block application and import always evaluate `well_formed`
+at the true block time.
+
+PR: <link to PR>
+Issue: https://github.com/midnightntwrk/midnight-node/issues/1924
--- a/ledger/src/versions/common/mod.rs
+++ b/ledger/src/versions/common/mod.rs
@@ -97,6 +97,13 @@ pub const MINT_COINS_DOMAIN_SEPARATOR: &[u8; 10] = b"mint_coins";
 pub struct StrictTxValidationKey {
 	state_hash: Hash,
 	tx_hash: Hash,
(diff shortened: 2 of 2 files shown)


gilescope:
Given that the state key changes after each application, only the first tx in a block will get cached, so it seems simpler to remove the caching. Oscar has proposed this approach here: https://github.com/midnightntwrk/midnight-node/pull/1932

gilescope:
https://github.com/midnightntwrk/midnight-node/pull/1947 is due to land and will close this issue. Rest assured caching will come back later in an even more efficient form. Thank you for the PR.

PR #1939, opened 2026-07-28
fix: filter db-sync observation queries by tx.valid_contract

MicroProofs:
# Overview

cardano-db-sync stores a failed script transaction's actual ledger effects in the
regular tables: consumed collateral becomes `tx_in` rows and the collateral-return
output becomes a regular `tx_out` row, flagged `tx.valid_contract = false`.

This PR adds `tx.valid_contract = true` filtering to every query that interprets
outputs at watched script addresses:

- cNIGHT registrations, with the symmetric filter on deregistrations (a UTxO
  ignored at creation is also ignored when spent)
- governance body UTxO (council / technical committee) and genesis auth-script
  verification
- committee-selection token datum (`get_token_utxo_for_epoch`, both the midnight
  candidates data source and the vendored partner-chains copy)
- bridge transfers (`get_bridge_txs`) — a failing transaction is never observed
  as a transfer; tokens it strands at the bridge address stay locked on purpose

Deliberately unfiltered: cNIGHT asset create/spend tracking, UTxO-set queries,
and the bridge checkpoint hash lookup. Those track real ledger effects, and
db-sync's rows for invalid transactions describe what actually happened on-chain;
filtering them would desync observed balances from the real Cardano UTxO set (and
break checkpoint resolution). Rationale documented in
`primitives/mainchain-follower/README.md`.

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [x] I have included a change file
- [x] No new todos introduced

## 🧪 Testing Evidence

- `partner-chains-db-sync-data-sources`: 79/79 tests pass against a real Postgres
  (bridge, candidates, block-source fixtures) with the filters in place
- `.sqlx` offline metadata regenerated with `cargo sqlx prepare` against a Postgres
  built from the official cardano-db-sync schema migrations; all untouched query
  files reproduced byte-identical
- `SQLX_OFFLINE=true cargo check` passes for `midnight-primitives-mainchain-follower`
  (all features) and `midnight-node`

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [x] Node Client Update
- [ ] Other
- [ ] N/A

Data-source (client-side) change only; no runtime code touched. Note: nodes running
mixed versions could observe different registration/governance data if a failing
script transaction has ever placed outputs at watched addresses on that network.

Assisted-by: Claude:claude-fable-5


changes/node/changed/cnight-registration-valid-contract-filter.md
```diff
@@ -0,0 +1,28 @@
+#cnight-observation #federated-authority #db-sync
+# Exclude phase-2-invalid tx outputs from script-address observation queries
+
+db-sync stores a phase-2-failed transaction's collateral-return output as a
+regular `tx_out` row, so a UTxO created that way at an observed script address
+was previously indistinguishable from a normal output. Observation queries now
+require `tx.valid_contract = true` on the producing tx:
+
+- `get_registrations`, with the symmetric filter in `get_deregistrations` on
```
LGLO: The contract for registration is that we care about UTxOs at given address. Deregistration is consuming registration UTxOs.
Making transaction that fails some contract and at same time deposits proper datum at Registrations Validator address is very convoluted way of registering, but I don't see any reason why we should not consider it a valid registration.
I can't imagine how to consume UTxO at validator address (the act of deregistration) and fail the validator at same time.

MicroProofs: It wouldn't be able to mint the NFT so it wouldn't be a valid registration anyway. So to be on the safer side I filtered it out

LGLO: Registrations don't mint token. Valid registration is UTxO with a datum that contains user signatures of his keys. Deregistration executes validator that should prevent spending if ownership of one of the keys cannot be proven.

MicroProofs: Your are talking the observations for Cnight generating dust right? On the Cardano side a valid registration does mint a token

MicroProofs: I removed the filter anyway

PR #1943, opened 2026-07-29
fix(toolkit): don't report FAILED_TO_FINALIZE when the including block finalized

chrispalaskas:
# Overview

After a tx reached a best block, the toolkit sender's `wait_for_finalized` only matched `InFinalizedBlock` on the `submit_and_watch` subscription and silently swallowed terminal pool events (`Invalid`/`Dropped`/`Error`), subscription errors, and stream end — then reported `FAILED_TO_FINALIZE` without checking whether the already-known including block finalized.

This matters because the node's tx pool can kill the watcher *after* inclusion: on `214eebf`-lineage nodes under load (≥ ~24 concurrent watched txs), a stale pool copy of an already-included tx fails pre-dispatch with `ReplayProtectionViolation(IntentAlreadyExists)` at the submit node's own authoring slot, the pool marks it `Invalid`, and the subscription terminates — seconds before the including block finalizes. The toolkit then reported a landed, finalizing transaction as a finality failure, and batch tooling retried the same spend from cached state, producing false `DustDoubleSpend` (`Custom(196)`) / `NullifierAlreadyPresent` (`Custom(239)`) rejections that masqueraded as ledger double-spend bugs. (Node-side issue with full analysis to be filed separately; link will be added below.)

Changes:
- **Log terminal subscription events** the moment they arrive, with their reason — watcher deaths are now directly observable instead of silent.
- **Direct finality fallback:** on any watcher death (terminal event, stream end, or timeout), poll the node for whether the `InBestBlock` block finalized, via the new reorg-safe `MidnightNodeClient::is_block_finalized` (at or below finalized height *and* canonical at its height).
- **Distinct tag for recovered txs:** fallback-confirmed txs log `FINALIZED_VIA_FALLBACK` with a `reason` field, so pool watcher deaths remain countable per URL (`grep -c` friendly).
- **Minimum fallback window:** the direct check gets at least 10s of polling even when the watcher dies near the end of the 60s finality budget.
- **Sibling fix in `wait_for_best_block`:** `InFinalizedBlock` is now a success arm (skipping the finality wait) instead of being recorded as `last_status` and eventually reported as `FAILED_TO_REACH_BEST_BLOCK` if the subscription coalesces straight to finalization.
- `SenderError::FailedToFinalize` and the `FAILED_TO_FINALIZE` log line now carry the underlying reason.

## 🗹 TODO before merging

- [ ] Add node-repo issue link to the change file and Links section once filed
- [ ] Ready

## 📌 Submission Checklist

- [x] Changes are backward-compatible (or flagged if breaking) — log output gains new tags/fields; `SenderError::FailedToFinalize` gains a `reason` field (internal toolkit API)
- [ ] Self-reviewed the diff
- [x] I have included a change file, or skipped for this reason: `changes/toolkit/changed/sender-finality-fallback.md`
- [x] No new todos introduced

## 🧪 Testing Evidence

Validated on perfnet (8 validators + 1 RPC across 4 AWS regions, 6 s blocks) with an instrumented toolkit build of this branch: two identical 24-worker register-dust + funding-cascade runs of 500 wallets (1000 watched txs per leg), one submit node per leg:

| Submit node version | Watched txs | Watcher killed after inclusion (block finalized anyway) | True finality failures |
|---|---|---|---|
| `2.1.0-214eebf` | 1000 | **85 (8.5%)** — all `pool reported Invalid: Extrinsic marked as invalid` after `InBestBlock`, all recovered as `FINALIZED_VIA_FALLBACK` | 0 |
| `2.1.0-8fa07df` | 1000 | 0 | 0 |

Every fallback-recovered tx was confirmed applied on all 8 nodes and never reverted; chain finality stayed ≤ 3 blocks behind best with zero reorgs throughout. Before this fix, the same failure mode caused downstream retries to be rejected with `DustDoubleSpend`/`NullifierAlreadyPresent`.

Also: `cargo check`, `cargo clippy`, `cargo fmt` clean on `midnight-node-toolkit`.

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [x] Other: toolkit (client tooling) change only — no node/runtime impact, safe to roll out independently.
- [ ] N/A

## Links

Node-side issue (tx pool invalidates already-included txs and kills `submit_and_watch` subscriptions before finality): to be filed — link will be added here.

### DIFF
--- a/changes/toolkit/changed/sender-finality-fallback.md
+++ b/changes/toolkit/changed/sender-finality-fallback.md
@@ -1,0 +1,36 @@
+#toolkit
+# Fix sender reporting `FAILED_TO_FINALIZE` for txs whose including block finalized
+
+After a tx reached a best block, `wait_for_finalized` only matched
+`InFinalizedBlock` on the `submit_and_watch` subscription and silently
+swallowed terminal pool events (`Invalid`/`Dropped`/`Error`),
+subscription errors, and stream end — then reported
+`FAILED_TO_FINALIZE` without checking whether the already-known
+including block finalized. When the submit node's tx pool kills the
+watcher after inclusion (e.g. a stale pool copy of an already-included
+tx failing pre-dispatch with `IntentAlreadyExists` at the node's own
+authoring slot), a landed, finalizing transaction was reported as a
+finality failure. Downstream tooling then retried the same spend from
+cached state, producing false `DustDoubleSpend` / `NullifierAlreadyPresent`
+rejections.
+
+- Terminal subscription events are logged the moment they arrive, with
+  their reason.
+- On any watcher death (terminal event, stream end, or timeout), the
+  sender falls back to polling the node directly for whether the
+  `InBestBlock` block finalized, via the new reorg-safe
+  `MidnightNodeClient::is_block_finalized` (at or below finalized
+  height and canonical at its height).
+- Fallback-confirmed txs log `FINALIZED_VIA_FALLBACK` with a `reason`
+  field (instead of plain `FINALIZED`), so pool watcher deaths remain
+  countable per URL.
+- The fallback gets a minimum 10s polling window even when the watcher
+  dies near the end of the 60s finality budget.
+- Sibling fix in `wait_for_best_block`: `InFinalizedBlock` is now a
+  success arm (skipping the finality wait) instead of being recorded as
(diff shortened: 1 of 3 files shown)


ozgb (approved):
LGTM

PR #1945, opened 2026-07-29
POC: Acropolis Now

gilescope:
# Overview

Don't get your hopes up. This is an early POC dreaming up what a direct acropolis integration into midnight-node might look like as an option to reduce SPO complexities.

Rather than going through a GRPC interface to a co-located acropolis process, the serialisation / deserialisation + local network hop is skipped by node talking directly to the acropolis crates.

Links:

https://github.com/midnightntwrk/midnight-improvement-proposals/pull/69


gilescope:
Seems Dolos might be more likely to be long term supported rather than this.

Issue #1464, opened 2026-05-06
Remove support for separate parity-db instances


ozgb:
_Originally posted by @gilescope in https://github.com/midnightntwrk/midnight-node/pull/1451#pullrequestreview-4229940175_
            

Once the unified db is well tested and becomes the default, we should remove support for separate instances of parity-db for substrate and ledger storage. This is to reduce code maintenance and support burden.

DarrenOliveiro-Priestnall:
Can this be moved back to Q4?

gilescope:
Yes, I'd like to land it with the garbage collection (E.g. https://github.com/midnightntwrk/midnight-node/pull/1991 )

PR #1948, opened 2026-07-29
feat(node): migrate to unified storage in place

ozgb:
# Overview

`storage_separation = "unified"` keeps the Midnight ledger and Substrate storage items in one ParityDb instance, so the two cannot disagree when the node process is killed mid-write. It has been opt-in since #1278, largely because turning it on meant deleting the chain data and resyncing from genesis. This PR removes that cost and then flips the default.

The way it does this is via appending the ledger db columns to the existing substrate parity-db. This is possible because under `separate` mode, both ledger db columns are btree indexed, so can be iterated over to copy the data.

A step toward #1464, which asks for `separate` to be removed once `unified` is the default. Not closing it — dropping the mode is a separate change.

**Breaking, for one case:** a `unified` database written before commit 2 will refuse to start, because ledger nodes already stored outside a btree cannot be re-indexed in place. Those databases have to be resynced. `separate` databases — which is what deployed nodes run today — are unaffected and migrate in place. `unified` -> `separate` remains unsupported.

**NOTE:** This PR **does not make `unified` the default** - it only adds code that migrates `separate` -> `unified` in place

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] No new todos introduced

Three change files under `changes/node/changed/`: the in-place migration, the btree index and column layout, and the default flip.

## 🧪 Testing Evidence

- `node/tests/storage_migration.rs` — fresh `Separate` open, `Separate` -> `Unified` migration, resumption from an interrupted import, refusal when the marker is present but the source database is gone, and refusal of a pre-btree-index `unified` layout.
- Unit tests in `node/src/backend/separate_to_unified.rs` cover the column-layout matching and the metadata rewrite.
- `scripts/tests/storage-separation-migration-e2e.sh` (`just storage-separation-migration-e2e`, and the `storage_separation migration E2E` CI job) starts a node with `STORAGE_SEPARATION=separate`, produces blocks, restarts the same data directory with `unified`, and asserts the node continues from where it left off. Verified locally against a Docker node image.
- `cargo check -p midnight-node --tests` for the default flip.

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [x] Node Client Update
- [ ] Other:
- [ ] N/A

Client-only — no runtime or state change. The storage layout is local to each node, so nodes can be upgraded one at a time; each one migrates its own database on its next restart. Roll back by setting `storage_separation = "separate"` before the first restart on the new binary, since after the migration a rollback needs a resync.

## Links

- Related: #1464 (remove support for separate parity-db instances)
- Builds on #1278 (the original `storage_separation` option)


node/src/backend/separate_to_unified.rs
```diff
@@ -0,0 +1,516 @@
...
+fn separate_layout_metadata(paritydb_path: &Path) -> Result<Option<([u8; 32], u32)>, Error> {
+	let Some(metadata) = parity_db::Options::load_metadata(paritydb_path)? else {
+		return Ok(None);
+	};
+	let separate =
+		super::custom_parity_db::column_options(paritydb_path, StorageSeparation::Separate);
+	let unified =
+		super::custom_parity_db::column_options(paritydb_path, StorageSeparation::Unified);
+	if metadata.columns.len() != unified.columns.len() {
+		return Ok(None);
```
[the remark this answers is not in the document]
ozgb: fixed in https://github.com/midnightntwrk/midnight-node/pull/1948/commits/f8db637c315c21686cc13aa4a19b5bf1039727d2

changes/node/changed/storage-separation-default-unified.md
```diff
@@ -0,0 +1,11 @@
+#storage
+# `storage_separation` now defaults to `unified`
+
+`res/cfg/default.toml` sets `storage_separation = "unified"`, so every network preset that does not override it now keeps Midnight Ledger and Substrate storage items in a single ParityDb instance instead of two. This is the mode that avoids cross-instance data-integrity errors when the node process is terminated unexpectedly.
+
+Existing nodes do not need a resync. The first start-up after the upgrade folds `<base-path>/ledger_storage/` into the shared ParityDb in place and renames the source to `<base-path>/ledger_storage.migrated/` — expect one extra pass over ledger storage on that start, and keep the migrated directory until the node is healthy. A node that was already running `unified` from a build before the ledger columns were btree-indexed is the one case that still needs a resync.
+
+Operators who want the previous behaviour can set `storage_separation = "separate"` explicitly (TOML key or `STORAGE_SEPARATION=separate`). Doing so on a database that has already been migrated to `unified` requires a resync — `unified` -> `separate` is not supported.
```
LGLO: This seems like a choice. Do we want to suggest users they have a choice here or guide them to get into `unified` state?

scripts/tests/storage-separation-migration-e2e.sh
```diff
@@ -0,0 +1,213 @@
...
+fi
+
+# The chain continued rather than starting over.
+CHECKPOINT_HASH_AFTER=$(rpc chain_getBlockHash "[${BLOCKS_BEFORE_SWITCH}]" | jq -r '.')
+[ "$CHECKPOINT_HASH_AFTER" = "$CHECKPOINT_HASH" ] ||
+  fail "block ${BLOCKS_BEFORE_SWITCH} changed across the migration: ${CHECKPOINT_HASH} -> ${CHECKPOINT_HASH_AFTER}"
+
+# And the ledger state behind it came across intact. A resync from genesis, or a
+# half-copied ledger, would not reproduce this root.
+LEDGER_ROOT_AFTER=$(rpc midnight_ledgerStateRoot "[\"${CHECKPOINT_HASH}\"]")
```
LGLO, line 179: Why resync from genesis makes different Midnight ledger state root?

LGLO (dismissed):

changes/node/changed/storage-separation-default-unified.md
ozgb: I think guide them to `unified` state - we will remove the `separate` option altogether once this version has been rolled out

scripts/tests/storage-separation-migration-e2e.sh
ozgb: It doesn't - the comment is wrong, will fix

gilescope:
I think before we merge this PR and flip the default we should tackle the garbage collection of the ledger. (if this PR makes that PR easier then by all means we can land this)

ozgb:
> I think before we merge this PR and flip the default we should tackle the garbage collection of the ledger. (if this PR makes that PR easier then by all means we can land this)

These two changes are orthogonal imo - this PR changes where the storage lives rather than what is stored
Ledger GC is agnostic to whether the storage is unified or separate

Suggest we treat them separately and merge this one as soon as it's approved - this PR doesn't change the default, it just adds the ability to switch mode to unified without re-syncing

Issue #1895, opened 2026-07-17
[Aura→BABE] Committee rotation proceeds after session key registration failure


Klapeyron:
# Committee session key registration silently drops validators

| Field | Value |
|---|---|
| **Risk** | High |
| **Confidence** | High |
| **Likelihood** | Medium |
| **Grounding** | Runtime review (hybrid) |

## Summary

`register_committee_keys` ignores errors from `SessionInterface::set_keys`. When a duplicate or stale `KeyOwner` entry causes upstream `pallet_session` to reject registration, `new_session` still returns the full committee. Downstream, `pallet_session` omits validators without registered keys from the active authority set.

## What happens

1. Committee rotation calls `register_committee_keys` for each member.
2. `SessionInterface::set_keys(account, keys)` fails (e.g. `DuplicatedKey` when a session key is already owned by another validator).
3. The failure is logged and ignored.
4. `new_session` / `new_session_genesis` return **all** committee account IDs regardless.
5. When the session starts, `pallet_session` queue construction skips validators for which `NextKeys` is absent.
6. Aura/GRANDPA run with a partial or unexpectedly empty authority set.

## Root cause

The registration loop treats `set_keys` as best-effort. It does not:

- abort rotation on failure,
- reject the committee, or
- verify that every returned validator has matching keys in `pallet_session`.

## Why it matters

Aura/GRANDPA can omit an intended committee member or produce an unexpectedly empty/partial authority set, risking **liveness** and **finality**.

Issue #1742, opened 2026-06-23
[Aura→BABE] RT-03 — SessionKeys babe field + maybe_from aura fallback


DarrenOliveiro-Priestnall:
**Workstream:** Runtime / Pallets (midnight-node-runtime)  
**Part of:** Aura→BABE consensus migration — [MIP draft](https://github.com/midnightntwrk/midnight-improvement-proposals/blob/docs/mip-aura-to-babe-migration/mips/mip-xxx-aura-to-babe-migration.md)  
**First-pass sizing (to confirm with team):** 4–6 person-days

## Summary
Extend `SessionKeys` with a `babe` field via phantom holder and implement the `maybe_from` resolution with aura-sr25519 fallback.

## Scope
- Add `babe` key field (carrier per SPK-02 outcome)
- `maybe_from`: resolve distinct babe key if registered, otherwise derive from the validator's aura sr25519 key
- Keep unregistered candidates valid during the adoption window (transitional fallback)

## Acceptance criteria
- Requirement (2.9): all authorities resolve to usable babe keys; malformed/absent keys fall through to aura fallback

## Dependencies
- Blocked by SPK-02 (session-key carrier decision)

---
_First-pass breakdown derived from the MIP Implementation section. Sizing and implementation plan to be reviewed and confirmed by the team._

Klapeyron:
Fallback support was implemented as part of associated PRs, babe key was intentionally not introduced and will be added in phase 2

PR #1953, opened 2026-07-31
feat(runtime): add the beefy session key and combine the v1-to-v2 and session-key migrations

MicroProofs:
# Overview

Stacked on `feat-aura-to-babe-migration`; two commits.

**Commit 1 — combine the migrations (pallet refactor, per prior review feedback: one migration must handle both the `QueuedCommittee` addition and a `SessionKeys` shape change, and the pallet storage version must never go to 3).**

Replaces the pallet's pair of migrations (the queue-seeding `v2::V1ToV2Migration` and the `FROM`/`TO`-generic `AuthorityKeysMigration`) with a single combined `VersionedMigration<1, 2>` in `migrations::v2` that translates `CurrentCommittee`/`NextCommittee`, seeds `QueuedCommittee` from the translated current committee, and upgrades `pallet_session` key storage in one gated step, ending at storage version 2.

This fixes the base branch's wiring, which ran the seed before the key translation: the seed's typed read decodes old-shaped committee bytes with the current type, fails silently, and seeds an **empty** `QueuedCommittee` — the next rotation would then hand `pallet_session` an empty validator set. The runtime keeps Tomasz's structure and names: `AddBabeToSessionKeysMigration` stays, still guard-gated on `AddBabeSessionKeysMigrated` (genesis-safe; the node-side committee decoders read it); only its inner migration is now the combined versioned step, and the separate seed leaves `SingleBlockMigrations`.

The pallet migration also hardens the placeholder contract: keys of newly added types must be pairwise distinct per validator (a shared placeholder collapses `KeyOwner` onto one validator; the next rotation's `set_keys` fails with `DuplicatedKey` for everyone else, and those failures are currently swallowed, #1895). The try-runtime `post_upgrade` round-trips `KeyOwner` over every key type of the new shape, and a rotation-count test pins activation timing: migrated placeholders serve the first two post-upgrade sessions; registered keys become active at the third rotation, no validator dropped.

**Commit 2 — add the BEEFY session key.**

`opaque::SessionKeys` gains `beefy`, translated by the same cutover migration. Placeholder = the validator's aura bytes behind the invalid SEC1 tag `0x00` (distinct per validator, collides with no real key). Candidate registrations without a BEEFY key fall back to the same derivation, mirroring the base branch's BABE-from-aura fallback. Genesis takes the configured `beefy_pubkey`; dev seeds derive a proper `BeefyId`.

All live networks (devnet, qanet, preview, preprod, mainnet) were verified at on-chain pallet storage version 1 on 2026-08-13 via `state_getStorage` of the storage-version key. Fresh-genesis chains start at version 2 and set the guard at genesis, so the migration no-ops there.

Also includes `cargo fmt` fixes for three node files on the base branch (whitespace only) so the fmt check passes.

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [x] Changes are backward-compatible (or flagged if breaking) — cutover behavior preserved and fixed; enacted only by the upgrade built from the base branch
- [ ] Self-reviewed the diff
- [x] I have included a change file (`changes/runtime/changed/combined-session-migration.md`, `changes/runtime/changed/add-beefy-session-key.md`)
- [x] If the changes introduce a new feature, I have bumped the node minor version — rides the base branch's version handling
- [x] Update documentation (if relevant) — migration module docs updated
- [x] Updated AGENTS.md if build commands, architecture, or workflows changed — N/A
- [x] No new todos introduced (removes the `todo: add the beefy` marker)

## 🧪 Testing Evidence

- `pallet-session-validator-management` tests: 27 passed (v1 start state, seed assertion, re-run no-op, fresh-genesis no-op, `KeyOwner` round-trip over added key types, rotation-count activation timing)
- `midnight-node-runtime` lib tests: 18 passed (includes the base branch's committee-decode tests)
- `midnight-node` lib tests: 151 passed
- `cargo check --workspace` clean; `--features try-runtime` clean
- `cargo fmt --check` clean for all files in this diff

## 🔱 Fork Strategy

- [x] Node Runtime Update — `SessionKeys` shape change plus the cutover migration; enacted by the upgrade built from the base branch. Metadata rebuild needed after merge.
- [ ] Node Client Update
- [ ] Other
- [ ] N/A

## Links

Issue: https://github.com/midnightntwrk/midnight-node/issues/1742

https://claude.ai/code/session_01SxL4B36LFRxeRX8HRyeSTy


runtime/src/lib.rs
```diff
@@ -374,9 +376,11 @@ impl frame_system::Config for Runtime {
 	type MaxConsumers = frame_support::traits::ConstU32<16>;
 	type RuntimeTask = RuntimeTask;
 	type SingleBlockMigrations = (
-		// Initializes the QueuedCommittee storage added in v2
-		pallet_session_validator_management::migrations::v2::V1ToV2Migration<Runtime>,
-		// See migrations::authority_keys when opaque::SessionKeys changes shape.
+		// v0->v1/v1->v2 migrations were dropped: all networks are at storage version >= 2, and
```
LGLO: `preview`, `preprod`, and `mainnet` are not at PartnerChains v2. Are they?

Unfortunately custom migration is required: one that would be able to go from V1 to V2 together with session keys migration. `pallet_session_validator_management::migrations::v2::V1ToV2Migration<Runtime>` was not able to do it. Careful combination of both migrations is required here.

runtime/src/migrations.rs
```diff
@@ -54,7 +55,22 @@ pub mod authority_keys {
 
 	impl From<LegacySessionKeys> for SessionKeys {
 		fn from(old: LegacySessionKeys) -> Self {
-			SessionKeys { aura: old.aura, grandpa: old.grandpa }
+			SessionKeys {
+				aura: old.aura,
+				grandpa: old.grandpa,
+				// No cross-chain id in scope here, so pallet_session key entries get a
+				// placeholder; they are rewritten from the committee at the next rotation.
+				beefy: sp_core::ecdsa::Public::from_raw([0u8; 33]).into(),
```
LGLO: Would beefy recover from a period of time when node cannon find a matching private key and sign blocks?

MicroProofs: Pushing up some changes in a moment

MicroProofs: Yes as long as beefy eventually gets 2/3 vote on a stuck round then it continues until it catches up

pallets/beefy-registration/src/lib.rs
```diff
@@ -0,0 +1,391 @@
...
+		#[pallet::call_index(0)]
+		#[pallet::weight((
+			T::DbWeight::get().reads_writes(3u64 * updates.len() as u64, 3u64 * updates.len() as u64),
+			DispatchClass::Mandatory
+		))]
+		pub fn update_keys(
+			origin: OriginFor<T>,
+			updates: BoundedVec<(T::AuthorityId, Option<BeefyKeyBytes>), T::MaxUpdatesPerBlock>,
+		) -> DispatchResult {
+			ensure_none(origin)?;
```
LGLO: I've seen recommendations to have a storage (killed in in on_finalize) to prevent malicious nodes include multiple calls of None origin and fill up a block.
If this pallet will be removed before opening mainnet for registered candidates, I think it is okay to not add this complexity.

EDIT: I see we are looking for registered candidates keys as well, so this perhaps will stay and should be protected from multiple inclusion.

changes/node/added/beefy-key-tooling.md
```diff
@@ -0,0 +1,29 @@
...
+compressed SEC1 point. Unknown extra keys are not included. Existing
+registrations without a `beef` key can still be uploaded during the rollout,
+but validators must register the key before the Runtime A requirement is
+enabled. This change does not affect permissioned candidate generation at
+genesis.
+
+The `start-node` wizard now warns about a missing `beef` key instead of refusing
+to start. The node also logs a warning when it starts as an authority without
+this key because it cannot participate in BEEFY voting. BEEFY keys are never
+derived from the cross chain key, and the `AuraToBabeMigrationKeystore`
```
LGLO: Local-environment setup (res/local/permissioned-candidates-config.json and keystores in local-environment itself) will require update.

changes/runtime/changed/combined-session-migration.md
```diff
@@ -0,0 +1,15 @@
+#runtime #migration #committee-selection
+# Combine committee v1-to-v2 and session-key migrations
```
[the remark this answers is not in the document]
LGLO: Codex is stubborn and thinks #toolkit marker is for PC toolkit, but it is not.

PR #1965, opened 2026-07-31
fix: narrow tblock_correction to the first tx in a block (backport of #1964)

ozgb:
# Overview

Backport of #1964 onto `release/node-1.0.2`. The `tblock_correction` shipped in #1932 (which is the tip of this base branch, `b3f9e196`) was applied too broadly: **every** historical transaction was verified at `block_context.tblock + offset`. Preview then stalled at **#128536**, unable to import **#128537**:

```
midnight::ledger_v2: Transaction malformed: ... Intent TTL has expired.
                     TTL: Timestamp(1784987084), Current block: Timestamp(1784987088)
runtime::executive:  Transaction(3) failed due to Invalid(Custom(182)).
```

Two things were wrong.

**1. The offset was measured from the wrong base.** During mempool ingress the node verifies a transaction at `ParentTimestamp + slot_duration * (1 + MaxSkippedSlots)` (`pallets/midnight/src/lib.rs:438-455`) — `ParentTimestamp` because during pool validation `pallet_timestamp::Now` still holds the parent's timestamp. The including block's own timestamp is already one slot past the parent's, so adding the offset to *it* double-counts a slot. For #128537: parent = 1784987070, block = 1784987076, and the shipped code produced 1784987088 while the warm strict-cache entry had actually been verified at 1784987082 — under the TTL, and it imports.

The correction is now measured from `BlockContext::last_block_time`, which is already available natively (fed from `ParentTimestamp`, set in `on_initialize` before the timestamp inherent runs), so **no runtime or host-API change** is needed. Ledger 7's `BlockContext` has no such field, hence a version-tolerant `parent_block_time() -> Option<u64>` accessor — `None` on ledger 7, which correctly disables the correction there (ledger-7 blocks predate the issue).

**2. It was applied at every position in a block.** The strict cache could only have hit for the *first* ledger transaction, so that is the only position where a transaction can have been verified ahead of its block's timestamp. The gate is `Ledger::block_fullness == ZERO` — reset by `apply_post_block_update` and only ever accrued by `apply_verified_transaction` / `apply_system_tx`, so zero fullness ⟺ nothing has been applied yet in this block ⟺ `LedgerState` is byte-identical to the parent's post-block state, which is exactly the condition under which the old strict-cache hit could occur. It is free to read and adds no state.

**3. The mempool path is no longer corrected at all** — `validate_unsigned` already skews the block context it passes to the ledger.

`well_formed_tblock` is a total function of `(block_context, is_block_start, config)`: one unconditional rule, evaluated identically on every node, with no try-then-retry branch for consensus to depend on.

Config values are unchanged, but `tblock_correction_offset` is now relative to the **parent** block's timestamp and must equal `slot_duration_secs * (1 + MaxSkippedSlots)`. On this branch `MaxSkippedSlots` defaults to `1` (`pallets/midnight/src/lib.rs:174`) and slot duration is 6s, so `6 * (1 + 1) = 12` — matching the existing `tblock_correction_offset = 12`. Comments in `res/cfg/default.toml` and `MidnightCfg` updated to say so.

## Differences from #1964

Two, both forced by this branch being ledger-8 based rather than ledger-9:

1. The three new pallet tests call `midnight_node_ledger_helpers::ledger_8::extract_tx_with_context` instead of `ledger_9::` — this branch's helpers crate defines only `ledger_7` and `ledger_8`. Folded into the cherry-picked commit so every commit compiles.
2. An extra commit, `test(ledger): enable helpers can-panic for the crate's own tests`. `cargo test -p midnight-node-ledger` did **not** compile on this base (8 errors at `b3f9e196`, pre-existing and unrelated): the crate's own `#[cfg(test)]` modules import `extract_tx_with_context`, gated behind the helpers `can-panic` feature that nothing enabled for its test build. Without this the new `well_formed_tblock` unit tests could not run at all. `main` carries the same dev-dependency fix, from a commit outside the two backported here.

## 🗹 TODO before merging

- [ ] **Preview re-sync past #128536** — the real check. Not run on this branch; see Testing Evidence.
- [ ] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

Run on this branch, at `064725b9`:

```
cargo test -p midnight-node-ledger -p pallet-midnight   # 89 passed, 2 ignored (4 suites)
cargo test -p midnight-node-ledger well_formed_tblock   # 8 passed
cargo test -p pallet-midnight tblock_correction         # 3 passed
cargo check -p pallet-midnight -p midnight-node --tests # clean
cargo clippy -p midnight-node-ledger -p pallet-midnight -p midnight-node --all-targets  # clean
cargo fmt --all --check                                 # clean
```

The ledger code is compiled twice here (ledger 7/8, vs three times on `main`), so the new unit tests run once per instantiation — the `parent_block_time()` accessor is what keeps the ledger-7 build compiling and the ledger-7 assertions expect no correction.

**New unit tests** — `ledger/src/versions/common/mod.rs`, on `well_formed_tblock` directly:
- corrects the first tx of a historical block to `parent + offset`, and asserts that value is *not* the block's own `tblock` (ledger 7: no correction);
- uncorrected with no correction configured;
- uncorrected at and after `disable_after` (cutoff is exclusive);
- uncorrected once a tx has been applied in the block.

`ledger/src/versions/common/api/ledger.rs` — `is_block_start` follows `block_fullness` through the real apply path: true on a fresh block, false after `apply_verified_transaction`, true again after `post_block_update`.

**New pallet tests** — `pallets/midnight/src/tests.rs`, covering the wiring end-to-end by registering `TBlockCorrectionExt` on the test externalities (the harness registers no extensions today, so `tblock_correction` was `None` in all existing tests). They drive `on_initialize` → timestamp inherent in order so `ParentTimestamp` is real, and use a zero offset so the corrected timestamp *is* the parent's, isolating the base:
- the first tx of a block is verified at the parent's timestamp — the identical block is accepted without the extension and rejected with it;
- a block at `disable_after` is not corrected;
- `validate_unsigned` accepts while `send_mn_transaction` rejects, in the same externalities with the same correction registered.

The mutation-checking of these tests was done on #1964; the test bodies are unchanged here apart from the `ledger_8` retarget.

**Not yet done:** the preview re-sync. Success is importing #128537 and continuing, with `Intent TTL has expired. TTL: Timestamp(1784987084)` gone and 1784987082 used instead.

## 🔱 Fork Strategy

Node client only — `well_formed_tblock` lives in the native ledger bridge and reads the parent timestamp already present in `BlockContext`. No runtime change, no host-API change, no state or storage-layout change, so no metadata or genesis rebuild.

- [ ] Node Runtime Update
- [x] Node Client Update
- [ ] Other:
- [ ] N/A

## Links

Backport of #1964
Follows up #1932
Relates to #1924

Assisted-by: Claude:claude-opus-5

### DIFF
--- a/ledger/helpers/src/versions/common/types.rs
+++ b/ledger/helpers/src/versions/common/types.rs
@@ -323,6 +323,15 @@ impl<D: DB> StorableSyntheticCost<D> {
 			_marker: PhantomData,
 		}
 	}
+
+	/// True when every cost component is zero, i.e. this is [`Self::zero`].
+	pub fn is_zero(&self) -> bool {
+		self.read_time == 0
+			&& self.compute_time == 0
+			&& self.block_usage == 0
+			&& self.bytes_written == 0
+			&& self.bytes_churned == 0
+	}
 }
 
 impl<D: DB> From<SyntheticCost> for StorableSyntheticCost<D> {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9983,6 +9983,7 @@ dependencies = [
  "midnight-node-ledger-helpers",
  "midnight-node-res",
  "midnight-primitives",
+ "midnight-primitives-ledger",
  "pallet-timestamp",
  "parity-scale-codec",
  "scale-info",
--- a/changes/node/changed/fix-tblock-correction-first-tx.md
+++ b/changes/node/changed/fix-tblock-correction-first-tx.md
@@ -1,0 +1,26 @@
+# Narrow the tblock correction to the first transaction of a block
+
+The `tblock_correction` shipped in #1932 was applied too broadly: every historical transaction
+was verified at `block_context.tblock + offset`. Preview stalled at #128536, unable to import
+#128537, whose transaction has an intent TTL of `1784987084` — two slots earlier than the
+`1784987088` that the correction produced.
+
+The offset was measured from the wrong base. During mempool ingress the node verifies a
+transaction at `ParentTimestamp + slot_duration * (1 + MaxSkippedSlots)`, and the including
+block's own timestamp is already one slot past the parent's, so adding the offset to it
+double-counts a slot. The correction is now measured from the parent block's timestamp
+(`BlockContext::last_block_time`), reproducing exactly the timestamp that the producing node's
+warm strict-cache entry was verified at — `1784987082` for the block above, which imports.
+
+The correction is also now limited to the *first* ledger transaction of a block (no user or
+system transaction applied yet, so the ledger state is still the parent's post-block state).
+That is the only position where the producing node's strict cache could have hit, so it is the
+only position where a transaction can have been verified ahead of its block's timestamp. It is
+no longer applied on the mempool path at all, where `validate_unsigned` already skews the block
+context it passes to the ledger.
+
+The config values are unchanged, but `tblock_correction_offset` is now relative to the parent
+block's timestamp and must equal `slot_duration_secs * (1 + MaxSkippedSlots)` (`6 * 2 = 12`).
+
+PR:
+Issue: https://github.com/midnightntwrk/midnight-node/issues/1924
--- a/changes/node/changed/fix-tblock-correction-first-tx.md
+++ b/changes/node/changed/fix-tblock-correction-first-tx.md
@@ -22,5 +22,5 @@ context it passes to the ledger.
 The config values are unchanged, but `tblock_correction_offset` is now relative to the parent
 block's timestamp and must equal `slot_duration_secs * (1 + MaxSkippedSlots)` (`6 * 2 = 12`).
 
-PR:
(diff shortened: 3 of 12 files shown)


ledger/helpers/src/versions/common/types.rs
```diff
@@ -323,6 +323,15 @@ impl<D: DB> StorableSyntheticCost<D> {
 			_marker: PhantomData,
 		}
 	}
+
+	/// True when every cost component is zero, i.e. this is [`Self::zero`].
+	pub fn is_zero(&self) -> bool {
```
Klapeyron, line 328: consider structural dispatch like `let Self { read_time, .., _marker } = self` to protect agains accidental modification of the source struct without adjusting this place

Klapeyron (approved):

PR #1900, opened 2026-07-21
chore: update system_version to 3

Klapeyron:
## Overview

Bump `RuntimeVersion::system_version` from `1` to `3` in the Midnight runtime (and partner-chains demo runtime).
This opts the runtime into the FRAME pending-code upgrade path from [RFC-123](https://github.com/polkadot-fellows/RFCs/pull/123) / [paritytech/polkadot-sdk#6029](https://github.com/paritytech/polkadot-sdk/pull/6029) [paritytech/polkadot-sdk#64](https://github.com/paritytech/polkadot-sdk/issues/64): after a runtime upgrade (`set_code`), off-chain Runtime API calls must not run the new Wasm against unmigrated storage.
Without this bump, nodes keep the old upgrade behavior even on a modern polkadot-sdk.

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [ ] All commits are signed off (`git commit -s`) for the DCO
- [ ] Changes are backward-compatible (or flagged if breaking)
- [ ] Pull request description explains why the change is needed
- [ ] Self-reviewed the diff
- [ ] I have included a change file, or skipped for this reason: 
- [ ] If the changes introduce a new feature, I have bumped the node minor version
- [ ] Update documentation (if relevant)
- [ ] Updated AGENTS.md if build commands, architecture, or workflows changed
- [ ] No new todos introduced

## 🧪 Testing Evidence

Please describe any additional testing aside from CI:

- [ ] Additional tests are provided (if possible)

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A

PR #1982, opened 2026-08-04
fix(rpc): serve ledger state reads at the ledger-hardfork set_code block

ozgb:
# Overview

At the block where a runtime upgrade `set_code` is run, storage migrations have not yet been applied. This leads to a mis-match between runtime code and node state - the runtime API attempts to load pre-migration state using new code paths, causing errors.

The indexer code already keeps track of this code/data drift: https://github.com/midnightntwrk/midnight-indexer/blob/55ec85cc06989e9c14c70c5a76edeb7e04b7ba24/chain-indexer/src/infra/subxt_node.rs#L202-L211

This PR fixes three ledger RPC calls for the runtime-upgrade block - we dispatch based on the ledger state key tag to either the ledger 8 or ledger 9 host apis. One of these RPC calls is used by the current indexer implementation, so unless the indexer code changes to avoid this call, it's broken without this fix.

Once we get `system_version: 3` into the runtime, this issue will no longer exist - see https://github.com/midnightntwrk/midnight-node/pull/1900

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [x] Node Client Update
- [ ] Other: 
- [ ] N/A

### DIFF
--- a/changes/node/changed/hardfork-skew-block-rpc-reads.md
+++ b/changes/node/changed/hardfork-skew-block-rpc-reads.md
@@ -25,5 +25,5 @@ Not covered: `chain-indexer` reads `get_zswap_state_root` through a raw `state_c
 API rather than the `midnight_*` RPCs, so it is unaffected by this fix and still needs either
 to switch to `midnight_zswapStateRoot` or its own skip/fallback at that block.
 
-PR: https://github.com/midnightntwrk/midnight-node/pull/TBD
+PR: https://github.com/midnightntwrk/midnight-node/pull/1982
 Issue: https://github.com/midnightntwrk/midnight-node/issues/1959
--- a/changes/node/changed/hardfork-skew-block-rpc-reads.md
+++ b/changes/node/changed/hardfork-skew-block-rpc-reads.md
@@ -1,0 +1,29 @@
+#node #rpc
+# Serve ledger state RPCs at the ledger-hardfork `set_code` block
+
+On a chain that hardforks ledger 8 -> 9 via a governance `set_code`, exactly one historical
+block was permanently unreadable through `midnight_zswapStateRoot`, `midnight_ledgerStateRoot`
+and `midnight_contractState` (`-32602`, underlying `Deserialization(TypedArenaKey)`). The
+pre-fork runtime shipped `system_version: 1`, so `frame_system` overwrote `:code` *inside* the
+`set_code` block while pallet-midnight's v8 -> v9 state translation only ran in the next block's
+`initialize_block`. That block's committed state therefore pairs ledger-9 `:code` with a
+ledger-8 `StateKey` forever, and any read at that hash executes ledger-9 WASM against a
+ledger-8 arena root.
+
+The node's RPC layer now reads `Midnight::StateKey` straight from the state backend and checks
+its tagged-serialization header. If the key belongs to the older ledger version, the read is
+served by calling that version's host functions natively against the same process-global arena;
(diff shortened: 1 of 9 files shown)


changes/node/changed/hardfork-skew-block-rpc-reads.md
```diff
@@ -0,0 +1,29 @@
...
+served by calling that version's host functions natively against the same process-global arena;
+otherwise the existing runtime-API path is used unchanged. The `StateKey` tag is the signal
+rather than the pallet storage version, because the 2.0.0 runtime already ran ledger 9 while
+still reporting pallet-midnight storage version 1.
+
+Behaviour note: at the `set_code` block `midnight_ledgerStateRoot` now returns a **v8-tagged**
+root — correct, since that block's state *is* v8 — so consumers walking the fork see the tag
+flip at the migration block rather than a hole.
+
+Not covered: `chain-indexer` reads `get_zswap_state_root` through a raw `state_call` runtime
```
Klapeyron, line 24: are we sure about that? For 2.0.0 it looks like it is using `MidnightRuntimeApi_get_zswap_state_root` https://github.com/midnightntwrk/midnight-indexer/blob/9c8b04680b97bf4c9490a1dcc2a3cd86fd883b47/chain-indexer/src/infra/subxt_node/runtimes/v2_0_0.rs#L296

Klapeyron:
> "Once we get system_version: 3 into the runtime, this issue will no longer exist - see #1900"

This remark is only half-true. It prevents future skew blocks (code gets staged in ``:pending_code`). But the skew block from the 8→9 fork will be a committed state — so each sync will have to carry that exception.

> The indexer code already keeps track of this code/data drift: https://github.com/midnightntwrk/midnight-indexer/blob/55ec85cc06989e9c14c70c5a76edeb7e04b7ba24/chain-indexer/src/infra/subxt_node.rs#L202-L211

I am not sure about that handling mentioned there, what indexer does here it indeed detects that case, but only logs abut it. I do not see a proper handling of that situation

One general objection for this changes I have is that it is not addressing the runtime api issue but JSON-RPC only. Also this change will be permanent behavior for our and 3rd party tooling forever once skew block will land

ozgb:
> > "Once we get system_version: 3 into the runtime, this issue will no longer exist - see #1900"
> 
> This remark is only half-true. It prevents future skew blocks (code gets staged in ``:pending_code`). But the skew block from the 8→9 fork will be a committed state — so each sync will have to carry that exception.
> 
> > The indexer code already keeps track of this code/data drift: https://github.com/midnightntwrk/midnight-indexer/blob/55ec85cc06989e9c14c70c5a76edeb7e04b7ba24/chain-indexer/src/infra/subxt_node.rs#L202-L211
> 
> I am not sure about that handling mentioned there, what indexer does here it indeed detects that case, but only logs abut it. I do not see a proper handling of that situation
> 
> One general objection for this changes I have is that it is not addressing the runtime api issue but JSON-RPC only. Also this change will be permanent behavior for our and 3rd party tooling forever once skew block will land

Thanks for taking a look at this - I've created a new PR that applies this workaround to the ledger host API instead - a smaller diff, and it covers the runtime API too: https://github.com/midnightntwrk/midnight-node/pull/1985

Suggest we go with the new PR and close this one

PR #1984, opened 2026-08-04
fix(node): garbage-collect ledger_storage to honour state pruning

### LINKED ISSUE #1983: ledger_storage does not respect pruning: 197G on a pruned validator (paritydb: 6.8G)

Make `ledger_storage` respect block pruning, and reduce its on-disk footprint generally. ~200 GB at block **1,798,390** (for both archive and non archive nodes) seems like a lot of storage — after a conversation with Giles (@gilescope), he says this can be greatly improved.

On a preprod **archive** node (RPC syncbox) at block 1,798,390:

```
18G     ./chains
181G    ./ledger_storage
```

On a preprod **validator** with default pruning (not an archive node — only archive nodes should keep more than 256 blocks):

```
4.0K    ./chains/midnight_preprod/network
6.8G    ./chains/midnight_preprod/paritydb/full
6.8G    ./chains/midnight_preprod/paritydb
28K     ./chains/midnight_preprod/keystore
6.8G    ./chains/midnight_preprod
6.8G    ./chains
197G    ./ledger_storage
```

paritydb handles pruning properly (6.8G on the pruned validator vs 18G archive), but `ledger_storage` does not — it is ~197G on a node that should only retain ~256 blocks of state.

## Expected Benefit

- Pruned nodes (validators especially) get bounded disk usage as configured, instead of unbounded growth.
- Smaller volumes, lower storage cost, and fewer ENOSPC incidents for all node operators; cheaper/faster snapshots and bootstraps.

## Details

- Archive measurements: Shielded preprod syncbox, midnight-node 1.0.1, `--state-pruning archive --blocks-pruning archive`, `STORAGE_SEPARATION=separate`, measured 2026-08-04 at block 1,798,390.
- Validator measurements: Shielded preprod validator, default pruning, measured 2026-08-04.
- Filed on behalf of Shielded SRE; context: shieldedtech/shielded-iac#2016 (volume resize that worked around this on our side).

- Related: #1952 (same root cause, closed 2026-07-30 in favour of the new unified database backend).

Klapeyron:
ledger_storage retained every persisted arena root forever: each ledger host call that mutates state persists the resulting root, and nothing ever unpersisted them. Pruned validators therefore accumulated ~197G of ledger state (vs 6.8G paritydb) on preprod.

Add a background worker that, on non-archive nodes, periodically:

1. collects Midnight::StateKey values for every block whose Substrate state is still retained (canonical pruning window plus non-finalized forks), and
2. runs one time-budgeted slice of the storage crate's incremental mark-and-sweep collector (gc_override_gc_roots) with that set as the complete live root set.

Using an override root set sidesteps persist refcounting (the same root is persisted by both authoring and import, and intermediate per-transaction roots are superseded within a block). Live in-memory objects are protected by the backend's own root tracking, and any concurrent persist forces a root rescan before further sweeping.

Slices abort conservatively if the live window contains pre-ledger-8 state keys (initial sync through the ledger-7 era), and archive nodes skip the worker entirely.

Closes #1983

Assisted-by: Claude:claude-fable-5

# Overview

## 🗹 TODO before merging

- [ ] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [ ] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A

### DIFF
--- a/changes/node/changed/ledger-storage-pruning-gc.md
+++ b/changes/node/changed/ledger-storage-pruning-gc.md
@@ -1,0 +1,12 @@
+#ledger #storage #node
+# Garbage-collect ledger_storage to honour state pruning
+
+`ledger_storage` previously retained every persisted arena root forever, so a
+pruned validator's ledger DB grew without bound (~197G vs ~7G paritydb on
+preprod). A background worker now collects `Midnight::StateKey` values for the
+blocks still covered by Substrate `--state-pruning` (plus non-finalized forks)
+and runs the storage crate's incremental GC against that live root set.
+
+Archive nodes skip the worker and keep full ledger history.
+
+Issue: https://github.com/midnightntwrk/midnight-node/issues/1983
--- a/ledger/src/gc.rs
+++ b/ledger/src/gc.rs
@@ -1,0 +1,267 @@
+// This file is part of midnight-node.
+// Copyright (C) Midnight Foundation
+// SPDX-License-Identifier: Apache-2.0
+// Licensed under the Apache License, Version 2.0 (the "License");
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Garbage collection for the ledger storage arena.
+//!
+//! Every ledger host call that mutates state (`apply_transaction`,
+//! `apply_system_transaction`, `apply_post_block_update`, ...) persists the
+//! resulting state root in the storage arena, and nothing ever unpersists
+//! them. Left alone, the arena therefore accumulates every per-transaction
+//! and per-block state root since genesis — including roots from discarded
+//! block proposals — regardless of the node's pruning configuration
+//! (see <https://github.com/midnightntwrk/midnight-node/issues/1983>).
+//!
+//! This module exposes [`collect_garbage`], which runs the storage crate's
+//! incremental, time-budgeted mark-and-sweep collector
+//! ([`StorageBackend::gc_override_gc_roots`]) against an explicit set of
+//! *live* state keys supplied by the caller. Anything unreachable from those
+//! roots (or from in-flight, in-memory objects, which the backend tracks
+//! itself via `live_inserts`) is deleted, and any persisted root not in the
+//! set is effectively unpersisted. The node derives the live set from the
+//! `Midnight::StateKey` storage value of every block whose Substrate state
+//! is still retained, so the arena's retention exactly matches the node's
+//! state-pruning window.
+//!
+//! Using an override root set (rather than matched `unpersist` calls)
+//! sidesteps the persist reference-counting problem: the same root may be
+//! persisted several times (block authoring and import both execute the
+//! block; intermediate per-transaction roots are superseded within the same
+//! block) and would otherwise need an exactly matched number of unpersists.
+//!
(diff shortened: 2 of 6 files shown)


scottbuckel:
Reviewed this against the pinned `midnight-storage-core` 1.2.0 source (`gc-v1`), the pallet, and the host-call path. The design intent is right and most of the safety claims check out in the code (`live_inserts` merged every slice, `force_rescan()` on every flushed root-count change, incremental resume). One race I think needs addressing before this runs on validators, plus one sizing question.

**Race: roots persisted between StateKey collection and the GC slice are unprotected.**

The PR rationale is "live in-memory objects are protected by the backend's own root tracking, and any concurrent persist forces a root rescan". Both are true individually, but they don't compose to cover the collection window:

1. The host loads state per call via `get_lazy` and drops the `Sp` when the call returns (`ledger/src/versions/common/mod.rs`, `get_ledger`), and `flush_storage` flushes each block. So between host calls the tip root is protected *only* by its DB root_count — which `gc_override_gc_roots` deliberately overrides.
2. `collect_live_state_keys` runs *before* the backend lock is taken. A block imported in that window persists+flushes its root, setting `rescan = true` — and the next slice **consumes** that flag by re-initialising `last_roots` from the already-stale override set (`GcState::run`, the `if self.rescan` branch), which lacks the new tip. The sweep in that same slice can cull the tip's unique nodes.
3. On non-rescan ticks the `FnOnce` override is never invoked at all, so freshly collected keys are silently discarded. If the chain goes idle right after an unluckily-timed block (finality stall, quiet devnet), the stale mark set persists for the remainder of the sweep epoch and the cull becomes near-deterministic rather than a narrow race.

Impact: losing tip-state nodes makes the next `get_ledger` fail → the node can't author/import until a resync. Small per-tick probability, but ~2 880 ticks/day on every pruned node.

Suggestion: keep a small LRU (even 4–8 entries) of recently returned ledger-state `Sp`s in the host. `live_inserts` then always covers recent roots (including fork tips) and the race disappears by construction. Alternatives: re-read the tip StateKey inside the locked section, or extend the storage crate to auto-protect roots persisted since the epoch's rescan.

**Sizing question: `mark_set`/`grey_set` are O(all reachable nodes), unbounded by the time budget.**

The mark set holds every live arena-node hash in RAM. If the tip state is tens of millions of nodes, that's GBs of `HashSet` — some validator fleets run 4 GiB instances with memory watchdogs around 3 GiB, which would kill the node mid-epoch. Do we have an estimate of live-node cardinality on preprod/mainnet tip state?

**Smaller items**

- `retention_window` maps `Constrained { max_blocks: None }` to 256; in Substrate that means *unbounded* state retention, so GC would delete ledger data for states Substrate still serves. Not CLI-reachable today, but safer to treat as archive/skip.
- A transient `Err` reading a within-window StateKey logs and skips, leaving that block's roots unprotected for the slice; consider aborting the slice (like the ledger-7 path) instead.
- ParityDB doesn't shrink files on delete — existing ~197G nodes will stop growing but won't reclaim disk without a resync. Worth stating in the change file so operators don't conclude it didn't work.
- Fresh pruned syncs still accumulate ledger-7-era garbage forever (the legacy storage crate has no GC) — worth stating explicitly.
- `DEFAULT_INTERVAL`/`DEFAULT_BUDGET` being compile-time constants means no operational tuning; env overrides would help. The 250 ms backend-lock hold every 30 s also stalls host calls — fine as a default, but worth a comment.
- `state_key_storage_key_is_stable` recomputes its own definition, so it can't fail; asserting against the runtime metadata would make it meaningful.

For a state-deletion feature I'd also want soak evidence on a preprod validator (du trend + a canary querying state at the window edge) in Testing Evidence before rollout.

*(AI-assisted review — Claude, on behalf of Shielded SRE.)*

Klapeyron:
Closing in favor of https://github.com/midnightntwrk/midnight-node/pull/1991

PR #2008, opened 2026-08-07
pipe num-validators through local-env cli

justinfrevert:
# Overview

`--num-validators <count>` lets a well-known network fork come up with fewer mock
validators than the network's Compose topology defines, instead of always
materializing the full set. Smaller forks start faster and fit on a laptop, which
is what most fork testing actually needs.

The flag is threaded through every local-env command that brings up or rolls a
forked network: `run`, `image-upgrade`, `full-upgrade`, and the runtime-upgrade
path, plus the `fork-network` CI workflow.

1. When forking from a snapshot the count drives `mock-authorities convert` (how
   many validator keysets get synthesized) and a generated Compose override that
   starts only the selected `node*` services, parking the rest behind a profile
   this tooling never activates.
2. The active selection is persisted in the generated override
   (`x-midnight-mock`), so later restarts can omit the flag and reuse it, and
   `verify-finality` probes the validators that are actually running rather than
   the full configured topology.
3. The count must be a positive integer no larger than `mock.validatorServices`.
   Changing it requires `--from-snapshot` so the authority set and seeds can be
   regenerated; it is rejected on reuse runs, in `--from-genesis` mode, and for
   the standalone `local-env` stack, whose topology and keys are fixed.

Also fixes a GRANDPA finality stall found while testing this: the fork tooling
defaulted to `mock-authorities` images that predate the client-side aux-storage
rewrite, which deadlocks the GRANDPA voter on a >=2.0.0 runtime. Details in
`changes/node/changed/local-env-fork-grandpa-finality-fix.md`. That stall was not
caused by the reduced validator set.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

Unit tests cover the selection resolver, the persisted-selection round trip, the
disabled-service profile, and the stale-override cleanup
(`local-environment/src/lib/mockValidatorSelection.test.ts`).

Manually verified by forking devnet (`2.0.0-rc.3`) from a snapshot with
`--num-validators 3`: all three validators author and finalize, and the client
and runtime GRANDPA set-ids match.

## 🔱 Fork Strategy

- [x] N/A 

## Links

Consumer of this tooling: midnightntwrk/midnight-indexer#1364


ozgb (dismissed):

LGLO (dismissed):

erik-axeljakobsson (dismissed):

changes/node/changed/local-env-fork-grandpa-finality-fix.md
```diff
@@ -0,0 +1,33 @@
...
+full-upgrade modes rely on. The `MOCK_AUTHORITIES_IMAGE` env var and the
+`mock_authorities_tag` workflow input still override the default.
+
+This was not caused by the `--num-validators` feature: the reduced mock set was
+installed correctly on both the runtime authority set and the client GRANDPA
+voter set. Verified by forking devnet (`2.0.0-rc.3`) from a snapshot with
+`--num-validators 3` and `4d8d772`, where all three validators finalize and the
+client and runtime GRANDPA set-ids match.
+
+PR: <link>
```
LGLO: Fix this one please

LGLO (approved):

erik-axeljakobsson (approved):

PR #2048, opened 2026-08-19
chore: stop duplicating binaries and res/ in a second image layer

ozgb:
# Overview

Each binary currently appears **twice** in the shipped images, in two different layers.

The node and toolkit images `COPY` their files as root and then run
`chown -R appuser:appuser …` over the same paths. `chown` rewrites every file it
touches, so buildkit records them all as changed and stores a second full copy in
the `chown` layer.

Fix: copy with `COPY --chown=appuser:appuser` (the `appuser` account moves above the
`COPY`s in both base Dockerfiles so the name resolves), and leave `chown` covering
only the directories that `RUN mkdir` creates.

Two things worth noting beyond the binaries:

- It is not just the binaries. `chown -R … ./res` was duplicating the whole 74M
  `res/` tree in `+node-image` too, and `/toolkit-js` (node_modules) in `+toolkit-image`.
- `chown -R … ./bin` never reached the file it was aimed at: `/bin` is a usr-merge
  symlink and GNU `chown -R` does not traverse symlinks, so `/usr/bin/.envrc` stayed
  root-owned. It is `appuser`-owned now, via the base Dockerfile `COPY`.

`+srtool-build` / `+srtool-info` have the same `chown -R` pattern and are deliberately
left alone: they are build-only targets that are never pushed.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

A/B measured on the real `images/node/Dockerfile` base image with a 133M stand-in
binary in place of the built one, comparing the old and new tails:

```
before:  210MB  RUN chown -R appuser:appuser /midnight-node /node ./bin ./res
         133MB  COPY midnight-node-fake /midnight-node

after:      0B  RUN chown -R appuser:appuser /node
         133MB  COPY --chown=appuser:appuser midnight-node-fake /midnight-node
```

~210MB removed from the node image: the duplicated binary plus the duplicated
`res/` tree. Ownership in the resulting image is unchanged:

```
-rw-r--r--. 1 appuser appuser 133000000 /midnight-node
drwxr-xr-x. 1 appuser appuser           /node
drwxr-xr-x. 1 appuser appuser           /res
-rw-r--r--. 1 appuser appuser      5430 /usr/bin/.envrc   (was root-owned before)
```

The equivalent toolkit-image saving was not measured separately; it is the same
mechanism applied to `/midnight-node-toolkit` and `/toolkit-js`. CI building both
images is the check that the `--chown` copies and the trimmed `chown` still produce
working images.

## 🔱 Fork Strategy

- [x] Other: image build only. No node, runtime, or on-chain behaviour changes; the
      image contents are identical, only the layer they live in changes.

## Links

Follow-up to a review finding on #36 (out of scope there).

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -197,9 +197,11 @@ node-image-minimal:
     USER root
 
     RUN mkdir -p /node
-    COPY +build-node-only/artifacts-$NATIVEARCH/midnight-node /
+    COPY --chown=appuser:appuser +build-node-only/artifacts-$NATIVEARCH/midnight-node /
 
-    RUN chown -R appuser:appuser /midnight-node /node ./bin ./res
+    # Only /node (created above as root) needs fixing: everything else is copied
+    # with --chown, so no `chown -R` rewrites it into a duplicate layer.
+    RUN chown appuser:appuser /node
     SAVE IMAGE localhost/node-minimal:latest
 
 # Grabs metadata.scale file from the latest node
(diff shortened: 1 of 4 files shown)


gilescope (approved):

gilescope:
Nice find!

PR #2080, opened 2026-09-01
fix: split system transaction executor into per-pallet allow-listed traits

ozgb:
# Overview

`MidnightSystemTransactionExecutor::execute_system_transaction` let a caller apply
any serialized `SystemTransaction` variant to the ledger, with no restriction.
The root-origin `send_mn_system_transaction` extrinsic already gated on an
allow-list (only `OverwriteParameters` passes), but the trait path used by
`pallet-cnight-observation` and `pallet-c2m-bridge` bypassed that gate entirely.
Both callers only ever construct a fixed set of transaction kinds today, so this
was latent rather than exploited, but nothing at the trait boundary stopped a
future bug or compromised caller from submitting any other ledger-recognized
variant (e.g. treasury payouts, reserve distribution) through either pallet.

This replaces the single shared trait with per-caller traits, each restricted at
the trait boundary to the system transaction variants that pallet is actually
allowed to construct:

- `MidnightSystemTransactionCNightExecutor` (cnight-observation): only
  `CNightGeneratesDustUpdate`.
- `MidnightSystemTransactionBridgeExecutor` (c2m-bridge): only
  `UnlockToTreasury`, `DistributeReserve`, and
  `DistributeNight(ClaimKind::CardanoBridge, _)`.
- The governance extrinsic path keeps its existing allow-list
  (`OverwriteParameters` only).

Each is backed by a dedicated ledger host function
(`apply_governance_system_transaction`, `apply_cnight_system_transaction`,
`apply_bridge_system_transaction`) that deserializes the transaction once and
checks its allow-list before applying, rather than a separate classifier call
ahead of `apply_system_transaction` (which would deserialize and cross the
WASM host-function boundary twice per call). The previous
`apply_system_transaction` and `is_governance_allowed_system_tx` host functions
are kept in place, unused by new pallet code, since host functions are part of
the node<->runtime ABI and removing one an already-published runtime WASM can
call would break replay of historical blocks.

`runtime/src/lib.rs` needs no changes: `pallet-midnight-system` now implements
both new traits, and the runtime's existing
`type MidnightSystemTransactionExecutor = MidnightSystem;` wiring for both
pallets keeps compiling as-is.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

- `cargo check -p pallet-midnight-system -p pallet-cnight-observation -p pallet-c2m-bridge -p midnight-primitives -p midnight-node-runtime -p midnight-node-ledger`
- `cargo test -p pallet-midnight-system -p pallet-cnight-observation -p pallet-c2m-bridge` — all pass, including 5 new unit tests in `pallet-midnight-system` that construct real system transactions against a real ledger (via `init_storage_paritydb_separate`) and assert each executor's allow-list accepts/rejects the right variants.
- `cargo test -p midnight-node-ledger` (error-code uniqueness tests, to confirm the new `SystemTransactionError::NotAllowedForCaller` variant doesn't collide with an existing or retired wire code).
- `cargo clippy` / `cargo fmt` on all touched crates.
- Full workspace `cargo check --all-targets` — confirms `runtime/src/lib.rs` compiles unmodified.

## 🔱 Fork Strategy

- [x] Node Runtime Update
- [x] Node Client Update
- [ ] Other:
- [ ] N/A

## Links


pallets/midnight-system/src/lib.rs
```diff
@@ -69,6 +77,10 @@ pub mod pallet {
 		ContractNotPresent,
 		#[codec(index = 14)]
 		BeneficiaryNotFound,
+		#[codec(index = 15)]
+		SystemTransactionNotAllowedForCNight,
+		#[codec(index = 16)]
+		SystemTransactionNotAllowedForBridge,
```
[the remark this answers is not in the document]
ozgb: https://github.com/midnightntwrk/midnight-node/pull/2080/commits/272350300ac6ce9fcf0af6dbb698764308dbb85c

ledger/src/ledger_8/types.rs
```diff
@@ -87,6 +87,9 @@ pub enum SystemTransactionError {
 	TreasuryDisabled,
 	MerkleTreeError,
 	UnknownError,
+	/// The caller's `apply_*_system_transaction` guard rejected this transaction's
```
LGLO: This comment has a potential for being stale when we organize code somehow different. I think this variant can be left without it like the others here.

LGLO (dismissed):

ledger/src/ledger_8/types.rs
ozgb: removed here: https://github.com/midnightntwrk/midnight-node/pull/2080/commits/c8ca06817cc44cdffc9faccc97c850133e3c6c4a

LGLO (approved):

PR #1938, opened 2026-07-28
perf(toolkit): faster resilient chain sync and replay, working wallet cache on mainnet

jacek-kurkowski-shielded:
# Overview
Closes #1937

Syncing the toolkit against mainnet was slow and opaque, fragile to connection loss, and re-did all work on every run (#1937). This PR fixes the full path:

- **Fetch**: multi-threaded runtime (the whole pipeline previously shared one core), reconnect-with-backoff on dropped WebSockets, all RPC moved out of the compute stage (which also removes a panic on connection loss), events RPC skipped for timestamp-only blocks, and the job pusher chases the finalized tip so a sync ends at the current head. A 10s heartbeat reports progress, RPC-rate-based ETA and backlog.
- **Replay**: finalized history is verified in proof-erased form, skipping ZK-proof/signature re-verification (~3.5x faster through tx-dense ranges). Correctness is still enforced per block: the locally computed state root is compared against the on-chain `Midnight.StateKey`; any divergence aborts the replay. Validated over 170k mainnet blocks (incl. the 7->8 fork) with zero mismatches. Historical partially-failed txs now log at debug instead of spamming stdout.
- **Wallet-state cache**: snapshots are version-tagged and ledger-8 chains (mainnet today) are fully supported. Previously the cache silently never saved below ledger 9, so every `show-wallet` replayed from genesis (~20 min); a warm rerun now takes seconds. Mid-replay checkpoints (default 5 min, `MN_REPLAY_CHECKPOINT_SECS`) let interrupted replays resume.

Known pre-existing gap (out of scope, noted in #1937): replay has no ledger 8->9 fork transition yet (`fork_context_8_to_9`).
## 🗹 TODO before merging

- [X] Ready

## 📌 Submission Checklist

- [X] All commits are signed off (`git commit -s`) for the DCO
- [X] Changes are backward-compatible (or flagged if breaking)
- [X] Pull request description explains why the change is needed
- [X] Self-reviewed the diff
- [X] I have included a change file, or skipped for this reason: 
- [ ] If the changes introduce a new feature, I have bumped the node minor version
- [X] Update documentation (if relevant)
- [ ] Updated AGENTS.md if build commands, architecture, or workflows changed
- [X] No new todos introduced

## 🧪 Testing Evidence

[...]

Issue #1937, opened 2026-07-28
Toolkit: mainnet sync/replay is slow and opaque; wallet-state cache never persists on ledger-8 chains


jacek-kurkowski-shielded:
### Context & versions

Toolkit from current `main`, syncing Midnight mainnet (`wss://rpc.mainnet.midnight.network`, ~1.9M blocks, specVersion 1000000 = ledger 8) on a laptop over a residential connection.

### Steps to reproduce

1. `midnight-node-toolkit fetch -s wss://rpc.mainnet.midnight.network` with a fresh cache.
2. After sync completes, `midnight-node-toolkit show-wallet --seed <seed> -s wss://rpc.mainnet.midnight.network`.
3. Run the same `show-wallet` command a second time.
4. Optionally interrupt any run (Ctrl-C, laptop sleep, dropped connection) and rerun.

### Expected behavior

- Sync uses available cores and reports progress (rate, ETA).
- Transient connection loss is retried; an interrupted run resumes where it left off.
- The second `show-wallet` run reuses cached wallet state and completes in seconds.

### Actual behavior

- The whole pipeline runs on one core (`new_current_thread` runtime); a full sync takes many hours with long stretches of no output even with `--verbose`.
- A dropped WebSocket permanently breaks the jsonrpsee client; retries reuse the dead client and the run aborts (`RestartNeeded`, worker panics in the compute stage which does hidden RPC calls).
- Replay re-verifies ZK proofs and signatures of finalized history (~19 blocks/s in tx-dense ranges) and prints every historical partially-failed transaction to stdout via `println!`.
- The wallet-state cache silently never saves on pre-ledger-9 chains (mainnet today): `try_save_cache_v2` early-returns with a debug-only log and the snapshot format is ledger-9 only, so every `show-wallet` run replays the full chain (~20 min). There are no mid-replay checkpoints, so any interruption loses all replay work.
- ETA/progress accounting is attached to a starved verification queue and can show 0% forever; the fetch target is snapshotted at startup, so a long sync ends hours behind the tip.

### Additional remarks

Found while investigating mainnet sync UX. A fix covering the above is prepared and will be linked to this issue. Related pre-existing gap, likely worth its own issue: replay has no ledger 8->9 fork transition (`fork_context_8_to_9` does not exist), which will bite when mainnet schedules the ledger-9 fork.

PR #2101, opened 2026-09-07
perf(toolkit): recycle throttled fetch connections

jacek-kurkowski-shielded:
# Overview

Fetch workers now recycle a connection that has been throttled. Found while syncing a public network with the toolkit against its remote RPC endpoint. Follow-up to the sync/replay performance work in #1938.

Against a public RPC endpoint a long-lived WebSocket was observed to drop from ~500 blocks/s to ~20 blocks/s after 10 to 20 minutes at full speed. The drop came with no error, hit all worker connections at once, and a freshly opened connection ran at full speed again. That points at per-connection rate limiting on the endpoint side rather than anything in the node or polkadot-sdk; the toolkit does not try to game a node-side mechanism, it only replaces a connection that the endpoint has slowed down.

A job that fetched at under a quarter of the worker's best observed rate, and took at least 10 seconds, makes the worker reconnect before its next job. The rate is measured over the successful attempt only, so a retried job does not count its failed attempt and backoff as slowness (review feedback).

The runtime metadata cache that was part of the first version of this PR has been dropped here at reviewers' request and will be raised as a separate PR, scoped to the connected chain.

Toolkit only. Node, runtime and cache formats are unchanged.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [ ] If the changes introduce a new feature, I have bumped the node minor version. Toolkit only, node version unchanged.
- [ ] Update documentation (if relevant). Not needed.
- [ ] Updated AGENTS.md if build commands, architecture, or workflows changed. Not needed.
- [x] No new todos introduced

## 🧪 Testing Evidence

`cargo clippy` clean, 36 toolkit fetcher unit tests pass on this branch.

Connection recycling was exercised live on the 1.0.3 line against a 2.53M-block network with 88.8k transactions, together with the metadata cache that is now split out. A cold `show-wallet` completed in 78 minutes with 207 throttle reconnects, each restoring full fetch speed. Warm reruns took 48 to 66 seconds and the wallet output was identical across four runs.

- [ ] Additional tests are provided (if possible). No new tests; the behaviour is network-dependent.

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [x] Other: Toolkit only. No node, runtime or network impact.
- [ ] N/A

## Links

Relates to #1937.
Follow-up to #1938


changes/toolkit/changed/sync-metadata-cache-and-slow-connection-recycle.md
```diff
@@ -0,0 +1,17 @@
...
+# Cache runtime metadata across blocks and recycle throttled fetch connections
+
+Found while syncing a public network against its remote RPC endpoint.
+
+- The subxt client config now caches runtime metadata per spec version, shared by
+  every client in the process. It had no cache, so subxt downloaded the full
+  metadata (~130KB, two runtime calls) for every block; against a remote node fetch
+  ran at ~30 blocks/s with the process idle. With the cache the same sync runs at
+  ~600 blocks/s with 8 fetch workers.
+- Fetch workers recycle a connection that has gone slow. Public RPC endpoints were
```
LGLO: Adding two changes in one PR has a risk that one will be objected.

What is the reason of "observed throttle"? Is there any mechanism in polkadot-sdk it tries to game? I think it demands some more explanation that observed drop in speed.

ozgb:
image

This is an incredible find! wow

changes/toolkit/changed/sync-metadata-cache-and-slow-connection-recycle.md
ozgb: Agreed - the metadata cache is a clear win, would be great to approve that one separately 
@jacek-kurkowski-shielded are you happy to raise a new PR with just the metadata cache fix?

jacek-kurkowski-shielded: @LGLO : The change file now says it is observed on public RPC endpoints with no error and on all connections at once, while a fresh connection is fast again, which points at per-connection rate limiting on the endpoint side rather than anything in polkadot-sdk.

jacek-kurkowski-shielded: @ozgb sure, will do it!

PR #2102, opened 2026-09-07
refactor(cnight): group observation events by Cardano transaction; make row-limited pulls gap-free

ozgb:
# Overview

Ports the two structural halves of #1966 (which targeted the since-reverted sliding-window branch) onto `main`.

**Whole transactions become structural.** `CNightGroupedUtxos` (new, `primitives/mainchain-follower/src/data_source/cnight_grouped.rs`) holds observation events grouped by Cardano transaction; events enter and leave only in whole-transaction units. The ad-hoc sort-and-truncate loop in `get_utxos_up_to_capacity` is gone. Emitting part of a transaction is what makes validators re-derive a different inherent and reject the block, so it is worth having the type refuse to do it rather than each call site remembering not to.

**Row-limited pulls become gap-free.** The four category queries are each `ORDER BY block_no, block_index LIMIT tx_capacity * 64`. If one saturates while another has events beyond its frontier, the old loop saw fewer than `tx_capacity` distinct transactions, concluded "range complete", and set `end = tip` — silently skipping the saturated category's remaining events forever. The merged set is now cut at the earliest saturated category's frontier (dropping the frontier transaction, which the limit may have sliced mid-tx) and the cursor resumes at the cut.

Saturation is judged on raw row count, before the per-row filters. This matters: `get_asset_create_utxos` / `get_asset_spend_utxos` drop non-base holder addresses and the registration queries drop undecodable datums, so a saturated query can arrive with far fewer surviving events than the limit and would otherwise look complete. That also weakens the counting argument (`changes/node/changed/revert-cnight-query-capacity-reduction.md`) that the 64x factor alone kept us safe.

Genesis construction terminated its scan on `end.block_hash == cardano_tip`. That already lost events when the 1000-tx cap fired with the last admitted transaction inside the tip block, and a frontier cut would add a second instance. It now terminates on no cursor progress, and returns `ScanStalled` rather than spinning if the cursor can never advance. Costs one extra empty query round at the end of a scan.

`UTXO_PER_TX_OVERESTIMATE` moves to `midnight-primitives-cnight-observation` and is re-exported from the pallet, so the data source's row limit and the runtime's acceptance bound are one constant instead of two.

**Not changed:** the inherent data provider, the mock data source, the trait signature, `ObservedUtxos` (serialized into chainspec JSON), the inherent encoding, and the pallet's logic.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

The admitted set is deliberately unchanged. The shipped loop broke on the first event of the `tx_capacity`-th transaction, so it admitted at most `tx_capacity - 1` transactions; that off-by-one is preserved, because validators re-derive this inherent locally and a changed admitted set is consensus-visible whenever the cap fires.

`matches_shipped_truncation_when_no_category_saturates` is the runnable proof of that claim: it reimplements the shipped loop verbatim and asserts identical events and cursor over 1..=8 transactions x 1..=3 events across several blocks, for `tx_capacity` 2..=10. Removing the `- 1` makes it fail (verified), so it is not vacuous.

Also covered: grouping and sort-equivalence of flatten, the envelope caps (`max_utxos`, `tx_capacity`, lone oversized transaction, zero capacity), `truncate_at_position` boundaries, and three end-to-end cursor scenarios — a category's truncation not being masked by a later category's events, a row-limited fetch not advancing to the tip, and a cut at the cursor holding rather than stepping over an unobserved transaction.

```
cargo test -p midnight-primitives-mainchain-follower --lib   # 20 passed
cargo test -p pallet-cnight-observation                      # 32 + 3 passed
cargo clippy -p midnight-primitives-mainchain-follower -p midnight-node --all-targets   # clean
cargo fmt --check
```

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [x] Node Client Update
- [ ] Other:
- [ ] N/A

Client-only. Inherent derivation changes only when a category's raw query actually hits its row limit — the case that today silently skips events. Nodes on old and new binaries agree in every other case, including when the transaction-capacity cap fires.

## Links

Ports https://github.com/midnightntwrk/midnight-node/pull/1966

### DIFF
--- a/changes/node/changed/cnight-grouped-utxos.md
+++ b/changes/node/changed/cnight-grouped-utxos.md
@@ -1,0 +1,30 @@
+#node
+# Group cNIGHT observation events by Cardano transaction and make row-limited pulls gap-free
+
+The cNIGHT observation data source now builds its inherent from
+`CNightGroupedUtxos`, a type whose events enter and leave only in whole
+Cardano-transaction units. "Never emit part of a transaction" is now a property
+of the type rather than a call-site sort-and-truncate loop, so a partially
+emitted transaction (which every validator would then re-derive differently)
+cannot be reintroduced by a future edit.
+
+Each of the four observation queries is row-limited independently. A query that
+returns its full row limit is only proven complete below the position of its
+last row, and previously the loop reported the range as covered up to the
+Cardano tip regardless, silently skipping that category's remaining events. The
+merged set is now cut at the earliest saturated category's frontier, and the
+cursor resumes at the cut instead of the tip. Saturation is judged on raw rows,
+before the decoding and address filters drop any, so a query that survives
+filtering as far fewer events is still recognised as saturated.
+
+Genesis construction terminates its scan when the cursor stops advancing rather
+than when it enters the tip block, which is correct for the capacity cap and the
+new frontier cut alike; a cursor that cannot advance anywhere else now fails
+with an explicit error instead of looping forever.
+
+Inherent bytes are unchanged unless a category's raw query actually hits its row
+limit — the case that previously dropped events. A test reimplements the shipped
+truncation loop and asserts the new pipeline produces identical events and
+cursor over generated inputs.
+
+PR:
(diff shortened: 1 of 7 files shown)


ozgb:
Closing - not enough benefit given the LOC added - and not helpful if/when cardano txs have larger number of utxos per tx (we'll have to have a more granular CardanoPosition)

PR #2105, opened 2026-09-07
fix: for consistency use Display everywhere

gilescope:
#logging #hygiene
# Log transaction rejection reasons via Display rather than Debug

The mempool, pre-dispatch and apply sinks rendered a rejected transaction's
`TransactionInvalid` with `{:?}`, which prints the error's full payload -
including any ledger `StateValue` embedded in it - instead of the one-line
diagnostic its `Display` impl already provides. The `PartialSuccess` sink did
the same for its whole per-segment result map.

Switch all of them to `Display` (the segment map is now rendered as
`[<segment>: <reason>, ...]`). Log lines become shorter and readable; the
node-side `InvalidError` code returned to callers is unchanged, so nothing
downstream of the log is affected.

### DIFF
--- a/changes/node/changed/rejection-reason-display.md
+++ b/changes/node/changed/rejection-reason-display.md
@@ -12,4 +12,4 @@ Switch all of them to `Display` (the segment map is now rendered as
 node-side `InvalidError` code returned to callers is unchanged, so nothing
 downstream of the log is affected.
 
-PR: <link>
+PR: https://github.com/midnightntwrk/midnight-node/pull/2105
--- a/changes/node/changed/rejection-reason-display.md
+++ b/changes/node/changed/rejection-reason-display.md
@@ -1,0 +1,15 @@
+#logging #hygiene
+# Log transaction rejection reasons via Display rather than Debug
+
+The mempool, pre-dispatch and apply sinks rendered a rejected transaction's
+`TransactionInvalid` with `{:?}`, which prints the error's full payload -
+including any ledger `StateValue` embedded in it - instead of the one-line
+diagnostic its `Display` impl already provides. The `PartialSuccess` sink did
+the same for its whole per-segment result map.
+
+Switch all of them to `Display` (the segment map is now rendered as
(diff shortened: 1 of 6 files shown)


changes/node/changed/rejection-reason-display.md
```diff
@@ -0,0 +1,15 @@
...
+including any ledger `StateValue` embedded in it - instead of the one-line
+diagnostic its `Display` impl already provides. The `PartialSuccess` sink did
+the same for its whole per-segment result map.
+
+Switch all of them to `Display` (the segment map is now rendered as
+`[<segment>: <reason>, ...]`). Log lines become shorter and readable; the
+node-side `InvalidError` code returned to callers is unchanged, so nothing
+downstream of the log is affected.
+
+PR: <link>
```
gilescope:
```suggestion
PR: https://github.com/midnightntwrk/midnight-node/pull/2105
```

ledger/src/utils.rs
```diff
@@ -11,6 +11,32 @@
...
+
+/// Lazy `Display` for a `PartialSuccess` per-segment result map.
+///
+/// Renders directly into the formatter rather than building a string, so a log line that the
+/// level filter drops costs nothing and an enabled one allocates no intermediate.
+pub struct SegmentResults<'a, E>(pub &'a BTreeMap<u16, Result<(), E>>);
+
+impl<E: Display> Display for SegmentResults<'_, E> {
+	fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+		for (i, (id, res)) in self.0.iter().enumerate() {
```
LGLO, line 27: Cosmetic: since SegmentResults is plural, I think that square brackets display belongs here, not in call-sites.

LGLO (approved):
LGTM, happy to approve again if you decide to move brackets.

ozgb (approved):

ledger/src/utils.rs
gilescope: I'm going to leave that as a nit for another day. I'd like to get this landed.

PR #2123, opened 2026-09-10
fix(node): Sort ObservedUtxo in a way no dust registration or deregistration is lost

### LINKED ISSUE #2121: node: DUST address change leaves cNIGHT untracked on mainnet — valid mapping, no utxoOwners entry, 0 DUST/h
### Context & versions

- **Node:** mainnet public RPC `wss://rpc.mainnet.midnight.network` — `system_version` `1.0.2-eb71e64e`, `specVersion 1000000`
- **Network:** Midnight mainnet / Cardano mainnet (epoch 654)
- **DUST dApp:** https://midnight-dust-mainnet.nethermind.io/
- **Frequency:** observed once, in 1 of 1 DUST-address changes performed. Unknown whether deterministic.

Affected registration:

| | |
|---|---|
| Cardano address | `addr1qy33rj9e7xsltksh04vhtmpke08h0e24vjdw2ndkyuuhstn2uhk9588e2vvc8tr8aqh3scfyaq0mvhs7z5z2dkcnemkqgtr3sg` |
| Reward address (registered) | `0xe16ae5ec5a1cf9531983ac67e82f186124e81fb65e1e1504a6db13ceec` |
| DUST address | `mn_dust1w0q68584q35wjnfkc0u75vjaa6e0sxnc374l0y0wcw8lfp80ckzxjle9a5p` |
| DUST public key | `0x73c1a3d0f50468e94d36c3f9ea325deeb2f81a788fabf791eec38ff484efc58469` |
| Registration tx | [`b2ac0e33…67bccf`](https://cexplorer.io/tx/b2ac0e3396581ff3c307208f227ef958f9808184a5351c847f986d3b8967bccf), Cardano block 13,914,561, 2026-09-08T14:08:42Z |
| Midnight block that observed it | [`0x60c54eb7…f2a3ed`](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Frpc.mainnet.midnight.network#/explorer/query/0x60c54eb773064197e567b9e1c9d4f2370ac017d8519dc0edaf25ee34aef2a3ed) (#2,500,317) |

### Steps to reproduce

1. Hold cNIGHT on a Cardano address that already has a DUST registration.
2. Change the DUST address in the mainnet DUST dApp. The dApp builds a **single** transaction that
   burns the old registration token, mints the new one, and rotates the existing cNIGHT UTxOs into one
   new output — here `b2ac0e33…67bccf`, with the registration token at output `#0` and 568.463624
   cNIGHT at output `#1`.
3. Wait for the Cardano block to fall inside the observed window (confirmed below — the node has
   processed 961 blocks past it).
4. Check the registration is on Midnight — present and correct:

   ```bash
   curl -s -H 'Content-Type: application/json' -d '{"id":1,"jsonrpc":"2.0","method":"state_getStorage","params":["0xbbf4abef611bc3c9ca8cee3af9d8f7d14732239c3eff01cda7d8c21ecf0674531ce3bef62e9e78322c162d70ca84a49de16ae5ec5a1cf9531983ac67e82f186124e81fb65e1e1504a6db13ceec"]}' https://rpc.mainnet.midnight.network
   ```

   (`cNightObservation.mappings(0xe16ae5…ceec)`, equivalently via polkadot.js Chain state.)

5. Check `cNightObservation.utxoOwners` for the rotated cNIGHT UTxO `b2ac0e33…67bccf#1`. The map key is
   `BlakeTwo256(b"asset_create" || utxo_tx_hash || utxo_tx_index_be_u16)` =
   `0x0798f326d14c098fc83f669f1e3a8f001f2edd2b6f416925e5210560fb64a0f6`:

   ```bash
   curl -s -H 'Content-Type: application/json' -d '{"id":1,"jsonrpc":"2.0","method":"state_getStorage","params":["0xbbf4abef611bc3c9ca8cee3af9d8f7d18c9d9891ef9f047ab210d99af498ad09df1c74db974be5b09c13a8e0795bfd7e0798f326d14c098fc83f669f1e3a8f001f2edd2b6f416925e5210560fb64a0f6"]}' https://rpc.mainnet.midnight.network
   ```

6. Check the DUST balance / generation rate in the dApp and in the Midnight wallet.

### Expected behavior

The rotated cNIGHT UTxO is unspent and sits at an address whose reward address has a valid
registration, so it should be tracked as DUST-generating: `cNightObservation.utxoOwners` should hold an
entry for `b2ac0e33…67bccf#1` pointing at the registered DUST public key, and the dApp/wallet should
show a non-zero generation rate.

### Actual behavior

Generation rate is **0.000 DUST/h**; the dApp and the Midnight wallet both show 0 DUST (the wallet is
otherwise empty — no other shielded or unshielded tokens). No DUST has ever been received.

`cNightObservation.utxoOwners` has **no entry** for the rotated UTxO. Checked output indices 0–4 of the
registration tx, all `null`:

```
utxoOwners idx=0 nonce=185e39eea93ae34ddf627027251045a0a9c2d83489d65524bd1594c133112f9f -> None
utxoOwners idx=1 nonce=0798f326d14c098fc83f669f1e3a8f001f2edd2b6f416925e5210560fb64a0f6 -> None
utxoOwners idx=2 nonce=0903b8e76096340789f914b0fc4baaaa959fc84aee661a7c80bae244a93af6b9 -> None
utxoOwners idx=3 nonce=cbeb0774bc0948a79491ae19e6d06d50e5b21ba35cebdc0604ef35923076f411 -> None
utxoOwners idx=4 nonce=0bd0c3c5552ac4016773c9820854295d2635119b7b55919a726eef7ecb5b731d -> None
```

Everything else about the registration looks right:

- **Registration is on chain and unique.** `cNightObservation.mappings(0xe16ae5…ceec)` returns exactly
  one entry with the expected DUST key and registration UTxO:

  ```
  cardanoRewardAddress: 0xe16ae5ec5a1cf9531983ac67e82f186124e81fb65e1e1504a6db13ceec
  dustPublicKey:        0x73c1a3d0f50468e94d36c3f9ea325deeb2f81a788fabf791eec38ff484efc58469
  utxoTxHash:           0xb2ac0e3396581ff3c307208f227ef958f9808184a5351c847f986d3b8967bccf
  utxoIndex:            0
  ```

- **Midnight block # 2,500,317 has the events one would expect** — `Deregistration` + `MappingRemoved`
  for the old DUST address, then `Registration` + `MappingAdded` for the new one, then
  `SystemTransactionApplied`.

- **The Cardano tx is well-formed** (Koios `utxo_info`, `_extended: true`):
  - output `#0` — mapping-validator address `addr1w9e7ft4rrdd4rkdseguxr9hudfxyytm5ckh2qy0yhz7lfeg9lvhq7`,
    1 registration token (policy `73e4aea3…`), unspent, inline datum
    `{"constructor":0,"fields":[{"constructor":0,"fields":["6ae5ec5a1cf9531983ac67e82f186124e81fb65e1e1504a6db13ceec"]},"73c1a3d0f50468e94d36c3f9ea325deeb2f81a788fabf791eec38ff484efc58469"]}`
  - output `#1` — `addr1qy33rj9e7…gtr3sg`, stake address `stake1u94wtmz6rnu4xxvr43n7stccvyjws8aktc0p2p9xmvfuamq5ckkpt`
    (= the registered reward address), `568463624` of policy `0691b2fecca1ac4f53cb6dfb00b7013e561d1f34403b957cbb5af1fa`
    asset `4e49474854`, **`is_spent: false`**
  - the old registration token was burned and the two previous cNIGHT UTxOs were consumed as inputs

- **The block containing it has already been processed.** `cNightObservation.nextCardanoPosition`
  decodes to Cardano block **13,915,522** (`0x99e56f23…0a13`, 2026-09-08T19:35:44Z, `tx_index_in_block: 4`)
  — 961 blocks past the registration block, so this is not "not yet observed".

- **The mechanism works for other UTxOs.** `cNightObservation.utxoOwners` currently holds **5,424**
  entries chain-wide, and spot-checking unrelated registered addresses finds cNIGHT UTxOs that *do*
  resolve, e.g. `0bf7b041c059…#1` (Cardano block 13,839,176) → DUST key `0x730635f5…d77037c564`.
  So the storage-key derivation used above is confirmed against a working case.

### Impact

The affected holder's entire cNIGHT position (568.463624 cNIGHT) generates no DUST at all, with no
indication of a problem anywhere in the registration path — the dApp shows the correct balance and the
correct new DUST address, and the on-chain mapping is correct. Anyone changing their DUST address
through the dApp may be in the same state, and there is no user-visible signal to distinguish it from
"DUST is on its way". After cardano finalization (~12h) user will see in the DApp 0 DUST/h generation rate.

### Workaround

moving the cNIGHT into a fresh UTxO in a transaction separate from the registration (the dApp does not offer this, so it would have to be built by hand, e.g. sending 1 cNIGHT to your other wallet).

---

Investigated and written up with AI assistance (Claude Code, claude-opus-5); all on-chain values above
were queried against mainnet, not inferred.

LGLO:
# Overview

Sorts ObservedUtxos in a way that AssetCreate is not lost because it happens between registration and deregistration (when there are ambiguous mappings).

This consensus-sensitive change gated on spec_version.

Fixes #2121

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A

### DIFF
--- a/primitives/cnight-observation/src/lib.rs
+++ b/primitives/cnight-observation/src/lib.rs
@@ -372,6 +372,10 @@ impl core::fmt::Display for ObservedUtxoHeader {
 )]
 pub struct UtxoIndexInTx(pub u16);
 
+/// Legacy observed-UTXO ordering, used by nodes when runtime is below
+/// [`DATA_ORDERED_UTXOS_SPEC_VERSION`]. Orders by `tx_position`, then puts
+/// created UTXOs ahead of spent ones, then breaks ties on `utxo_tx_hash` and
+/// `utxo_index`. Needed for historical blocks verification.
 impl PartialOrd for ObservedUtxoHeader {
 	fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
 		match self.tx_position.partial_cmp(&other.tx_position) {
@@ -392,6 +396,59 @@ impl PartialOrd for ObservedUtxoHeader {
 			ord => return ord,
 		}
 		self.utxo_index.0.partial_cmp(&other.utxo_index.0)
+	}
+}
+
+/// First runtime `spec_version` whose `process_tokens` expects observed UTXOs
+/// ordered tx_position and then by variant rank.
+pub const DATA_ORDERED_UTXOS_SPEC_VERSION: u32 = 2_000_000;
+
+impl ObservedUtxoData {
+	/// Mapping mutations run before asset events so an `AssetCreate` always
+	/// observes the final registration state, and removals run before
+	/// additions.
+	pub fn ordering_rank(&self) -> u8 {
+		match self {
+			Self::Deregistration(_) => 0,
+			Self::Registration(_) => 1,
+			Self::AssetSpend(_) => 2,
+			Self::AssetCreate(_) => 3,
+		}
+	}
+}
+
+/// Total order over observed UTXOs.
+///
+/// `tx_position` stays the primary key and the variant rank must never be
+/// promoted above it: a single inherent's batch spans a whole observation
+/// window, so a UTXO's `AssetCreate` and the `AssetSpend` that consumes it can
+/// both land in one batch, and only the real on-chain Cardano order keeps the
+/// create ahead of the spend.
+///
+/// The trailing `utxo_tx_hash`/`utxo_index` tie-break is not semantically
+/// meaningful — order within one variant is irrelevant to processing — but it
+/// must stay deterministic across validators.
+pub fn cmp_observed_utxos_by_data(a: &ObservedUtxo, b: &ObservedUtxo) -> core::cmp::Ordering {
+	a.header
+		.tx_position
+		.block_number
+		.cmp(&b.header.tx_position.block_number)
+		.then_with(|| {
+			a.header
+				.tx_position
+				.tx_index_in_block
+				.cmp(&b.header.tx_position.tx_index_in_block)
+		})
+		.then_with(|| a.data.ordering_rank().cmp(&b.data.ordering_rank()))
+		.then_with(|| a.header.utxo_tx_hash.0.cmp(&b.header.utxo_tx_hash.0))
+		.then_with(|| a.header.utxo_index.0.cmp(&b.header.utxo_index.0))
+}
+
+pub fn sort_observed_utxos(utxos: &mut [ObservedUtxo], spec_version: u32) {
+	if spec_version >= DATA_ORDERED_UTXOS_SPEC_VERSION {
+		utxos.sort_by(cmp_observed_utxos_by_data);
+	} else {
+		utxos.sort();
 	}
 }
 
--- a/primitives/mainchain-follower/src/data_source/cnight_observation.rs
+++ b/primitives/mainchain-follower/src/data_source/cnight_observation.rs
@@ -24,6 +24,7 @@ use cardano_serialization_lib::{
 };
 use midnight_primitives_cnight_observation::{
 	CNightAddresses, CardanoPosition, CardanoRewardAddressBytes, DustPublicKeyBytes, ObservedUtxos,
+	sort_observed_utxos,
 };
 use sidechain_domain::{McBlockHash, McBlockNumber, McTxHash, McTxIndexInBlock, TX_HASH_SIZE};
 pub use sqlx::PgPool;
@@ -110,6 +111,7 @@ impl MidnightCNightObservationDataSource for MidnightCNightObservationDataSource
 		start_position: &CardanoPosition,
 		current_tip: McBlockHash,
 		tx_capacity: usize,
+		spec_version: u32,
 	) -> Result<ObservedUtxos, Box<dyn std::error::Error + Send + Sync>> {
 		let cnight_asset_name = config.cnight_asset_name.as_bytes();
 
@@ -261,7 +263,7 @@ impl MidnightCNightObservationDataSource for MidnightCNightObservationDataSource
 		utxos.extend(asset_create_utxos);
 		utxos.extend(asset_spend_utxos);
 
-		utxos.sort();
+		sort_observed_utxos(&mut utxos, spec_version);
 
 		// Truncate UTXOs but include full transactions
 		let mut truncated_utxos = Vec::with_capacity(utxo_capacity);
--- a/changes/node/changed/cnight-observed-utxo-data-ordering.md
+++ b/changes/node/changed/cnight-observed-utxo-data-ordering.md
@@ -1,0 +1,42 @@
+#node #cnight #consensus
+
+# Order observed cNIGHT UTXOs by data variant, gated on `spec_version`
+
+The `Vec<ObservedUtxo>` the node passes to `pallet-cnight-observation`'s
+`process_tokens` inherent is now sorted by `header.tx_position` and then by the
+`data` variant — `Deregistration`, `Registration`, `AssetSpend`, `AssetCreate` —
+instead of by the legacy "created UTXOs before spent UTXOs" rule.
+
+The runtime processes that vector in order, and `handle_create` needs a valid
+registration (exactly one mapping for the owner) at the moment it runs. Under
+the legacy rule an `AssetCreate` in the same Cardano transaction as a
+registration change is processed before the mapping settles, so it silently
+produces no DUST generation event and no `UtxoOwners` entry — permanently, for
+that UTXO. Ordering all mapping mutations ahead of all asset creates removes
+both cases: registering and receiving cNIGHT in one transaction, and rotating a
+mapping (which passes through an ambiguous two-mapping state) in one
+transaction. Putting removals before additions also stops the rotation from
+emitting a transient `Deregistration`/`Registration` pair that leaves the
+indexer's `cnight_registrations` row without its backing UTXO reference.
+
+`tx_position` deliberately stays the primary sort key: one inherent's batch
+spans a whole observation window, so a UTXO's `AssetCreate` and the
+`AssetSpend` consuming it can both land in one batch, and only the real
+on-chain Cardano order keeps the create ahead of the spend.
+
+Because the runtime's `check_inherent` compares the whole UTXO vector for
+equality, changing the order is consensus-breaking: a node using a different
+rule than the block author rejects the block, including on a full sync from
+genesis. The rule is therefore gated on
+`DATA_ORDERED_UTXOS_SPEC_VERSION`, resolved against the runtime in state at the
+parent block — the same runtime that executes the child block and that
+`check_inherents` is called against, so the authoring and verifying paths always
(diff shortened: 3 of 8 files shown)


primitives/cnight-observation/src/lib.rs
```diff
@@ -395,6 +399,59 @@ impl PartialOrd for ObservedUtxoHeader {
 	}
 }
 
+/// First runtime `spec_version` whose `process_tokens` expects observed UTXOs
+/// ordered tx_position and then by variant rank.
+pub const DATA_ORDERED_UTXOS_SPEC_VERSION: u32 = 2_100_000;
```
LGLO: This change will be perhaps backported to v2.1.0 (with value of 2_000_000 here). Using 3_000_000 here on `main` would cause that during upgrade from v2.1.0 to v3.0.0, the new nodes would stop using new sorting algorithm and fork would happen.

primitives/mainchain-follower/src/data_source/cnight_observation.rs
```diff
@@ -261,7 +263,7 @@ impl MidnightCNightObservationDataSource for MidnightCNightObservationDataSource
 		utxos.extend(asset_create_utxos);
 		utxos.extend(asset_spend_utxos);
```
Klapeyron, line 264: if the right order is Deregistration -> Registration -> Spend -> Create, will be it sufficient to just extend vec in right order here?

primitives/cnight-observation/src/lib.rs
```diff
@@ -395,6 +399,59 @@ impl PartialOrd for ObservedUtxoHeader {
 	}
 }
 
+/// First runtime `spec_version` whose `process_tokens` expects observed UTXOs
+/// ordered tx_position and then by variant rank.
+pub const DATA_ORDERED_UTXOS_SPEC_VERSION: u32 = 3_000_000;
```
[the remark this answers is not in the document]
LGLO: All nodes in the network observe the same spec_version and version in code is not version deployed in networks.

primitives/cnight-observation/src/lib.rs
```diff
@@ -395,6 +399,59 @@ impl PartialOrd for ObservedUtxoHeader {
...
+		.then_with(|| a.header.utxo_index.0.cmp(&b.header.utxo_index.0))
+}
+
+pub fn sort_observed_utxos(utxos: &mut [ObservedUtxo], spec_version: u32) {
+	if spec_version >= DATA_ORDERED_UTXOS_SPEC_VERSION {
+		utxos.sort_by(cmp_observed_utxos_by_data);
+	} else {
+		utxos.sort();
+	}
+}
```
Klapeyron, line 453: I am not a big fan of the SPEC_VERSION dependencies, they represent rather entire runtime configuration and are quite volatile, rather I suggest to stay closer to the interface we operate on, like `midnight_primitives_cnight_observation::CNightObservationApi`, which we can just version like `#[api_version(2)]` without changing any of the functions in that Api. Having that will allow us to build logic around it like for instance following helper:

```rust
/// How the native inherent data provider must sort `ObservedUtxo`s so that
/// `check_inherent`'s vector equality matches the parent runtime.
///
/// Native infers this from `CNightObservationApi`'s `api_version` at the
/// parent block. Version [`Self::RUNTIME_API_VERSION`] and newer use
/// [`Self::Data`]; older WASM falls back to [`Self::LegacyHeader`].
#[derive(Copy, Clone, Eq, PartialEq, Debug)]
pub enum ObservedUtxoOrdering {
	/// Header `PartialOrd`: create-before-spend within a transaction.
	LegacyHeader,
	/// `tx_position`, then variant rank (see [`cmp_observed_utxos_by_data`]).
	Data,
}

impl ObservedUtxoOrdering {
	/// First `CNightObservationApi` version whose `process_tokens` expects
	/// [`Self::Data`] ordering. Advertised by the runtime impl.
	pub const RUNTIME_API_VERSION: u32 = 2;

	pub fn from_api_version(api_version: Option<u32>) -> Self {
		match api_version {
			Some(v) if v >= Self::RUNTIME_API_VERSION => Self::Data,
			_ => Self::LegacyHeader,
		}
	}
}

pub fn sort_observed_utxos(utxos: &mut [ObservedUtxo], ordering: ObservedUtxoOrdering) {
	match ordering {
		ObservedUtxoOrdering::Data => utxos.sort_by(cmp_observed_utxos_by_data),
		ObservedUtxoOrdering::LegacyHeader => utxos.sort(),
	}
}
```
Which we can later just call on the node side:
```rust
		let ordering = ObservedUtxoOrdering::from_api_version(
			api.api_version::<dyn CNightObservationApi<Block>>(parent_hash)?,
		);
```

primitives/mainchain-follower/src/data_source/cnight_observation.rs
LGLO: No. Primary ordering is by TxPosition (block number and tx index in block). This has to be sorted this way (as explained by the comment:
```
/// Total order over observed UTXOs.
///
/// `tx_position` stays the primary key and the variant rank must never be
/// promoted above it: a single inherent's batch spans a whole observation
/// window, so a UTXO's `AssetCreate` and the `AssetSpend` that consumes it can
/// both land in one batch, and only the real on-chain Cardano order keeps the
/// create ahead of the spend.
///
/// The trailing `utxo_tx_hash`/`utxo_index` tie-break is not semantically
/// meaningful — order within one variant is irrelevant to processing — but it
/// must stay deterministic across validators.
```

primitives/cnight-observation/src/lib.rs
LGLO: I think `spec_version` is simple and just works in this scenario. It is sound to assume it only ever increases (just like pallet API version).
By using it I didn't have to touch code in the pallet. We don't need new metadata for sake of this marker.

Also, why introduce new Runtime API version when Runtime API didn't change?

Klapeyron: Runtime API is the first-look interface between runtime and the node, in a case like here we would like to know some details of the runtime data, we can use that information via runtime API version, it is also idiomatic way of solving that kind of issues in polkadot-sdk https://github.com/search?q=repo%3Aparitytech%2Fpolkadot-sdk+has_api_with&type=code

LGLO: It is idiomatic for the SDK and it makes sense in such a case. This repo is not an SDK. Midnight has its well defined versions and spec_version follows them.

API didn't change. This is not about API. It is about synchronization over the network.
We know that runtime upgrade will be performed when all nodes are updated. This is the signal for the nodes.
It does not require additional governance action and additional code.

Klapeyron (approved):

PR #2160, opened 2026-09-15
feat(runtime): 2.1.0 benchmark weights

### LINKED ISSUE #2126: [Release 2.1.0] Re-run substrate benchmarks and land the weights PR
Re-run the substrate benchmarks for 2.1.0 on the benchmark machine and land the resulting weights PR on `release/node-2.1.0`. Weights generated on dev hardware are not acceptable.

Done when:

- [ ] Get SRE to recreate benchmarking machine.
- [ ] Benchmarks run on the benchmark machine
- [ ] Weights PR merged to `release/node-2.1.0`
- [ ] Any weight that moved more than 2x called out here

Part of #2124.

_Issue drafted with AI assistance (Claude Code)._

justinfrevert:
Weights and benchmarks for 2.1.0. Benchmarks were run on the spec machine — the mainnet benchmark syncbox (`c7i.xlarge`, Intel Xeon Platinum 8488C), the same CPU as the previous run.

Of 122 functions compared, these moved more than 2x:

| | ratio | function | before → after |
|---|---|---|---|
| ↑ | 2.51x | `pallet_c2m_bridge::add_approved_mc_tx_hashes` | 1,117,825 → 2,801,984 |
| ↑ | 2.17x | `pallet_c2m_bridge::set_subminimal_transfers_config` | 1,390,000 → 3,021,000 |
| ↑ | 2.12x | `pallet_partner_chains_bridge::set_main_chain_scripts` | 1,880,000 → 3,979,000 |
| ↓ | 0.39x | `pallet_cnight_observation::set_cnight_identifier` | 10,000,000 → 3,896,000 |
| ↓ | 0.38x | `pallet_cnight_observation::set_mapping_validator_contract_address` | 10,000,000 → 3,847,000 |
| ↓ | 0.36x | `pallet_cnight_observation::set_auth_token_asset_name` | 10,000,000 → 3,604,000 |
| ↓ | 0.28x | `pallet_timestamp::on_finalize` | 5,967,000 → 1,676,000 |

The upward moves are a correction: those two files were previously generated on a developer workstation rather than the spec machine. The downward cNIGHT ones replace hand-written placeholders.

Issue: https://github.com/midnightntwrk/midnight-node/issues/2126


LGLO (dismissed):

pallets/c2m-bridge/src/lib.rs
```diff
@@ -190,6 +190,12 @@ pub mod pallet {
 
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		// Pre-account for on_finalize weight (storage write to reset the transfer counter).
```
LGLO: ~I'm also thinking about removal of `on_finalize`, because TransferCounter could use regular addition with a wrap around logic and it will be as good for nonce reasons.~

So, this is overcharging, because cost of `writes(1)` assumes disk operation, but the `TransferCounter` is never written to disk (it is None at the end of each block). All happens in memory.

LGLO (approved):

chrispalaskas:
Heads-up from deploying this to perfnet: **this PR changed the runtime without raising `spec_version`**, which makes the new weights impossible to enact on an already-running chain via the supported path.

Weights live in the runtime WASM and the WASM lives in chain state, so adopting them needs a runtime upgrade — and `frame_system::can_set_code` rejects a runtime whose `spec_version` does not increase (`SpecVersionNeedsToIncrease`). On a chain already reporting `002_001_000`, both `set_code` and `apply_authorized_upgrade` refuse the 2.1.0 runtime. The only remaining route is `set_code_without_checks`, which costs more than it looks:

- `frame_executive` gates `on_runtime_upgrade` on the version changing, so migrations are skipped (harmless for a weights-only change, but the mechanism is silently bypassed);
- `last_runtime_upgrade` keeps reporting the old value, so **the chain cannot tell you which runtime it is executing**. For a network whose output is comparative benchmark numbers, a measurement that cannot be attributed to a runtime defeats the point of re-benchmarking.

There is also a self-verification angle: the toolkit's `runtime-upgrade` command confirms enactment by watching for the `spec_version` bump and errors `UpgradeNotEnacted` otherwise, so without a bump there is no positive signal that the code actually swapped.

We worked around it locally with a one-line bump to `002_001_001` (`transaction_version` and `authoring_version` untouched, since extrinsic encoding is unchanged) and completed the upgrade on perfnet today: spec `002_001_000` → `002_001_001`, live from block 84,041.

Not proposing a change to this PR since it is merged — flagging it so the next person upgrading a live chain onto 2.1.0 does not hit the same wall, and so the bump can ride along with whatever lands next.

PR #2161, opened 2026-09-15
fix(toolkit): accept spec version 1.0.300 when replaying chain history

rsporny:
# Overview
RuntimeVersion::try_from matched exact spec versions and knew 001_000_003 (1.0.3) but not 001_000_300 (1.0.300), which devnet ran before the 2.1.0 hard fork. Any replay reaching a pre-fork devnet block aborted with UnsupportedBlockVersion(1000300). Both versions are ledger-8 era — LedgerVersion::from_spec_version already treats the whole 000_022_000..=001_999_999 range as Ledger8 — so they share a decoder and map to one variant.

Closes: https://github.com/midnightntwrk/midnight-node/issues/2158

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [x] All commits are signed off (`git commit -s`) for the DCO
- [x] Changes are backward-compatible (or flagged if breaking)
- [ ] Pull request description explains why the change is needed
- [ ] Self-reviewed the diff
- [ ] I have included a change file, or skipped for this reason: 
- [ ] If the changes introduce a new feature, I have bumped the node minor version
- [ ] Update documentation (if relevant)
- [ ] Updated AGENTS.md if build commands, architecture, or workflows changed
- [ ] No new todos introduced

## 🧪 Testing Evidence

Please describe any additional testing aside from CI:

- [ ] Additional tests are provided (if possible)

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [ ] Node Client Update
- [ ] Other: 
- [ ] N/A

PR #2163, opened 2026-09-15
fix(toolkit): accept spec version 1.0.300 when replaying chain history

### LINKED ISSUE #2158: toolkit: replaying devnet history aborts with UnsupportedBlockVersion(1000300) — spec 1.0.300 missing from the runtime version table
### Context & versions

- **Toolkit:** built from `release/node-2.1.0`
- **Network:** devnet — chain ran node **1.0.300** (`spec_version 001_000_300`) before the 2.1.0 hard fork, so all pre-fork blocks carry that spec version
- **Frequency:** deterministic — fails on every replay that reaches a pre-fork devnet block

### Steps to reproduce

1. Point the toolkit at devnet with a cache that does not already cover the pre-fork range:
   ```
   ./target/release/midnight-node-toolkit show-night-pools \
     -s wss://rpc.devnet.midnight.network \
     --fetch-cache redb:$HOME/qanet-cache/fetch_cache.db \
     --ledger-state-db $HOME/qanet-cache/ledger_cache_db \
     --fetch-concurrency 8
   ```

### Expected behavior

The toolkit replays devnet history, including the 1.0.300 blocks, and prints the NIGHT pools.

### Actual behavior

The compute stage aborts as soon as it reaches a pre-fork block:

```
chain head (finalized): 56977, verified in cache: 42113, fetching 14865 blocks
spawning 8 fetch workers (capped from requested, 149 jobs)
spawning 8 compute workers
compute task error
Error: ComputeTaskError(RuntimeVersionError(UnsupportedBlockVersion(1000300)))
```

### Impact

Any toolkit command that replays devnet history from genesis is unusable: `show-night-pools`, `show-wallet` on a cold cache, `generate-txs` without a warm one. A warm cache covering the pre-fork range hides it, so it surfaces on new machines and fresh caches — which also means it will bite anyone reproducing an issue from scratch.

Blocks reading the genesis NIGHT pools, which is currently needed for Cardano-to-Midnight bridge activation work on devnet.

### Workaround

Reuse a cache warmed by a toolkit built from the 1.0.300 tree. Nothing available for a fresh cache.

### Analysis

Verified against `release/node-2.1.0`.

`util/toolkit/src/fetcher/runtimes.rs` maps spec versions to `RuntimeVersion` with an exhaustive match on exact values:

```rust
001_000_000 => Ok(Self::V1_0_0),
001_000_003 => Ok(Self::V1_0_3),
002_000_000 => Ok(Self::V2_0_0),
```

`001_000_003` is the 1.0.3 release line. devnet ran **1.0.300** — `001_000_300` — which no arm matches, so it falls through to `UnsupportedBlockVersion`.

Note the rest of the stack already handles it correctly. `LedgerVersion::from_spec_version` (`ledger/helpers/src/fork/raw_block_data.rs:194`) classifies by **range**:

```rust
000_022_000..=001_999_999 => Some(LedgerVersion::Ledger8),
002_000_000..              => Some(LedgerVersion::Ledger9),
```

So `001_000_300` is already Ledger8 there. Only this exact-value table rejects it.

Because both versions are ledger-8 era and share a decoder, aliasing them onto one variant is exact rather than approximate — `to_spec_version()` feeds straight back into the range-based `from_spec_version`, so the round trip lands on Ledger8 either way:

```rust
001_000_003 | 001_000_300 => Ok(Self::V1_0_3),
```

No new match arm is needed in `compute_task.rs`. `cargo check -p midnight-node-toolkit` passes.

Worth a wider look: an exact-value table will keep breaking whenever a release line uses a spec version nobody added here, and the failure mode is an opaque abort mid-replay. Matching by range, as `from_spec_version` already does, would avoid the whole class.

justinfrevert:
# Overview

The toolkit's block fetcher rejects any block whose `spec_version` it does not recognise.
`RuntimeVersion::try_from` knew `001_000_003` (1.0.3) but not `001_000_300` (1.0.300), the runtime
devnet ran before the 2.1.0 hard fork. Any replay reaching a pre-fork devnet block aborted:

```
Error: ComputeTaskError(RuntimeVersionError(UnsupportedBlockVersion(1000300)))
```

This breaks `show-night-pools`, `show-wallet` on a cold cache and `generate-txs` without a warm
one. A cache already covering the pre-fork range hides it, so it only surfaces on fresh machines —
including anyone reproducing an issue from scratch.

### Why aliasing onto `V1_0_3` is exact, not approximate

The 1.0.300 runtime metadata was bundled on `jacek-toolkit-1.0.3-sync-perf` by #2103. Diffing it
against the `1.0.3` blob on `main`, the two files are byte-identical except for **two bytes** — the
`spec_version` u32 in the `System.Version` constant (`6C 43 0F 00` = 1000300 vs `43 42 0F 00` =
1000003). Type registry, pallet indices and call/event schemas are the same, so the bundled 1.0.3
subxt metadata decodes 1.0.300 blocks correctly.

Both are also ledger-8 era: `LedgerVersion::from_spec_version`
(`ledger/helpers/src/fork/raw_block_data.rs`) already classifies the whole
`000_022_000..=001_999_999` range as `Ledger8`. `to_spec_version()` feeds straight back into it, so
the round trip lands on `Ledger8` either way.

No new match arm is needed in `compute_task.rs`, and no second near-duplicate 129 KB metadata blob
needs bundling.

### Relationship to #2161 and #2104

- #2161 makes the same one-line change but targets `release/node-2.1.0`. Release branches are not
  merged back into `main` (`release/node-2.0.0` and `release/node-1.0.3` tips are not ancestors of
  `main`), so `main` needs its own fix — this PR.
- #2104 solved it on the `jacek-toolkit-1.0.3-sync-perf` feature branch by *swapping* 1.0.3 out for
  1.0.300 rather than accepting both. That branch's node is on 1.0.300 only; `main` has to replay
  chains carrying either, so this PR accepts both.

## 🗹 TODO before merging

- [x] Ready

## 📌 Submission Checklist

- [ ] Self-reviewed the diff
- [x] No new todos introduced

## 🧪 Testing Evidence

Two regression tests added in `util/toolkit/src/fetcher/runtimes.rs`:

- `spec_1_0_300_maps_to_ledger8` — `001_000_300` decodes and round-trips to the same
  `LedgerVersion` the node itself assigns it.
- `supported_variants_have_a_ledger_version` — every dispatchable variant has a `LedgerVersion`, so
  none can fail with `LedgerVersionMissing` mid-replay.

```
$ cargo test -p midnight-node-toolkit --lib fetcher::runtimes
test fetcher::runtimes::tests::spec_1_0_300_maps_to_ledger8 ... ok
test fetcher::runtimes::tests::supported_variants_have_a_ledger_version ... ok
test result: ok. 2 passed; 0 failed
```

`cargo clippy -p midnight-node-toolkit --all-targets` and `cargo fmt --check` are clean.

The second test excludes `V0_21_0` deliberately: spec `000_021_000` is below the `000_022_000`
ledger-8 floor, so that variant can only ever fail with `LedgerVersionMissing`. That is pre-existing
and intentional (pre-ledger-8 chains are no longer supported) — noted here rather than changed,
since it is out of scope for this issue.

## 🔱 Fork Strategy

- [ ] Node Runtime Update
- [x] Node Client Update
- [ ] Other:
- [ ] N/A

## Links

Closes #2158

Related: #2161 (same fix, `release/node-2.1.0`), #2103 / #2104 (1.0.300 metadata on
`jacek-toolkit-1.0.3-sync-perf`)

### Follow-up worth considering (not in this PR)

An exact-value table will keep breaking whenever a release line ships a spec version nobody added
here, and the failure mode is an opaque abort mid-replay. Two options, neither obviously free:
matching by range as `from_spec_version` does would trade a loud abort for a possible silent
mis-decode when metadata genuinely changes; keeping exact matching but reporting the nearest known
version in the error keeps it loud and makes the fix self-evident. Worth a separate discussion.


rsporny (approved):

util/toolkit/src/fetcher/runtimes.rs
```diff
@@ -37,7 +37,12 @@ impl TryFrom<u32> for RuntimeVersion {
 			000_021_000 => Ok(Self::V0_21_0),
 			000_022_000 => Ok(Self::V0_22_0),
 			001_000_000 => Ok(Self::V1_0_0),
-			001_000_003 => Ok(Self::V1_0_3),
+			// devnet ran spec 001_000_300 ("1.0.300") on the 1.0.x line, alongside the
+			// 001_000_003 ("1.0.3") releases. The two runtimes' metadata differ only in the
+			// spec_version constant itself, and both are ledger-8 era — `LedgerVersion::
+			// from_spec_version` classifies the whole 000_022_000..=001_999_999 range as
+			// Ledger8 — so they share a decoder and mapping them to one variant is exact.
+			001_000_003 | 001_000_300 => Ok(Self::V1_0_3),
```
LGLO, line 45: If 1_000_300 / v_1_0_300 is going to be mainnet, then I don't think making this unusual mapping helps. Lets make a proper `RuntimeVersion` for it.

PR #2187, opened 2026-09-22
fix: remove gdb

gilescope:
Check 1.0.300 release passes all checks.

There's some critical vulnerabilities in optional deps for gdb. Gdb is not needed for production images and should not be included by default.


ozgb (dismissed):

jacek-kurkowski-shielded (dismissed):
BEST PR EVER 😄

ozgb (approved):

